PR #9, opened 2018-11-19
Channel Calendar: Add {cur_month} variable to assist highlight/lowlight on calendar display

stevesweets:
Have used this tag recently to output a calendar for data not stored in a channel (loaded from an api).  Needed to be able to lowlight date boxes not in the currently displayed month and discovered this wasn't possible with current variables if you're not displaying a channel.

Thus, add 'cur_month' variable to assist highlighting / lowlighting of month labels when using this tag to output a calendar for other means.


kevincupp:
Closing due to inactivity, but let us know if you're able to make the requested changes.

PR #10, opened 2018-11-19
Improve DocBlocks

litzinger:
• Fix some typos
• Remove unused $types variable
• Start using short array syntax since PHP 5.6 is the new minimum version
• Add missing parameters, throws, and return statements to docblocks
• Properly case string and array in docblocks
• Add typehint for anything that returns a DB result (IDE’s appreciate this)

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/Service/Addon/Addon.php
+++ b/system/ee/EllisLab/ExpressionEngine/Service/Addon/Addon.php
@@ -47,13 +47,11 @@ class Addon {
 	 */
 	public function isInstalled()
 	{
-		$types = array('modules', 'fieldtypes', 'extensions');
-
 		ee()->load->model('addons_model');
-
 
 		if (is_null(self::$installed_modules))
 		{
+			/** @var \CI_DB_result $query */
 			$query = ee()->addons_model->get_installed_modules();
 
 			self::$installed_modules = array();
@@ -71,6 +69,7 @@ class Addon {
 
 		if (is_null(self::$installed_extensions))
 		{
+			/** @var \CI_DB_result $query */
 			$query = ee()->addons_model->get_installed_extensions();
 
 			self::$installed_extensions = array();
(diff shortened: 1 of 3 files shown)


litzinger:
I could go on and on about fixing the doc blocks, unused vars, and unused "use" statements... this is just a start :)

kevincupp:
Thanks, Brian, I think this looks fine. Could you rebase or merge upstream stability into this branch? We've since hooked up our automated testing, just want to run it through that, even though code changes are minimal.

system/ee/EllisLab/ExpressionEngine/Service/Addon/Addon.php
```diff
@@ -47,13 +47,11 @@ public function __call($fn, $args)
...
 	{
-		$types = array('modules', 'fieldtypes', 'extensions');
-
 		ee()->load->model('addons_model');
 
-
 		if (is_null(self::$installed_modules))
 		{
+			/** @var \CI_DB_result $query */
 			$query = ee()->addons_model->get_installed_modules();
```
litzinger: @kevincupp I wonder if there is a better way to do this. Other systems have had issues with the global whatever() function, which may be the root cause of this b/c some of these methods do have the correct return type in the doc blocks, but the method chain may be breaking down?

kevincupp: Yeah I'd prefer not having to put a comment everywhere something like this is used. What's the problem, exactly? Your IDE doesn't know the return type of `ee()`? Or it can't handle chaining methods like this?

derekjones:
Hey @litzinger just checking to see if you intend to take this further, and if there is a resolution with the return value of `ee()->` singleton calls?

obfuscode:
Closing PR due to lack of response.

litzinger:
For anyone stumbling across this PR, adding this file to the root of your project, if using PhpStorm, will greatly improve things.

https://gist.github.com/litzinger/1b52d0363bbf1c53ed3d6515ebd1cbc9

With that said, if I'm being honest, the doc blocks in EE's core classes, particularly legacy files, are a hot mess. I hope there is improvement going forward.

PR #17, opened 2018-11-21
Use bycrypt for password hashing

kevincupp:
This uses bcrypt for new member accounts, and updates old ones as they log in.

Be aware: if you test this by logging in and then switch to another branch, you cannot login again without putting back your old hash and salt.

Why bcrypt? Well, our current hashing method is a little too quick by modern standards, making them easier to crack if there is a database leak. The main advantage of bcrypt is it takes more time to encrypt, and thus crack, these password hashes, based on a cost factor of our choosing. So, cracking via GPU isn't as effective, and there's no salt needed.

We're currently using the default cost factor, which seems to be a good balance of time-to-crack and speed of encryption, see:

https://security.stackexchange.com/questions/85966/how-does-phps-password-hash-bcrypt-cost-factor-translate-into-cracking-comput


mrw:
Hi @kevincupp! Regarding this:

> We currently use good, unique salts for each member, but it's a little pointless to have them stored right next to the password in the database, assuming the members table is what gets leaked in a breach.

It's fine to do this (store them next to each other) and in fact that's how bcrypt works as well; the salt is there as part of the output hash and just text delimited inside of it. 

I agree with using bcrypt with a high cost factor in general but just want to make sure you're optimizing for the correct changes/worried about the right things when making this change.

kevincupp:
Thanks, @mrw. It's actually not fine to do this, unless you're using algorithms designed for password hashing, which we aren't. If you're going to use these weaker algorithms, the salts should at least be securely stored elsewhere to be effective, so it is still a valid thing to worry about.

kevincupp:
I appreciate that, Matt, but you haven't explained why it is fine. So, I also don't want people reading this ticket to think what you're describing is acceptable practice. Sure, salts per user and in the database helps when trying to crack an entire database of passwords, but not by much on smaller databases, and not at all for a targeted attack on a single account like a super admin. It could be that we just have different standards for "fine" in which case we can agree to disagree.

mrw:
Salts per user and stored in the database is standard practice and is what
you’re getting with bcrypt as well, which also stores salts per-user and in
the database as plain (unencrypted) text. It’s totally fine to do that.
Your security benefit here comes from an improved hashing algorithm, not a
difference in salt storage. Sounds like it might be better to take this
offline. Happy to discuss further on slack or similar if you’d like. And no
problem if not. But my goal was just to note that there’s nothing insecure
about EE’s current salt storage as-is, so current users don’t get nervous
about that specific piece, and also to thank you for working on the bcrypt
move- which is really important and an awesome step forward!

kevincupp:
I understand how bcrypt works, my stated disadvantages are in the context of using weak hashing algorithms. Again, I'm not making a case against salt storage in general, just in the case of using these weaker algorithms. Please see my first response: https://github.com/ExpressionEngine/ExpressionEngine/pull/17#issuecomment-441340023 The salt storage is insecure in relation to the hashing algorithms used.

I'm curious why we should take this offline. It seems it would be beneficial to resolve the discussion where it started for the benefit of everyone else.

mrw:
> I'm curious why we should take this offline.

Because we’re not getting anywhere here. I was trying to help make sure security implications of the current method are clear. I disagree with what you said and believe that salt storage in the database is always a totally fine practice, but to your point before, we can agree to disagree. I appreciate your work on Bcrypt here which we both agree will help everyone, which is the important part.

PR #44, opened 2018-12-13
Feature/no snippets for actions

low:
## Overview

By default, Snippets are loaded for each PAGE *and* ACTION request. The latter does not automatically invoke the template renderer, so it seems overkill to automatically load them for those requests. Limiting the automatic load to PAGE requests only improves performance for ACTION requests.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [ ] Yes
- [x] No

### DIFF
--- a/changelogs/patch.rst
+++ b/changelogs/patch.rst
@@ -18,6 +18,7 @@ Patch Release
    - Fixed Bug (#<issue number>) where <bug behavior>.
 
 - Optimized the member delete routine's heir assignments.
+- Optimized ACTION request to not automatically load Template Partials.
 - Fixed a bug where a PHP error may appear when the CP homepage newsfeed cannot be fetched.
 - Fixed a bug where extension hooks may run during a one-click upgrade.
 - Fixed a bug where a supplied class was not added to "select" fields in the shared form view.
--- a/system/ee/legacy/libraries/Core.php
+++ b/system/ee/legacy/libraries/Core.php
@@ -347,7 +347,7 @@ class EE_Core {
 		}
 
 		// Load up any Snippets
-		if (REQ == 'ACTION' OR REQ == 'PAGE')
+		if (REQ == 'PAGE')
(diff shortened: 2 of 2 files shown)


derekjones:
Thanks @low! This is _not_ backwards compatible however, as it changes what's available to a developer in the `ACTION` context, that has been present [for over nine years](https://github.com/ExpressionEngine/ExpressionEngine/blob/6029ffdef522146923ed63f92b4ff8b6a1ec00f7/system/expressionengine/libraries/Core.php#L334).

On example that comes to mind are any add-ons that instantiate the Template parser in their ACTION. For instance this breaks the the built-in Comment module's comment preview. Other add-ons that allow rich email notifications using the Template parser would be broken as well, and I'm sure there are other cases.

I don't disagree with the change, because it's the minority of ACTIONs that need it. But breaking changes are breaking changes and bad for end users and frustrating for developers. So this will need to retarget the v6 branch, and give some consideration to what developers need to add who were relying on this behavior. If there is a way that it could occur *without* breaking, then that would be even best, though the API change would still probably require this be a major version adjustment. If that's not possible then there also needs to be a public method and documentation that developers have to do to manually add them (my gut tells me the latter is onerous).

I changed the base but it brought it too many changes due to the parent of your feature branch I suspect. Can you reopen this PR from a branch made off of an updated `release/6.0.0` please? Alternatively if you can think of a way to allow all existing ACTIONS that use the Template library to have this that doesn't change behavior, it could be re-opened and remain in `stability`. Since there may be hooks in-between I'm not sure that's possible, but maybe you are thinking of something I am not. Thanks!

low:
Thanks @derekjones. If I can find the time, I'll see if I can elaborate more.

Issue #45, opened 2018-12-13
After default member fields are migrated in EE4, undefined index to the url member field


matthewjohns0n:
**Description of the problem**
When upgrading a site that has content in the `url` member field (and it was in use), it is migrated to a new custom member field. When a field with "url" in it is used, for example the `{url_title}` variable, ExpressionEngine tries to set the member_url variable in Simple_variable.php at the top of the `_url()` function. It seems clear it is being set at the top of the function so it can be used in some of the later `else` cases. However, sometimes `$mfields['url']` is set, but `$data['m_field_id_'.$mfields['url'][0]]` is not set, and PHP gets sad. PHP thows a notice that m_field_id_12 is an undefined index. See error below.

**How To Reproduce**
1. Use the default member field `{url}` in ExpressionEngine 3
2. Upgrade site from ExpressionEngine 3 to 4 (or continue to 5).
3. Use the `{url_title}` in a context where the member fields are not loaded. I believe this is any channel entries loop.
4. See error below.

**Error Messages**
<pre><code>
Notice
Undefined index: m_field_id_12
ee/legacy/libraries/channel_entries_parser/components/Simple_variable.php, line 296

Severity: E_NOTICE
</code></pre>

**Environment Details:**
 - Version: 5.0.1
 - PHP Version: 7.1
 - MySQL Version: 5.7
 - OS: Docker running on a Mac
 - Web Server: nginx

**Possible Solution**
It also needs to be verified that `$data['m_field_id_'.$mfields['url'][0]]` is set before attempting to use it. PR coming.

kevincupp:
I'm unable to reproduce this, maybe there are some steps missing to reproduce? Here's what I tried:

1. Installed a fresh copy of latest EE3 with default theme
2. Filled in the URL in my profile
3. Went ahead and created a test template with the following:
```
{exp:channel:entries}
{url_title}
{/exp:channel:entries}
```
4. Upgraded to EE5, double-checked the DB that the member field was converted correctly
5. Went to visit template, showed all my URL titles. Tried this:
```
{exp:channel:entries disable="member_data"}
{url_title}
{/exp:channel:entries}
```
6. Still no PHP error
Let me know what else I can do to reproduce.

kevincupp:
Closing due to inactivity, but let us know if you have more information to share about this.

PR #46, opened 2018-12-13
Bug/member field url check in simple parser

matthewjohns0n:
## Overview

Added a check to make sure `$data['m_field_id_'.$mfields['url'][0]]` is set before using it. PHP throws a warning if it it tries to access it and it is not set.

Resolves [#45](https://github.com/ExpressionEngine/ExpressionEngine/issues/45).

## Nature of This Change

- [X ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X ] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/channel_entries_parser/components/Simple_variable.php
+++ b/system/ee/legacy/libraries/channel_entries_parser/components/Simple_variable.php
@@ -285,8 +285,12 @@ class EE_Channel_simple_variable_parser implements EE_Channel_parser_component {
 	 */
 	protected function _urls($data, $tagdata, $key, $val, $prefix, $mfields)
 	{
-			// URL was moved to a custom member field or dropped
-			$member_url = (isset($mfields['url'])) ? $data['m_field_id_'.$mfields['url'][0]] : '';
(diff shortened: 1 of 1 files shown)


kevincupp:
Thanks, @matthewjohns0n! Do you mind further investigating why that `$data` array might not be populated with the field? As far as I can tell, if the field is in `$mfields`, the data should be there, too. Does it only happen when you have `disable="member_data"` set? Just want to make sure this isn't masking some underlying problem, or if it is, that it's the best way to get around it. Also curious why this hasn't come up yet, I'd think it'd be a wider issue if the steps to reproduce were that simple. Maybe something else happened during the upgrade of your install?

Also, just a couple nitpicky style things:

https://docs.expressionengine.com/latest/development/guidelines/general.html#bracket-and-parenthetic-spacing
https://docs.expressionengine.com/latest/development/guidelines/general.html#code-indenting

kevincupp:
Closing due to inactivity, but let us know if you find out more about this.

PR #82, opened 2019-01-07
Add dutch lang

reinos:
## Overview

I just add, based on the EE5 branch, our language pack to the core of EE. By doing this the language files are not forgotten as now happens. By moving it to the core, we found it better to update the translation.

It should be an ongoing process, as we translate 90%.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [x] Yes
- [ ] No


kevincupp:
Thanks for this, @reinos. What do you mean the files are forgotten?

We're not sure we want every installation to have to have every language, seems like a lot that may not be used, over 16k lines for just this one.

But more importantly, this seems to introduce a workflow issue. Are you suggesting this be the new location for all language packs, and thus Dutch users for example would not have to manually update/install the language? I understand the desire for that, but we would have to wait for translations to be updated before we release. Or if we don't, you would have to wait for the next release at the soonest to get updated translations for the previous release.

I know the current system has more manual work, but it seems best for getting updates into your install quicker. They can be there as soon as the translator has made the changes.

Ultimate goal for language packs, and other add-ons, is to make them installable/updatable from the control panel. I think once we have this working, updating language packs on your sites should be much more of a breeze.

derekjones:
Closed due to inactivity. If you still wish to submit your Dutch language pack (please do!), just have a look at [contributing language packs](https://docs.expressionengine.com/latest/general/languages.html#contributing). Thanks!

PR #95, opened 2019-01-17
Allow mailto: as an optional protocol in URL fields

derekjones:
## Overview

This addresses a feature/bug fix (hence `feaug`) to allow `mailto:` as an optional protocol in URL fields.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/19

### DIFF
--- a/changelogs/patch.rst
+++ b/changelogs/patch.rst
@@ -17,6 +17,7 @@ Patch Release
    - Added <new feature>
    - Fixed Bug (#<issue number>) where <bug behavior>.
 
+- Added the ability for URL fields to accept the ``mailto:`` protocol. Or fixed that bug, depending on your 🏔🔭.
 - Fixed a bug where some member validation language keys may appear unparsed in some third-party contexts.
 - Fixed a potential issue (#76) where some jQuery selectors weren't specific enough.
 - Fixed a bug where some SVGs in File fields would not render a preview on the publish form.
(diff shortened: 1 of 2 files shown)


derekjones:
@ExpressionEngine/maintainers are you okay with this going in a patch release? It's technically additive, but it could also be considered a bug since `mailto:` also commonly finds its way into `href=` which is often the target of URL fields. Even if it's not considered a bug, it doesn't impact any existing fields, and I'd rather make this beneficial change available to users more quickly rather than waiting for a 5.2.0 release.

kevincupp:
I'm personally not picky about small additions going into patch releases.

Curious, what's the reason this would be used over an Email field?

derekjones:
> Curious, what's the reason this would be used over an Email field?

For content that's going to be targeting an `<a href=` that might be a link or an email, like a contact or more info link in a directory where not all of the entities in the directory function identically.

kevincupp:
Ah right, I forgot this field can have multiple allowed protocols. 👍

CreativePoweredGeek:
I saw this a little late...  wouldn't offering `tel: `as an option be beneficial as well?

derekjones:
> I saw this a little late... wouldn't offering tel: as an option be beneficial as well?

Yes, true, mind putting in a feature request or pull request?

Issue #98, opened 2019-01-18
Per-file upload directory/uploader information in File Manager


PaulBaileyAtChan:
**Is your feature request related to a problem? Please describe.**
In a set-up with multiple file upload directories, and where multiple members can upload files, it's often useful to know, per-file, the uploader and upload directory. This information is not currently easy to obtain in the CP File Manager. For example, when listing files by recency, or searching by keyword, it's not possible to see uploader and upload directory information.

Example usage: auditing recent file uploads to make sure file upload directories are sensible; identifying the uploader of a recent file upload where there's a problem that needs to be addressed.

Related to this, it's surprisingly hard to obtain, in the CP, a mapping from the name of a file upload directory to/from its ID. This information is readily available for channels in the (Developer) Channel Manager, but not present in the File Manager. This might be because IDs are considered to be information for developers, not users. But since the File Manager has no presence in Developer, there's nowhere for file upload directory IDs to live in the CP.

Example usage: I see an image in entry content, which has been uploaded to {filedir_X}, and I want to manage that image somehow. How do I know which file upload directory it's in? I might search for the filename, but filenames are not necessarily unique across all directories.

**Describe the solution you'd like**
The obvious analogy is with the Entry Manager, which displays, per-entry:

> — by: [Member Name], in: [Channel Name]

Suggested form for the File Manager would be something like:

> — uploaded by: [Member Name], in: [File Upload Directory name | ID]

Both `uploaded_by_member_id` and `upload_location_id` are available in each row of `exp_files` table.

An alternative to providing the file upload directory ID with each file would be including IDs with upload directory names in the File Manager side-menu.

PR #123, opened 2019-02-11
Add upload directory column in files index page

HopStudios-Development:
## Overview

All files are shown in the Files index page. However, there is no upload directory column. Will be handy to show the upload directory as well.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [x ] 🚀 Implements a new feature
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ x] Yes
- [ ] No

## Documentation


kevincupp:
Pinging @jmathias for feedback here, not sure if he's had any plans in this area.

To summarize, James, when you're viewing all files, there is now an extra column for the upload destination name:

screen shot 2019-02-12 at 4 00 08 pm

The column goes away once you select a folder on the right.

It doesn't work great if the upload destination name is too long, though:

screen shot 2019-02-12 at 4 01 39 pm

Worth mentioning the ideas in this issue which I kind of liked: https://github.com/ExpressionEngine/ExpressionEngine/issues/98

jmathias:
I don't mind the addition of that column, but I don't believe the table UI for the file manager really supports it right now. We need to get the new toolbars in, and I think we'll likely want to use the new table listing as found in the new Entry Manager UI, which will support more columns of data better.

That said I do have plans for improving the file manager, in similar ways as we've started with the Entry Manager, but I'm not in process on that UI currently.

Maybe a compromise for the time being would be to use the current pattern as found in the current shipping Entry Manager.

screen shot 2019-02-12 at 3 18 48 pm

But instead of Author / Channel it could be.

File name / Directory.

`— filename.ext, in: upload_directory_name`

Thoughts?

kevincupp:
I like it. We had a request here #98 to also include the uploader information. Is that piling on too much?

jmathias:
Maybe, but could try it.

nep:
The reason we thought to give it its own column is because of the sortability you get -- It's nice to be able to search for a name and then sort by directory to narrow down the search.  Yes, you can filter by Directory, but that's one at a time and the File Manager is a place where you'd often want to scan for a file -- file name are often not enough to identify something.

Agree that long Directory names aren't pretty.

How about stacking the date/time, which is much wider than I think it needs to be, like so:
2019-11-01
4:38 PM 

Also, the space reserved for the filename seems, in the screenshots, to be extreme, given the average length of most filenames... thoughts?

PaulBaileyAtChan:
A couple of comments from the author of #98:

Totally without considering how much work would be involved, if things were starting from scratch a File Manager modelled more closely on the Entry Manager would make sense to me — moving the management of the directories themselves to Developer, ditching the side-menu and replacing with drop-down filters, including Uploader and Directory, analogous to Author and Channel.

There is definitely an advantage to *seeing* information in a listing, over and above being able to filter on it. It's really useful to see author and channel per-file in the Entry Manager without needing to filter on those, and having something analogous in the File Manager just makes sense to me. There are definitely some screen real-estate constraints currently, but removing the side-menu would help in that direction.

HopStudios-Development:
Additional design tweaks based on feedback from EllisLab. @jmathias the following image shows how the changes will look like :
Screen Shot 2019-05-21 at 6 04 37 PM

obfuscode:
Closing due to age and CP redesign in EE6.

Issue #170, opened 2019-04-03
Imported members - first login sends Password changed email


blue-dreamer:
**Description of the problem**
Unexpected behaviour, you wouldn't expect them to be sent because the password hasn't actually changed.

**How To Reproduce**
1. Import a bunch of members using the XML format, with username, password, screen name and email.
2. When a member logs in the Password changed email is sent 

**Environment Details:**
 - Version: 5.1.2
 - PHP Version 7.2.16
 - MySQL Version 5.0.12 
 - Web Server: Apache w/litespeed

scbarber:
I'm not able to replicate. When the member logs in, are they logging into the CP or the frontend?

blue-dreamer:
This is front end.

Logins were sent to each member manually.
They visit the site and login.
Logging in triggers the email.

scbarber:
I'm still not replicating. The passwords set in the XML file, are they text, md5, or sha1? Do you have any add-ons/extensions that could be interacting with logins?

blue-dreamer:
Passwords in XML were "text" format.

The only addon I'm using is User but that shouldn't affect logging in.

robinsowell:
I think I see why it happens.  Every time someone logs in, the Auth library checks their password and when it does so, it also checks to see if the security of the password can be improved.  So if the password was encrypted using md5 and now sha1 is available, it will change to use sha1.

If it does change the encryption/salt, it uses the member model to update the member record.  And the member model checks to see if the password as saved in the db is different after save.  Which it is, because we changed the encryption.  So it sends an email.

I'm just not sure of the fix.  We can update encryption when we import, which would solve most of the issue for imports.  Still seems a bit odd to email in other cases (i.e., when you have access to a new encryption format so everyone who logs in will get the message).  Maybe a different message or no message based on setting.  But I'm not entirely sure we want to do that.

@scbarber what are you thinking?

robinsowell:
Talked it over with Seth and we pushed a fix that will get the hashing of the text passwords consistent with the Auth library, which should eliminate the emails.

PR #189, opened 2019-04-23
Altered password change notifications related to hash algorithm changes.

robinsowell:
## Overview
Altered notifications when hash alogrithms cause a password hash update.  Related to issue #170 but this is a more general behavior change. That particular bug is already fixed.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/Model/Member/Member.php
+++ b/system/ee/EllisLab/ExpressionEngine/Model/Member/Member.php
@@ -300,8 +300,11 @@ class Member extends ContentModel {
 		{
 			if (isset($changed['password']))
 			{
+				// Did the hash length change? Then the algorithm changed
+				$password_change_type = (strlen($changed['password']) != strlen($this->password)) ? 'member_hash_algo_changed' : 'member_changed_password';
+
 				ee()->logger->log_action(sprintf(
-					lang('member_changed_password'),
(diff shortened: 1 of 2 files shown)


system/ee/EllisLab/ExpressionEngine/Model/Member/Member.php
```diff
@@ -342,8 +345,11 @@ public function onAfterUpdate($changed)
 
 		if (isset($changed['password']))
 		{
-			// email the current email address telling them their password changed
-			$this->notifyOfChanges('password_changed_notification', $this->email);
+			if (strlen($changed['password']) == strlen($this->password))
+			{
+				// email the current email address telling them their password changed
+				$this->notifyOfChanges('password_changed_notification', $this->email);
+			}
```
scbarber: I just caught that we have two `if (isset($changed['password']))` blocks. Would you mind combining them into one? One to log the change the other to conditionally email the member?

robinsowell: Discussed- it's all good, as the log is inside a cp conditional check.

PR #213, opened 2019-05-22
might fix #209

### LINKED ISSUE #209: Password field uses new-password instead of current-password
** Problem **
A password conformation fields have the HTML attribute "new-password" instead of probably "current-password."  Browsers won't use saved password, and some [Safari] try to auto-fill new hash => annoyance.
Example:
/cp/members/profile/email&id=1

**Environment Details:**
 - Version: 5.2.2
 - PHP Version 7.2

**Possible Solution**
field.php 86 <input type="password" name="<?=$field_name?>" value="<?=$value?>" autocomplete="new-password"<?=$attrs?> class="<?=$class?>">

https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_an_html_input_element

bmackenty:
newbie contributor here. This fix made the required changes on my dev setup. I still need to test if it actually changed the thing the original issue referenced (about auto-complete not working). #209

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/View/_shared/form/field.php
+++ b/system/ee/EllisLab/ExpressionEngine/View/_shared/form/field.php
@@ -82,9 +82,16 @@ case 'short-text': ?>
 case 'file': ?>
 	<input type="file" name="<?=$field_name?>"<?=$attrs?> class="<?=$class?>">
 <?php break;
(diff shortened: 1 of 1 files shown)


bmackenty:
There's an additional complication here please see discussion in #209. I'm going to close this pull request until we decide how we want to manage this change request.

PR #214, opened 2019-05-24
Modify the anchor tag to link from add-ons to their manuals 

HopStudios-Development:
## Overview

Modify the anchor tag to link from add-ons to their manuals 

https://github.com/ExpressionEngine/ExpressionEngine/blob/stability/system/ee/EllisLab/ExpressionEngine/Controller/Addons/Addons.php
line 348

## Nature of This Change

- [ x] 🐛 Fixes a bug
- [ x] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x ] Yes
- [ ] No

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/Controller/Addons/Addons.php
+++ b/system/ee/EllisLab/ExpressionEngine/Controller/Addons/Addons.php
@@ -345,7 +345,8 @@ class Addons extends CP_Controller {
 
 						if ($info['manual_external'])
 						{
-							$toolbar['manual']['target'] = '_external';
+							$toolbar['manual']['target'] = '_blank';
+							$toolbar['manual']['rel'] = 'external';
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


derekjones:
Can you describe the issue more plainly? From the diff I'm not immediately seeing what the problem is with the current implementation for add-ons with external docs. Thanks!

HopStudios-Development:
It is a small code change to update the value for the target attribute from _external to _blank as described here https://www.w3.org/TR/WCAG20-TECHS/H83.html

HopStudios-Development:
Currently, links to external docs open with _external (not a valid value for target attr) making tabs overlap when you click in another add-on manual link. With the _blank as a value tabs will not overlap each other.

derekjones:
Yeah that's a bit strange. I imagine (without looking deeply) that the idea is that those links would have rel="external" which we have JS to handle opening in a new tab and not passing referrer data. Odd....

jmathias:
All external links should have a `rel="external"` we shouldn't be using the `target` attribute.

HopStudios-Development:
Nice. We have removed the target attribute as per your comments. Thank you

derekjones:
Thanks @HopStudios-Development!

PR #245, opened 2019-06-25
Added check on select query column aliases

low:
This avoids duplicate fields in the SELECT part (eg. ChannelEntry_channel_title.entry_id AS ChannelEntry__entry_id, ChannelEntry_channel_data.entry_id AS ChannelEntry__entry_id) as well as targeting the non-main table in WHERE statements (eg. ChannelEntry_channel_data.site_id = 1), which could potentially lead to performance issues.

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/Service/Model/Query/Select.php
+++ b/system/ee/EllisLab/ExpressionEngine/Service/Model/Query/Select.php
@@ -202,6 +202,9 @@ class Select extends Query {
 			foreach ($table_fields as $column)
 			{
 				// remember the name so we can translate filters and order_bys
+				// if alias column already exists, skip it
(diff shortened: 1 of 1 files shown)


derekjones:
I'm curious if this adds a new problem when populating related models, that without its own table aliased field in the query data, either a null value or a PHP error results?

low:
@derekjones Not sure when/how this should occur. How would one test this?

derekjones:
Sorry, I must have missed the notification on this. An example would be two models that both have site_id. If you retrieve the model Child, it has `$child->site_id` and the Parent, let's call is `$child->Parent->site_id`. I'm wondering if when accessing the latter—a duplicated column name in a relationship but accessed from separate objects—is populated properly.

obfuscode:
Closing due to lack of activity.

PR #250, opened 2019-07-10
Add hook_id to Channel model

litzinger:
## Overview

The Channel model does not currently publish even hooks when a channel is created, updated, or deleted. As an add-on developer it would be nice to be able to subscribe to these events.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/EllisLab/ExpressionEngine/Model/Channel/Channel.php
+++ b/system/ee/EllisLab/ExpressionEngine/Model/Channel/Channel.php
@@ -20,6 +20,8 @@ class Channel extends StructureModel {
 
 	protected static $_primary_key = 'channel_id';
 	protected static $_table_name = 'channels';
+	
+	protected static $_hook_id = 'channel';
 
 	protected static $_typed_columns = array(
 		'deft_comments'              => 'boolString',


derekjones:
Agreed! Retargeted `5.3.0` as semver prefers additive changes to be in major or minor rather than in patch releases. Simpler for add-ons to target as a minimum requirement if need be as well. Thanks @litzinger!

derekjones:
Do you have time to put in a PR to the docs as well? https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/tree/stability/docs/development/extension-hooks/model

litzinger:
I _might_ request a few more to things... let me know if these are good or bad ideas :)

- ChannelFormSettings (not 100% sure what that does yet)
- ChannelLayout
- Site
- RevisionTracker
- SpecialtyTemplate
- Snippet
- GlobalVariable

I think that would complete the list of things people might need to migrate from one database to another (dev to prod for example).

litzinger:
Yeah, I'll try to update the docs too.

derekjones:
`ChannelFormSettings` stores the Channel's preferences for Channel Form (guest author, etc.) so I agree, that would be good to include as well.

PR #285, opened 2019-08-13
Fix a bug with the :url_slug variable modifier

zaasman:
Fix a bug where the _lowercase_ parameter of the `:url_slug` variable modifier wasn't being applied.

## Overview

This fix modifies the `Formatter\Text` class, specifically the `urlSlug()` method, and simply adds the `get_bool_from_string()` function to the _lowercase_ option check. Without this, the check always fails and the option (ie. the parameter to the `:url_slug` variable modifier) always defaults, rendering the option unusable.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/changelogs/patch.md
+++ b/changelogs/patch.md
@@ -9,6 +9,7 @@ Bullet list below, e.g.
    - Fixed a bug (#<linked issue number>) where <bug behavior>.
 
    - Fixed a bug where the current entry revision did not always show the correct revision author.
+   - Fixed a bug where the lowercase parameter of the :url_slug variable modifier wasn't being applied.
 
 
 EOF MARKER: This line helps prevent merge conflicts when things are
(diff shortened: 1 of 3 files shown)


zaasman:
Thanks @derekjones, good call. Though I've set the test to use 'no', the alternative to the default 'yes', and also what's currently expected.

derekjones:
Thank you and welcome as a [new contributor](https://github.com/ExpressionEngine/ExpressionEngine/blob/stability/AUTHORS.md)!

zaasman:
Thanks!

PR #308, opened 2019-09-09
adding an override for saving template global vars, separate from tem…

adrienne:
## Overview

This adds a simple config override, `save_tmpl_globals`, for template global variables to be saved (or not saved) as files. This is separate from the overall `save_tmpl_files` config override. 

Rationale: in some instances, it may be a bad idea to save global variables as files. It's not always appropriate to treat them as if they're ordinary templates; certain third-party addons (such as Low Variables) add UI such that it may not be obvious that they're going to be saved as files. Additionally, in some MSM setups, people make use of symlinks in the templates directory, which could result in subsites getting the wrong variable data.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/58

### DIFF
--- a/changelogs/minor.rst
+++ b/changelogs/minor.rst
@@ -15,6 +15,7 @@ Minor Release
 
   - Added event hooks to the Channel, ChannelFormSettings, ChannelLayout, Site, Snippet, and Specialty Template models.
   - Fixed a bug ([#306](https://github.com/ExpressionEngine/ExpressionEngine/issues/306)] where {encode} variable output didn't pass the WC3 validator.
+  - Added a config override `save_tmpl_globals` to allow separate saving behavior for global variables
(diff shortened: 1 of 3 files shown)


derekjones:
Thank you @adrienne and welcome as a [new contributor](https://github.com/ExpressionEngine/ExpressionEngine/blob/bfe0735220af712a9f145bd7d9bf6ee020fccaa7/AUTHORS.md)!

PR #339, opened 2019-12-17
Use correct variable name in Filemanager to avoid thumb creation error on Windows

intoeetive:
## Overview

Uploading files was not possible on Windows due to wrong variable name used in legacy Filemanager library for setting the path. This commit is fixing it.

## Nature of This Change

- [x] 🐛 Fixes a bug

- [x] Yes


obfuscode:
Merging without signed CLA as CLA service is temporarily offline and contributor is a staff member.

Issue #248, opened 2019-07-03
The after_channel_entry_save event hook is called twice for every add-on using it


litzinger:
**Description of the problem**
It appears the `after_channel_entry_save` hook is getting called twice for every add-on using it, even if there is only 1 row per add-on in the `exp_extensions` table for the hook. 

**How To Reproduce**
Install an add-on using that hook, step debug, or log to file inside the add-on's method using the hook and you'll see it gets called twice every time an entry is saved. I tested this with 1 add-on installed, then a 2nd add-on, and they both have their `after_channel_entry_save` hook called twice. I also tested this in two different EE environments. One using 5.2.3, and another 5.2.4.

**Environment Details:**
 - Version: 5.2.3 & 5.2.4
 - PHP Version 7.2
 - MySQL Version 5.6
 - OS: macOS
 - Web Server: Apache (vagrant machine)

derekjones:
Curious. Are either of those add-ons calling a new `save()` on the model?

litzinger:
Nope, one of the is Reading Time Pro, and there are no instances of `->save` in its codebase. Publisher does call save on some models, but I commented those calls out while testing and it didn't make a difference... the hook was still called twice.

litzinger:
Ok, I uninstalled Publisher and just tested RTP, and it wasn't called twice. So maybe it was a Publisher issue, but I can't pinpoint it yet.

litzinger:
I take that back. When creating a new entry, the hook in RTP is called twice. When editing its saved once. Publisher is not installed.

derekjones:
Is this occurring with the latest public release of Reading Time Pro so I can replicate or just in dev?

litzinger:
Yes it is.

derekjones:
Can you check to see if you have two rows for that hook and method in your `exp_extensions` table? At the top of your method, I added:

```
file_put_contents('/path/to/file/logit.txt', "\nHere at: ".ee()->localize->now, FILE_APPEND);
```

Watching the file live while saving, it only adds a single line for each save.

derekjones:
Scratch that, reading comprehension. I'm also seeing it on new entries.

dougblackjr:
Added PR #352 to address this one. I'm not sure it's the most elegant solution, so I invite feedback.

PR #352, opened 2020-01-16
Adding check for hooks that should run

dougblackjr:
## Overview

This addresses #248, where the `after_channel_entry_save` extension hook would run twice when saving a new channel entry.

Resolves [#248 ](https://github.com/ExpressionEngine/ExpressionEngine/issues/248).

## Nature of This Change

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

From what I see, this hook was firing twice due to the `save` function in `EllisLab\ExpressionEngine\Service\Model\Model.php`. The `save` function fires the hook immediately upon invoking this function. When creating a new channel entry, the `insert` method is called on the facade. It appears that, when this happens, it runs through `updateEntryStats` in ChannelEntry.php, which then reruns the save function above, but this time updates the entry. The save hooks are being called twice.

The check I added checks against the specific object that is created to see if it is processed; if the save hook has already run, it won't run it again.

I'm not _entirely_ sure this is the most elegant solution, so I gladly welcome any feedback.

### DIFF
--- a/changelogs/patch.md
+++ b/changelogs/patch.md
@@ -5,8 +5,7 @@ ExpressionEngine uses semantic versioning. This file contains changes to Express
 ## Patch Release
 
 Bullet list below, e.g.
-   - Added <new feature>
-   - Fixed a bug (#<linked issue number>) where <bug behavior>.
+- Fixes bug where `after_channel_entry_save` hook would run twice.
 
 
 
--- a/system/ee/EllisLab/ExpressionEngine/Service/Model/Model.php
+++ b/system/ee/EllisLab/ExpressionEngine/Service/Model/Model.php
@@ -68,6 +68,11 @@ class Model extends SerializableEntity implements Subscriber, ValidationAware {
 	protected $_foreign_keys = array();
 
 	/**
+	 * @var check if entry has saved already. Used in hooks
+	 */
+	protected $_has_saved = false;
+
+	/**
 	 * @var Type names and their corresponding classes
 	 */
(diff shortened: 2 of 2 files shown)


litzinger:
I feel like adding a global change like this could have adverse side effects. If the issue is that updateEntryStats is calling `save()` a second time, perhaps adding a parameter to the save function to tell it not to forward the hooks, e.g. `save(false)`, and then just don't call `$this->forwardEventToHooks();` or, add a property to the model, e.g. `$forwardToEventHooks`, and then create a setter and chain the methods in the `updateEntryStats()` method, e.g. `$stats->setForwardToEventHooks(false)->save();`. A change like that would ensure nothing is accidentally broken and makes the action explicit.

dougblackjr:
@litzinger That was my concern as well. My only concern with adding it to the `save` function is that change would be even _more_ global, and have a lot more possible impact. Save exists on the Model class, which is involved just about everywhere. I chose this route because it seemed to be the least impactful, and specific only to this extension hook call. What do you think?

litzinger:
Well, regardless of where it is added, I feel like setting the extra property on the class is safer, b/c it would default to false and require a very explicit call to prevent forwarding to the hooks.

dougblackjr:
Closing in favor of #386

PR #360, opened 2020-01-22
Updates for PHP 7.4 compatibility

rjpdev:
## Overview

***NEEDS TESTING***

This collection updates just a few lines for PHP 7.4 errors/warnings/deprecation notices.

It's basically just a few ternary statements to make sure something exists before it's called (various files), plus little syntax adjustments in one file (mod.channel.php), and a PHP version check for magic_quotes in another (Input.php).

Changed files include:

system/ee/EllisLab/ExpressionEngine/Model/Channel/ChannelLayout.php
system/ee/EllisLab/Addons/channel/mod.channel.php
system/ee/EllisLab/Addons/rte/libraries/Rte_lib.php
system/ee/legacy/libraries/channel_entries_parser/components/Category.php
system/ee/legacy/libraries/Template.php
system/ee/legacy/core/Input.php

## Is this backwards compatible?

Needs proper testing with PHP 5.6 - 7.3.

PR #349, opened 2020-01-16
Fix `strpos` Warning

dougblackjr:
## Overview

Fixes strpos `Non-string needles will be interpreted as strings in the future.` notice in PHP7.3 and higher.

Resolves [#346 ](https://github.com/ExpressionEngine/ExpressionEngine/issues/346).

## Nature of This Change
- [X] 🐛 Fixes a bug
- [ ] 🚀 Implements a new feature
- [ ] 🛁 Refactors existing code
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔥 Removes unused files / code
- [ ] 🔒 Improves security

## Is this backwards compatible?

- [X] Yes
- [ ] No

PR #348, opened 2020-01-16
Bug/344 remove magic quotes

dougblackjr:
## Overview

This addresses #344, and removes the deprecated `get_magic_quotes_gpc` function checks.

Resolves [#344](https://github.com/ExpressionEngine/ExpressionEngine/issues/344).

## Nature of This Change
- [X] 🐛 Fixes a bug
- [ ] 🚀 Implements a new feature
- [ ] 🛁 Refactors existing code
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔥 Removes unused files / code
- [ ] 🔒 Improves security

## Is this backwards compatible?

- [X] Yes
- [ ] No

PR #373, opened 2020-02-18
PHP 7.4 work

dougblackjr:
## Overview

This PR makes EE compatible with PHP7.4, as well as removing deprecated functionality in PHP7+. This goes along with the killer work done in #360, and the other work done in #348 and #349.

The most significant changes are:

1. Removal of Overloaded Functions: MB overloaded functions were entirely deprecated in PHP7.2. This PR adds a helper around those methods, so that those using multibyte setups can still get accurate code.

2. Remove mcrypt: Just pulled it out. This would mean that EE may stop working on PHP5.3 (which I don't think is bad).

3. `implode` glue/pieces switching: This was fully deprecated in 7.4 and should not impact the current app.

4. Validation Rule Names and Namespaces: Some of these were soft reserved PHP keywords.

5. Remove ini.safe_mode: This is fully deprecated since PHP5.4.

6. Lots of little things (Removing `each` functions, fixing curly bracket arrays, etc)

## Nature of This Change

- [X] 🐛 Fixes a bug
- [X] 🛁 Refactors existing code
- [X] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [X] 🔥 Removes unused files / code
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## FULL LIST OF CHANGES HERE: https://www.evernote.com/l/AN8REwo73IJKvJrcjhV7kp3-ItTUFM7Xocs


system/ee/legacy/helpers/multibyte_helper.php
```diff
@@ -13,16 +13,53 @@
  * Multibyte Helpers
  */
 
+if ( ! function_exists( 'ee_get_encoding' ) )
+{
+	function ee_get_encoding($encoding)
+	{
+
+		if (null === $encoding) {
+
```
litzinger: This isn't following EE's coding standard. Although I'd recommend new stuff follow PSR-1, EE's standard would be:

`
	function ee_get_encoding($encoding)
	{
        if (null === $encoding)
        {
            return ee()->config->item('charset') ?: 'utf8';
        }
        ...
    }
`

Basically remove the extra spaces and move the brackets to a new line.

litzinger:
I'd recommend squashing this down into 1 commit. It's all a single unit of work, adding 41 commits to the history is a little messy.

system/ee/EllisLab/ExpressionEngine/Controller/Publish/Edit.php
```diff
@@ -588,6 +589,33 @@ private function remove($entry_ids)
 
 		$entry_names = $entries->all()->pluck('title');
 
+		// Remove pages URIs
+		$site_id = ee()->config->item('site_id');
+		$site_pages = ee()->config->item('site_pages');
+
+		if ($site_pages !== FALSE && $entries)
```
litzinger: This can be condensed into a single conditional

```
if ($site_pages !== FALSE && $entries && count($site_pages[$site_id]) > 0)
{
    foreach ($entries->all() as $entry)
	{
    ...
}
```

system/ee/legacy/libraries/Filemanager.php
```diff
@@ -308,12 +308,17 @@ function get_image_dimensions($file_path)
 		{
 			$D = @getimagesize($file_path);
 
-			$image_size = array(
-				'height'	=> $D['1'],
-				'width'	=> $D['0']
+			if($D && is_array($D)) {
+
+				$image_size = array(
```
litzinger: Can switch to short array syntax

`$image_size = [ ... ]`

system/ee/legacy/libraries/Filemanager.php
```diff
@@ -308,12 +308,17 @@ function get_image_dimensions($file_path)
 		{
 			$D = @getimagesize($file_path);
 
-			$image_size = array(
-				'height'	=> $D['1'],
-				'width'	=> $D['0']
+			if($D && is_array($D)) {
```
litzinger: Since you're making updates to this code it would be a good time to change that variable name from `$D` to something useful 😄

system/ee/legacy/helpers/multibyte_helper.php
```diff
@@ -61,53 +61,16 @@ function ee_get_encoding($encoding)
...
-
-		}
+		$encoding = ee_get_encoding($encoding);
 
-		// If standard encoding, strlen works just fine
-		if ( ! in_array( strtolower($encoding), array( 'utf8', 'utf-8' ) ) ) {
+		if ('CP850' === $encoding || 'ASCII' === $encoding) {
 
-			return strlen( $str );
+			return \strlen($s);
```
litzinger: Single character variable names no bueno.

system/ee/legacy/libraries/Filemanager.php
dougblackjr: I mean, what's more descriptive than `$D`?

One thing I am finding here, that I would appreciate help on, is when this returns false, it is breaking tests. It's almost as if `getimagesize` is failing, though it is a basic PHP function. File is found, path is valid, but getimagesize returns null. Any thoughts?

system/ee/legacy/helpers/multibyte_helper.php
dougblackjr: Updated in latest commits.

system/ee/legacy/libraries/Filemanager.php
dougblackjr: Updated in latest commits.

system/ee/EllisLab/ExpressionEngine/Controller/Publish/Edit.php
dougblackjr: Updated in latest commits.

system/ee/legacy/helpers/multibyte_helper.php
dougblackjr: Updated in latest commits.

dougblackjr:
@litzinger And +1 on the squashing. Too many "Oh God, what have I done" commits in here.

system/ee/EllisLab/Addons/spam/Library/Vectorizers/ASCIIPrintable.php
```diff
@@ -26,17 +26,28 @@ class ASCIIPrintable implements Vectorizer {
...
+
 			$ratio = 1;
+
 		}
+
 		return $ratio;
+
 	}
 
 }
```
obfuscode: Can we remove the extra double spacing?

system/ee/EllisLab/ExpressionEngine/Service/Model/Column/ColumnObject/Timestamp.php
```diff
@@ -8,7 +8,7 @@
  * @license   https://expressionengine.com/license Licensed under Apache License, Version 2.0
  */
 
-namespace EllisLab\ExpressionEngine\Service\Model\Column\Object;
+namespace EllisLab\ExpressionEngine\Service\Model\Column\ColumnObject;
```
obfuscode: What is the reason / benefit for changing this from `Object` to `ColumnObject`?

system/ee/EllisLab/ExpressionEngine/Service/Validation/Rule/ExactLength.php
```diff
@@ -28,12 +31,8 @@ public function validate($key, $value)
 			return FALSE;
 		}
 
-		if (function_exists('mb_strlen'))
-		{
-			return (mb_strlen($value) != $length) ? FALSE : TRUE;
-		}
```
obfuscode: This switched from `!=` to `<` so is the result correct if the `$value` is longer than length? Also, it's a confusing conditional to check for truthiness to return false. I know it was like this originally but shouldn't this just be `return (ee_mb_strlen($value) == $length);` or `return (ee_mb_strlen($value) > $length);` ?

system/ee/EllisLab/ExpressionEngine/Service/Model/Column/ColumnObject/Timestamp.php
dougblackjr: `Object` is a loose protected term in PHP7. Throws a warning now, which means it will be a fatal error in PHP8.

system/ee/EllisLab/ExpressionEngine/Service/Validation/Rule/ExactLength.php
dougblackjr: Good catch, fixed on latest commit.

system/ee/EllisLab/ExpressionEngine/Service/Validation/Rule/MaxLength.php
```diff
@@ -28,12 +31,8 @@ public function validate($key, $value)
 			return FALSE;
 		}
 
-		if (function_exists('mb_strlen'))
-		{
-			return (mb_strlen($value) > $length) ? FALSE : TRUE;
-		}
+		return (ee_mb_strlen($value) < $length) ? true : false;
```
obfuscode: This doesn't follow the style guide for capital true/false values. Also looks like it was changed from `>` to `<` but if you flip the conditional, you'd need it to be `<=` so it includes the total length. Could also be shortened to `return (ee_mb_strlen($value) <= $length);`

system/ee/legacy/helpers/multibyte_helper.php
obfuscode: I prefer the variable order the original way, it prevents accidental assignment `if ($encoding = null)` will not throw an error but is incorrect whereas `if (null = $encoding)` throws a parse error.

system/ee/EllisLab/ExpressionEngine/Service/Validation/Rule/MaxLength.php
litzinger: > This doesn't follow the style guide for capital true/false values

IMO EE should move away from that and move towards PSR standards. EE was originally written before PSR was a thing, now that it is an industry standard EE should use it.

system/ee/legacy/libraries/typography/Markdown/Michelf/Markdown.php
```diff
@@ -797,7 +797,7 @@ protected function _doHeaders_callback_setext($matches) {
 		if ($matches[2] == '-' && preg_match('{^-(?: |$)}', $matches[1]))
 			return $matches[0];
 		
-		$level = $matches[2]{0} == '=' ? 1 : 2;
+		$level = $matches[2][0] == '=' ? 1 : 2;
```
obfuscode: These need to be converted back to `{}` as well.

system/ee/legacy/libraries/typography/Markdown/Michelf/MarkdownExtra.php
```diff
@@ -160,9 +160,9 @@ protected function doExtraAttributes($tag_name, $attr, $defaultIdValue = null, $
 		$attributes = array();
 		$id = false;
 		foreach ($elements as $element) {
-			if ($element{0} == '.') {
+			if ($element[0] == '.') {
```
obfuscode: These need to be converted back to `{}` as well.

system/ee/legacy/libraries/typography/Markdown/Michelf/Markdown.php
dougblackjr: Fixed on latest commit.

system/ee/legacy/libraries/typography/Markdown/Michelf/MarkdownExtra.php
dougblackjr: Fixed on latest commit.

obfuscode:
I didn't want to comment on each line but there are a number of places in the last two files (Markdown, MarkdownExtra) that need the `[]` reverted to `{}`.

system/ee/legacy/libraries/typography/Markdown/Michelf/Markdown.php
dougblackjr: @obfuscode Only issue with this is that this is fully deprecated in version 7.4.

RFC: https://wiki.php.net/rfc/deprecate_curly_braces_array_access

dougblackjr:
Closing in favor of #386

PR #396, opened 2020-04-09
Release of 5.3.1

obfuscode:


system/ee/EllisLab/ExpressionEngine/Core/Core.php
```diff
@@ -289,7 +289,7 @@ protected function getErrorRouting()
 	 */
 	public function setTimeLimit($t)
 	{
-		if (function_exists("set_time_limit") && @ini_get("safe_mode") == 0)
```
intoeetive, line 292: According to PHP docs https://www.php.net/manual/en/function.set-time-limit.php `set_time_limit` has no effect with safe mode on. So this line should be here

system/ee/legacy/libraries/simplepie/SimplePie/File.php
```diff
@@ -109,7 +109,7 @@ public function __construct($url, $timeout = 10, $redirects = 5, $headers = null
 				curl_setopt($fp, CURLOPT_REFERER, $url);
 				curl_setopt($fp, CURLOPT_USERAGENT, $useragent);
 				curl_setopt($fp, CURLOPT_HTTPHEADER, $headers2);
-				if (!ini_get('open_basedir') && !ini_get('safe_mode') && version_compare(SimplePie_Misc::get_curl_version(), '7.15.2', '>='))
```
intoeetive, line 112: The posts on internet are saying `CURLOPT_FOLLOWLOCATION` cannot be used when either `open_basedir` or `safe_mode` are set, but this might have changed? Making the change to keep it consistent with EE5

system/ee/EllisLab/ExpressionEngine/Library/Filesystem/Filesystem.php
```diff
@@ -538,7 +538,7 @@ public function chmod($path, $mode)
 	public function isWritable($path)
 	{
 		// If we're on a Unix server with safe_mode off we call is_writable
-		if (DIRECTORY_SEPARATOR == '/' AND @ini_get("safe_mode") == FALSE)
```
intoeetive, line 541: Based on the comments in this code we should leave `safe_mode` check where it is

PR #398, opened 2020-04-09
Debug/RTE CSS fixes

rjpdev:
**DEBUG**
In the debug area, the .col class is used. This class is a common naming convention for grid and flex layouts nowadays, so I added a few rules to ensure that the debug column isn't affected. Vendor prefixes on all of the debug classes may be the way to go long term.

**RTE**
The pre tag inside of RTE fields was blowing through the container over 2,000 px. I added a rule at the bottom of the file as a quick fix, it could be moved into another location or just piggyback on other rules in the sheet as well.


obfuscode:
@rjpdev Unfortunately, as the EE CP Styles are maintained outside of the open source repo at the moment, we cannot accept PRs for style changes. If you can open an issue with your suggested changes, we can add them on our end and get them merged in.

PR #400, opened 2020-04-10
fix path issues with one-click installer on Windows

intoeetive:
## Overview

Fixes moving installer files into proper locations on windows
Suppressing "access denied" errors that occationally showed up on Windows on cleanup step of install

## Nature of This Change

- [x] 🐛 Fixes a bug

- [x ] Yes


system/ee/EllisLab/ExpressionEngine/Library/Filesystem/Filesystem.php
```diff
@@ -319,7 +319,7 @@ public function rename($source, $dest)
 			throw new FilesystemException("Cannot rename, destination already exists: {$dest}");
 		}
 
-		rename(
+		@rename(
```
obfuscode, line 322: I've dug into the possible effects of silencing this rename and it looks like even if it fails, PHP just returns `false` and only throws a `Warning` if the item trying to be renamed is a directory and that directory already exists.

I'm ok with silencing it if you can add this comment above it so it's clear why it's been silenced:
> // Suppressing potential warning when renaming a directory to one that already exists.

intoeetive: np - done

Issue #399, opened 2020-04-10
After updating to 5.3.1: Error when trying to save entry


bnetz:
**Description of the problem**
I upgraded from 5.3.0 to 5.3.1 . When I create a new entry and try to save it, I cannot save due to an error in the "pages"-tab. I have to enter an URL in the field page-url, then the error vanishes and I can save. But I don't want to enter an URL there, because I don't want to become a blogpost a page.

**How To Reproduce**
Update your installation to 5.3.1

**Screenshot of error**
![200410_error_EE_531](https://user-images.githubusercontent.com/48646646/78981977-63663000-7b21-11ea-8722-a203d7b6efa0.jpg)

bnetz:
stilistical changes in title

robinsowell:
If you have either of those two fields filled in, the other field is required.  Here's the bug the fix addressed: https://github.com/ExpressionEngine/ExpressionEngine/issues/321

I did test to make sure that if I filled in nothing on Pages, I could still save, and I can.  But if either field has something in it, both fields need to have content.  

Or am I misunderstanding what's happening?

bnetz:
What you see in the screenshot is the default-setting (at least for my installation). The default-template for pages is chosen, but the URI-field is empty. This has been for years. When I create a page I enter an URI and this results in the entry being a page; when I don't enter an URI it's a blog post.

When I upgrade to 5.3.1 I get the error; I rolled back to 5.3.0 and everything is fine, see animation attached.

![200410_ee_error-531](https://user-images.githubusercontent.com/48646646/79015597-62f38680-7b6d-11ea-9928-5441bcdbb545.gif)

robinsowell:
AH- once you choose a template, you no longer get the 'Choose wisely' empty option.  And you must have a default template set for that channel.  I don't, so I wasn't running into the problem.

So, we definitely need to get the 'Choose wisely' option to always be available, and I'm wondering if we want to require a template if there's a uri, but NOT require a uri if there's a template.  I can see that one either way.

For the short term, can you switch the channel to not have a default?  And I'll hash out a fix next week.

tr0llx4ls:
Hi,
same problem here.
Don't know how to switch the channel to not have a default. Can you tell us how to do it?
We have seen that if you fill the "Page's URI" field you can continue creating or editing the entry.
Hurry up with the fix please!

bnetz:
[...]

PR #401, opened 2020-04-14
Altered page tab validation to ignore template select when there’s no…

robinsowell:
… page uri set.

Resolves [#399](https://github.com/ExpressionEngine/ExpressionEngine/issues/399).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


intoeetive:
@robinsowell to keep the code style consistent, I suggest moving from anonymous to named function

robinsowell:
Circling back- I couldn't get it working with the named function, talked to Yuri about it and since inline is used elsewhere we decided it's fine as is and just go with it.

PR #424, opened 2020-05-12
Config::site_url was adding a ? to urls when it shouldn’t.

robinsowell:
This mostly showed up using redirect().

## Overview

The forms add-on was set to redirect to force https.  We kept getting a stray ? in the resulting urls (mysite.com/?group/template).  I tracked it down to this.  The function also inconsistently used a deprecated suffix variable.  I dropped that variable.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


system/ee/legacy/core/Config.php
```diff
@@ -1805,8 +1805,7 @@ function site_url($uri = '')
 				$uri = implode('/', $uri);
 			}
 
-			$suffix = ($this->item('url_suffix') == FALSE) ? '' : $this->item('url_suffix');
-			return $this->slash_item('base_url').$this->item('index_page').'?'.trim($uri, '/').$suffix;
+			return reduce_double_slashes($this->slash_item('base_url') . $this->item('index_page') . trim($uri, '/'));
```
obfuscode, line 1808: This fix looks like it removes the ability to use the URL Suffix?

robinsowell: It does, but I'm pretty sure the url_suffix went away.  See https://expressionengine.com/forums/archive/topic/205246/my-url_suffix-is-not-working#958040 though I know that's not where I initially saw it.

robinsowell: OK- here's DJ replying to a bug report about url_suffix not working: https://expressionengine.com/have-the-courage-to-pursue.php?/cp/publish/comments/entry/10989

obfuscode: It's listed in the docs still - so we'll need to remove it from there if it's not a part of EE currently: https://docs.expressionengine.com/latest/development/legacy/helpers/url-helper.html

robinsowell: Cool- I removed from docs- see [pull request 88](https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/88).  Searched docs, that's the only reference that came up.

I also tested.  If I set the config, it had zero effect on the site_url, but interestingly it did allow urls with the suffix to be displayed while without the config, those same urls didn't resolve.

So then I tested with the above change and it didn't affect that behavior, which I take as a good thing.

obfuscode (approved):

Issue #441, opened 2020-06-03
Pass elements through EE template tags to HTML form tags 


TomJaeger:
The idea of adding an ability to pass elements through EE template tags to HTML form tags recently came up in Slack.  There has been a lot of conversation around it in Slack DM's as well as a public threaded conversation. This github issue is intended to bring the conversation fully public and add permanence to it.  This has also been discussed as a public contributed addition to the upcoming 5.4 release :)

The code that would need to be modified is located in the following method.
https://github.com/ExpressionEngine/ExpressionEngine/blob/5db58a8ea12fec2f61427d5375ccbb7e8cf757c6/system/ee/legacy/libraries/Functions.php#L421

We talked about adding a parameter named 'pass_through', and using the ee()->TMPL->fetch_param('pass_through') option to accept a pipe separated list of items that would then be added to the outputted HTML accordingly.

litzinger:
My initial thoughts are there are 2 ways to approach this. 

1) Make the updates necessary in the existing Functions file and don't mess too much with the existing code aside from making whatever changes also work with the hooks that are present in that method.

2) Take the opportunity to refactor this function into a new service. The only dependencies here are the hooks and the $this->get_site_index() call. Of course by dependencies I mean anything that method calls that is external to the method or class entirely. I don't know what it looks like, but it might be possible to make the hooks callbacks in the service, and then pass the site_index value into the service as a constructor arg, thus removing any external dependencies and making the new service unit testable. It's very difficult to unit test any methods that contain references to the ee() super object b/c you have to stub everything out. Possible, but a lot of work. I'd also recommend making a separate view file to build the form tag. Right now it is constructing the html string in the method, so there are no separation of concerns happening.

jcogs-design:
I'm happy to have a go at doing a version of "1" in the above post.  As and when that seems to work, then I suggest at some later date someone migrates that code into the new service option outlined in "2" above.

jcogs-design:
Based on our initial discussion in Slack one aspect that would benefit from a wider perspective is the choice of parameter format to use.  The solution needs to support the addition of multiple additional attributes to the form tag (not including ARIA and event handler global attributes, HTML specifies 38 HTML attributes that can be added to a <form> tag - of which EE currently handles 9).  So far four alternative parameter forms have been proposed:
- everything in a very long string including quoting - for example 
`pass_thru="target='_blank' novalidate name='my form'"`
- everything in a very long string with quoting done by EE - for example
[...]

PR #443, opened 2020-06-06
Feature/pass thru parameters to html forms as attributes

jcogs-design:
## Overview

This fixes a couple of minor errors in original code for the function `form_declaration` in Functions.php

It also adds a new attribute pass-through function.  When called the function inspects the current tagparams and if any of the parameters specified match valid HTML attributes for use in a form it creates these in the opening <FORM> tag including the parameter value as an attribute value if one is provided.  

If one of the parameters specified in the tag matches one that is also assigned because of information passed from calling function the value from the calling function is given priority.  

So, for example, if the calling function specifies a value for an attribute (e.g. `id="form"`) a separately specified parameter within the tag with the same name (e.g. `id="stuff"`) will be ignored.

The only exception to this case is for attribute `class`.  In this case the value of an additional `class` parameter is appended to any value for `class` generated from information provided by the calling function.

So a tag with this construction
```
{exp:search:simple_form 
    channel="articles" 
    url_title="eric" 
    role="main"
    aria-role="form"
    id="testing"
    onstalled="cats"
    form_class="whoop"
    class="it up"
    data-Stuff="and nonsense" 
    hidden
    return="{current_path}"
}
```
will generate this output (reformated to ease legibility)
```
<form 
    id="testing" 
    class="whoop it up"  <--- Note class elements are combined
    method="post" 
    action="https://2gc.jcogs.net/" 
    role="main"               <--- Note these additional parameters 'passed through' to HTML tag
    aria-role="form" 
    onstalled="cats" 
    data-stuff="and nonsense"
>
```
If no value is assigned to the parameter in the EE tag (e.g. `{exp... novalidate}` rather than `{exp... novalidate=""}`) the value is ignored since EE does not pass this element through tagparams.

If a parameter is assigned a null value (e.g. `{exp... novalidate=""}`) it is passed through to the HTML form as a simple token (e.g. `<form ... novalidate>`).

The list of valid attributes for HTML FORM elements combines attributes for [FORM]([https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form]), [HTML Global Attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) and [ARIA Search role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role). 

Resolves [#441](https://github.com/ExpressionEngine/ExpressionEngine/issues/441).

## Nature of This Change

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Functions.php
+++ b/system/ee/legacy/libraries/Functions.php
@@ -423,30 +423,61 @@ class EE_Functions {
 		// Load the form helper
 		ee()->load->helper('form');
 
+        // Load the default values for parameters that can be provided via $data array variable
+        // NB. secure and hidden_fields are not legit HTML form tags, and so excluded from pass_thru 
 		$deft = array(
 			'hidden_fields'	=> array(),
 			'action'		=> '',
-			'id'			=> '',
+            'id'			=> '',
+            'name'          => '',
 			'class'			=> '',
 			'secure'		=> TRUE,
 			'enctype' 		=> '',
 			'onsubmit'		=> '',
-		);
+        );
 
-
+        // Set values for 'missing' default keys in $data to their default values
 		foreach ($deft as $key => $val)
 		{
 			if ( ! isset($data[$key]))
 			{
 				$data[$key] = $val;
 			}
-		}
+        }
+        
+        // If name, id, onsubmit or class attributes are *not set* in the data passed to function by calling function
+        // but *are* specified in tagparams then grab those values now
 
+        if ($data['name'] == '' && ! empty(ee()->TMPL->fetch_param('name')))
+        {
+            $data['name'] = ee()->TMPL->fetch_param('name');
+        }
+
+        if ($data['id'] == '' && ! empty(ee()->TMPL->fetch_param('id')))
+        {
+            $data['id'] = ee()->TMPL->fetch_param('id');
+        }
+
+        if ($data['onsubmit'] == '' && ! empty(ee()->TMPL->fetch_param('onsubmit')))
+        {
+            $data['onsubmit'] = ee()->TMPL->fetch_param('onsubmit');
+        }
+
+        // if class is set in the data passed into function combine with content of any class parameter found in tagparams
+        // is there anything specified in tagparams?
+        if (! empty(ee()->TMPL->fetch_param('class')))
+        { 
+            // append the content of the class parameter to any value we already been passed by the calling function
+            // only add a space before the class content if there is already something in $data['class'] 
+            $data['class'] .= ($data['class'] == '' ? ee()->TMPL->fetch_param('class') : ' '.ee()->TMPL->fetch_param('class'));
+        }
+
+        // If not specified, put site_id value into hidden_fields array
 		if (is_array($data['hidden_fields']) && ! isset($data['hidden_fields']['site_id']))
 		{
 			$data['hidden_fields']['site_id'] = ee()->config->item('site_id');
-		}
-
+        }
+        
 		// -------------------------------------------
 		// 'form_declaration_modify_data' hook.
 		//  - Modify the $data parameters before they are processed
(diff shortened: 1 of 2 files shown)


system/ee/legacy/libraries/Functions.php
```diff
@@ -471,33 +502,79 @@ public function form_declaration($data)
...
+        // data-* or aria-* as these are processed differently).  The list taken from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form on 5/june/20
+
+        $valid_html_attributes = explode('|','accept|accept-charset|accesskey|autocapitalize|autocomplete|contenteditable|contextmenu|dir|draggable|dropzone|exportparts|hidden|inputmode|is|itemid|itemprop|itemref|itemscope|itemtype|lang|method|novalidate|onabort|onautocomplete|onautocompleteerror|onblur|oncancel|oncanplay|oncanplaythrough|onchange|onclick|onclose|oncontextmenu|oncuechange|ondblclick|ondrag|ondragend|ondragenter|ondragexit|ondragleave|ondragover|ondragstart|ondrop|ondurationchange|onemptied|onended|onerror|onfocus|oninput|oninvalid|onkeydown|onkeypress|onkeyup|onload|onloadeddata|onloadedmetadata|onloadstart|onmousedown|onmouseenter|onmouseleave|onmousemove|onmouseout|onmouseover|onmouseup|onmousewheel|onpause|onplay|onplaying|onprogress|onratechange|onreset|onresize|onscroll|onseeked|onseeking|onselect|onshow|onsort|onstalled|onsubmit|onsuspend|ontimeupdate|ontoggle|onvolumechange|onwaiting|part|rel|slot|spellcheck|style|tabindex|target|title|translate|role');
```
litzinger, line 542: This means EE is maintaining a list of valid values, and a user can't pass any custom values through. Even if EE went with this approach, this list should be an array in a config file somewhere, not a scalar value buried deep within EE's core.

jcogs-design:
It works. It is better than the code it replaces. I don’t see the case for supporting non HTML attributes even if there was a way to do this: I ree EE lies your symfony class idea could one day allow for arbitrary attributes but it would be a huge rewrite of the EE code Badr to get that to work, and until then nothing.

This is for code in the “legacy” library.  Elsewhere in those libraries EE has hardwired options for the image formats it works with so it is not like the rest sets standards. Which is perhaps why it is called the legacy libraries. 

Maybe you should work on your symfony rewrite for v6.1 and we move the function to your class at same time. 

Or, if you can see a way to improve what I have done do so. That after all is why we have collaborative development systems for open source. 

I was asked if I could make a contribution, here it is. You and your colleagues don’t like it?  Fix it or bin it: it is your collective call. But in the interim If EE continues to be unable to support other attributes when it could have gone do I think that would be a waste.

litzinger:
> Or, if you can see a way to improve what I have done do so.

Sorry if I wasn't clear, but that's what I was doing. If we want to take the predefined list approach, which is fine, then I think the list should be and array in the config files, similar to the mime types, not a scalar value like this. No approach was decided on (predefined list or totally free-form), just ideas being thrown around, and both have merits.

system/ee/legacy/libraries/Functions.php
```diff
@@ -471,33 +502,79 @@ public function form_declaration($data)
...
+            foreach ($_tagparams as $key => $val) {
+            
+                // Ignore the parameter if $key is defined in $deft (and so already being processed by function)
+                // or if the parameter is not in the list of approved attributes
+                // or if the parameter begins with either aria- or data-
+                if ((! array_key_exists($key, $deft) && in_array($key, $valid_html_attributes)) || (preg_match("/^data-|^aria-/",$key))) {
+        
+                    // Append the key to end of the $_pass_thru variable
+                    // only add a leading space if the variable already has some content
+                    $_pass_thru .= ($_pass_thru == '' ? $key : ' '.$key);
```
litzinger, line 558: Instead of building a string here it would probably be cleaner to add the key/value pair to an array, then at the end of the loop [assemble the string of attributes](https://stackoverflow.com/questions/18081625/how-do-i-map-an-associative-array-to-html-element-attributes#18081767).

litzinger:
This isn't something you necessarily have to do, but it would be nice if this could be unit tested if it is merged into the code base. Aside from those `ee()->extensions` references everything is easily testable. @obfuscode is there a separate test suite that we don't know about? If so are there already mocks for the `ee()` function?

system/ee/legacy/libraries/Functions.php
jcogs-design: Not sure I understand why creating an array with these values and then turning that array back into an string prior to appending the string to the other strings would be an improvement.  

The only purpose of the activity is to insert a string into the `<form>` tag that is being assembled from other strings.  We don't have any need to access or reorder the values within the string we are assembling.

Perhaps you could explain what I am missing here. Is there some performance benefit?

jcogs-design:
> In general this isn't PSR-12, even if it is a legacy library I think the goal is that all new code, or refactored code is PSR-12. https://expressionengine.com/blog/expressionengine-adopts-psr-12

Just to be clear, are you saying the code I've added needs to be PSR-12 or you want the whole of Functions.php to be PSR-12?

jcogs-design:
> This isn't something you necessarily have to do, but it would be nice if this could be unit tested if it is merged into the code base. Aside from those `ee()->extensions` references everything is easily testable. @obfuscode is there a separate test suite that we don't know about? If so are there already mocks for the `ee()` function?

I'd be happy to add something to existing tests if there are any.  Less comfortable creating new ones (not familiar with the EE testing approach, so likely to do the wrong thing and so better to not break things unnecessarily).

system/ee/legacy/libraries/Functions.php
jcogs-design: Have pushed an update that moves the array to a config file.

system/ee/legacy/libraries/Functions.php
```diff
@@ -535,11 +535,13 @@ public function form_declaration($data)
...
+        /**
+         * Valid HTML Form attributes are determined based on the list in
+         * on the config/valid_html_form_attributes.php file .
+         */
+
+        $valid_html_attributes = ee()->config->loadFile('valid_html_form_attributes');
```
litzinger, line 544: Should probably check to make sure the array is not empty, and in fact it is an array, before trying to use it in the in_array() call below.

system/ee/EllisLab/ExpressionEngine/Config/valid_html_form_attributes.php
```diff
@@ -0,0 +1,124 @@
...
+ * (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) and
+ * the ARIA Search role attribute
+ * (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role). 
+ * The list last updated on 5 june 2020
+ * 
+ * This list does not include data-* or aria-* as these are wildcards and
+ * need to be handled differently in the code.
+ *
+ */
+return array(
```
litzinger, line 26: I'd recommend short array syntax.

There are some line spacing issues too.

litzinger:
> > In general this isn't PSR-12, even if it is a legacy library I think the goal is that all new code, or refactored code is PSR-12. https://expressionengine.com/blog/expressionengine-adopts-psr-12
> 
> Just to be clear, are you saying the code I've added needs to be PSR-12 or you want the whole of Functions.php to be PSR-12?

I think the whole function should be, it's short, and you're making enough modifications to it that you might as well update the whole thing.

system/ee/legacy/libraries/Functions.php
litzinger: Using an array is just better programming practices than using a long string like that. It's not a performance issue. It's easier to modify an array to add or subtract elements than it is a string. If unit tests are added it'll be a little easier to write the tests etc

jcogs-design: Sorry - I simply don't buy the concept.  

Two functional lines later the whole lot is concatenated into a long string. 

So to traverse one line of code we build an array and then unbuild it back into a concatenated string.  I can think of no circumstances where in the processing of that intervening line anyone will need to modify or add / subtract elements.  

We literally don't care what they are: we know that they are legitimate to pass through, and that some may have values set.  

I am intrigued to know what circumstance you have in mind for the modify / add / subtract use case.

With regard to unit testing - as noted elsewhere I don't have any familiarity with what testing regime you are using, and so cannot comment.  But a testing regime that only tests array values doesn't sound right to me.  Bear in mind we are simply collecting up some values and dumping them out into an HTML tag.  I'm fascinated to know what kind of test you have in mind, and how the arrays fit in / make it easier.

system/ee/legacy/libraries/Functions.php
jcogs-design: $valid_html_attributes defined as an array on line 537.  Added check for non-empty.

system/ee/EllisLab/ExpressionEngine/Config/valid_html_form_attributes.php
jcogs-design: Uses identical format to captcha.php, mimes.php and stopwords.php in the same folder.
Not sure what the line-spacing issues you refer to are - if you clarify maybe they can be fixed.

jcogs-design:
> This isn't something you necessarily have to do, but it would be nice if this could be unit tested if it is merged into the code base. Aside from those `ee()->extensions` references everything is easily testable. @obfuscode is there a separate test suite that we don't know about? If so are there already mocks for the `ee()` function?

Maybe I am missing something, but I could not find anywhere in the **tests** section of repo where there are any tests on the output of this function.  I have not yet attempted to get deep into it, but cursory review of current tests suggests they are heavily focused on testing CP rather than site output.  These proposed edits could affect both CP and public HTML but are primarily intended to address need for inseration of attributes into public HTML; unclear how testing of such is handled currently and so how a test if it were to be created would be integrated into the testing setup.  

But as per earlier comment, have not yet found out where current testing regime is documented, so maybe once I get to that I'll be clearer.

In the interim, any guidance that can be offered on this would be helpful.

jcogs-design:
> Can you tell where the list of attributes you used was taken from? It does not seem to match current HTML specs https://html.spec.whatwg.org/multipage/forms.html#the-form-element and https://html.spec.whatwg.org/multipage/dom.html#global-attributes

As documented in code addition, the sources used were: 

> + * HTML Form Attributes
> + * 
> + * Valid HTML attributes that might be specified in a FORM tag 
> + * (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) plus
> + * HTML Global Attributes 
> + * (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) and
> + * the ARIA Search role attribute
> + * (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role). 
> + * The list last updated on 5 june 2020
> + * 
> + * This list does not include data-* or aria-* as these are wildcards and
> + * need to be handled differently in the code.
> + *

If you would prefer it to use a diferent source, that's fine with me.

intoeetive:
Oh! I should have looked at the comments :)

Thank you! I'll have to go through the code and make some updates and then will get back to you to see if that still reflects your idea

jcogs-design:
Bump? Or just delete? Seems pointless to have had it in limbo for so long (almost 2 years).

intoeetive:
@jcogs-design it's been a real while, but I took your code, made some extra updates and opened PR for EE7 in https://github.com/ExpressionEngine/ExpressionEngine/pull/3475

Took me less than 3 years ;)

PR #446, opened 2020-06-11
Added support for Cloudflare forwarding IP headers

litzinger:
This PR was inspired by an article found on dev.to https://dev.to/rogeriotaques/an-easy-way-to-get-the-real-client-ip-in-php-4pii

## Overview

This PR isn't based on any personal experience with getting IP addresses from behind CF, but I suspect this might be a helpful addition to the codebase. I also updated all references of '0.0.0.0' to use a constant. EE uses too many repeated scalar values like this this throughout the codebase, and they should be cleaned up. This is a start.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [x] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

I don't think this requires a documentation update, but I could be wrong.

### DIFF
--- a/system/ee/EllisLab/Addons/forum/mod.forum_core.php
+++ b/system/ee/EllisLab/Addons/forum/mod.forum_core.php
@@ -5524,7 +5524,7 @@ class Forum_Core extends Forum {
 	 	// Is the IP or User Agent unavalable?
 		if (ee()->config->item('require_ip_for_posting') == 'y')
 		{
-			if (ee()->input->ip_address() == '0.0.0.0' OR ee()->session->userdata['user_agent'] == "")
+			if (ee()->input->ip_address() == EE_Input::DEFAULT_IP_ADDRESS OR ee()->session->userdata['user_agent'] == "")
 			{
 				return $this->trigger_error();
 			}
--- a/system/ee/legacy/core/Input.php
+++ b/system/ee/legacy/core/Input.php
@@ -13,6 +13,8 @@
  */
 class EE_Input {
 
+	const DEFAULT_IP_ADDRESS = '0.0.0.0';
+
 	var $SID = ''; // Session ID extracted from the URI segments
 
 	var $ip_address				= FALSE;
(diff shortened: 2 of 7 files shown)


system/ee/legacy/core/Input.php
```diff
@@ -374,55 +376,56 @@ function server($index = '', $xss_clean = FALSE)
...
+			'HTTP_FORWARDED_FOR',
+			'HTTP_CLIENT_IP',
+			'HTTP_X_CLIENT_IP',
+			'HTTP_X_CLUSTER_CLIENT_IP',
+		];
+
 		$proxy_ips = config_item('proxy_ips');
-		if ( ! empty($proxy_ips))
-		{
+		$this->ip_address = $_SERVER['REMOTE_ADDR'];
```
litzinger, line 399: Changed the logic by moving this up and setting the default. There was an unnecessary `else` in the conditional below.

system/ee/EllisLab/Addons/forum/mod.forum_core.php
```diff
@@ -5524,7 +5524,7 @@ public function submit_post()
 	 	// Is the IP or User Agent unavalable?
 		if (ee()->config->item('require_ip_for_posting') == 'y')
 		{
-			if (ee()->input->ip_address() == '0.0.0.0' OR ee()->session->userdata['user_agent'] == "")
+			if (ee()->input->ip_address() == EE_Input::DEFAULT_IP_ADDRESS OR ee()->session->userdata['user_agent'] == "")
```
obfuscode, line 5527: There's no `use EE_Input` statement for this file, is this an oversight or is there a different way it's invoked?

system/ee/legacy/libraries/Messages_send.php
```diff
@@ -387,7 +387,7 @@ function send_message()
 		/** ----------------------------------------*/
 		if (ee()->config->item('require_ip_for_posting') == 'y')
 		{
-			if (ee()->input->ip_address() == '0.0.0.0' OR ee()->session->userdata['user_agent'] == '')
+			if (ee()->input->ip_address() == EE_Input::DEFAULT_IP_ADDRESS OR ee()->session->userdata['user_agent'] == '')
```
obfuscode, line 390: Similar to forum_core, does this file need the `use EE_Input` statement?

system/ee/legacy/libraries/Throttling.php
```diff
@@ -58,8 +58,7 @@ function run()
 
  	function throttle_ip_check()
  	{
-		if (ee()->config->item('banish_masked_ips') == 'y' AND ee()->input->ip_address() == '0.0.0.0' OR ee()->input->ip_address() == '')
-		{
+		if (ee()->config->item('banish_masked_ips') == 'y' && ee()->input->ip_address() == EE_Input::DEFAULT_IP_ADDRESS || ee()->input->ip_address() == '') {
```
obfuscode, line 61: Similar to forum_core, does this file need the `use EE_Input` statement?

system/ee/EllisLab/Addons/forum/mod.forum_core.php
litzinger: Probably oversight

TomJaeger:
Thinking on this PR, I'm inclined to hold off for now (the 5.4 release we're merging up), and cross the bridge in EE 6.  I think it brings up a good point specifically relating to scalar values in the code base that probably warrant a bit more conversation as a whole.

Changing PR to Draft for now.

robsonsobral:
Please, consider it again.

dougblackjr:
Hey all. Any update on this? We have a situation where we are running an EE site through a containerized reverse proxy, which catches the IP address as a 172.18.*.* address, a private IP address. We also have a bunch of sites on CloudFlare that could use exactly this. Unfortunately, this is far from ideal. Having something like this, or a config option where we can add additional headers to check, would be great.

intoeetive:
@litzinger - thank you for your work here. Since this is for v5 I'm going to close it, instead I've opened PR for v7: https://github.com/ExpressionEngine/ExpressionEngine/pull/3470

@dougblackjr and everyone - perhaps you could give that PR a test in real life to see if it works well

PR #447, opened 2020-06-11
Altered a js filename that mod_security flagged often.

robinsowell:
## Overview

I renamed a javascript file that tended to trigger a mod_security filter, breaking the cp whenever a file field was on it.  Had one person who ran into the issue test it out, this did solve it.

I did not delete the old javascript file as that should wait for at least 6 just to be on the safe side. 
 I could put log a deprecation notification in combo_load(), but don't love the overhead.  IDK?  

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


obfuscode:
It looks like this file is a duplicate of the `cp.js` but do we need the `cp.js` file around anymore or should it be removed?

robinsowell:
I don't think we need it, but just in case a third party was using it, I didn't remove it.  I would tend to wait and maybe remove it in v6 just to avoid any potential problem.

obfuscode (approved):

PR #452, opened 2020-06-26
Add support for SameSite cookies

litzinger:
## Overview

• https://web.dev/samesite-cookies-explained/
• Default to Lax, which as the article explains, is how cookies are treated if no SameSite value is set.
• Ensure users can’t set an invalid value as SameSite
• If they set the value to ‘None’, make sure the cookie is secure.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/83


system/ee/legacy/core/Input.php
```diff
@@ -208,11 +208,37 @@ protected function _set_cookie(array $data)
...
+		$data['httponly'] = get_bool_from_string($data['httponly']);
 
 		// Deal with secure cookies.
-		$data['secure_cookie'] = (bool_config_item('cookie_secure') === TRUE) ? 1 : 0;
+		$data['secure_cookie'] = bool_config_item('cookie_secure');
+
+		// SameSite explained: https://web.dev/samesite-cookies-explained/
+		// Allow default setting for all site cookies, then allow for overrides when this method is called.
+		// Cookies without a SameSite attribute will be treated as SameSite=Lax.
+		$sameSiteConfigValue = ee()->config->item('cookie_samesite');
```
litzinger, line 220: I'm honestly not sure what the behavior should be here. Should the config value set the site wide default setting of the SameSite cookie, or should it override _all_ cookie settings, regardless of what EE or a 3rd party add-on might want to set their cookies as? Part of me thinks this should change to just default all cookie to Lax, then move the cookie_samesite config override to the very end, to let site admins override settings to increase security, and potentially override what a 3rd party add-on, or even EE's own cookies are doing.

jamfwww:
It looks like for this to work in PHP <7.3, an additional change needs to be made: https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict

I can make that change when I get some direction on if the config value should be the default, or be the overriding value of all cookies set with this function (I'm leaning towards the later).

litzinger:
It looks like for this to work in PHP <7.3, an additional change needs to be made: https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict

I can make that change when I get some direction on if the config value should be the default, or be the overriding value of all cookies set with this function (I'm leaning towards the later).

system/ee/legacy/core/Input.php
obfuscode: I think the way you have it is fine. Default to lax or whatever level the site admin sets and then anyone setting a cookie can override it. As this option didn't exist until now, there shouldn't be any add-ons that accidentally set it to a value so any use of it going forward would be intentional on the dev's part and we should allow that. If anyone has any objections or concerns to this approach, please speak up.

litzinger: The reason I was thinking of moving the config option to the end is that if an add-on does end up setting a cookie to Lax, but then for security reasons, maybe the actual site maintainer wants to force everything to Strict.

We could take it a step further and add some sort of per cookie override. Either a hook every time a cookie is attempted to be set, or a config array to set the value for each array, e.g.

```
$config['cookie_samesite_overrides'] = [
    'publisher_site_language' => 'Strict',
    'untrusted_addon_cookie' => 'Strict',
    ...
];
```

litzinger: Oh wait, there is already a hook... `set_cookie_end`

litzinger: In that case, then I guess my current implementation is OK. If someone, a power user, or someone who want's to ensure all their cookies are perfectly set, can use the hook to change the value that a 3rd party add-on might set.

obfuscode: We could add another config override to override any other cookies set in the hook:
`ee()->config->item('enforce_addon_cookie_samesite');` or some better name I can't think of at the moment.

system/ee/legacy/core/Input.php
```diff
@@ -259,6 +259,18 @@ protected function _set_cookie(array $data)
 		/*
 		/* -------------------------------------------*/
 
+		if (PHP_VERSION_ID < 70300) {
+			// Older versions of PHP do not support an array as the 3rd parameter,
+			// thus the SameSite setting must be hacked in with the path option.
+			return setcookie($data['prefix'].$data['name'], $data['value'],
```
litzinger, line 265: For reference https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict

system/ee/legacy/core/Input.php
obfuscode: We discussed it internally and decided we should do it the way you suggested in your first comment:

- No config, not specified by add-on = default to LAX
- No config, specified by add-on = add-on setting
- Config, specified by add-on = config override (ignores add-on setting)

Would you be willing to update your PR to make the config override what is set by the add-on?

ignetic:
While Chrome have begun enforcing SameSite=Lax as their default cookie setting, I found that this has been causing issues with certain payment systems. Forcing this cookie to SameSite=None would normally solve this issue for Chrome:
https://www.chromium.org/updates/same-site

It is worth noting that this solution can cause problems in other browsers that are incompatible with SameSite=None cookies, or as with Safari 12.1 where it has a bug that reverts to SameSite=Strict if used:
https://www.chromium.org/updates/same-site/incompatible-clients

obfuscode:
Is there anything else you're waiting on us for before this is reviewed further? @litzinger What do you think about @ignetic's proposal to change it to `SameSite=None`?

ignetic:
> Is there anything else you're waiting on us for before this is reviewed further? @litzinger What do you think about @ignetic's proposal to change it to `SameSite=None`? I believe we would want to stick with `SameSite=Lax` as the issues raised by `None` would be too great in my opinion for the majority of clients.

@obfuscode I think I may have confused the issue here. Let me try to explain it better:

In my experience, I have needed to force cookies as SameSite=None where Chrome has set the default to SameSite=Lax. SameSite=None cookies are required to be used cross-site. This is where I have experienced issues with the cookies not being read when returning back to the site from an offsite payment where they remain Lax. Forcing this particular cookie as SameSite=None resolved this this issue.

In my previous comment, I was highlighting the issue regarding Chrome where Google making the browser use SameSame=Lax by default - just so that we are aware of this and that not all browsers support certain SameSite parameters. Note that SameSame=None must also to be secure.

Saying that, I do agree that SameSite=Lax should be the default and also think it would be a good feature to allow control over individual cookies where the SameSite parameter can be applied to them (or overridden). 

So far, I managed to create a working addon using the ideas that litzinger mentioned with using the `set_cookie_end` hook. This enables you to override SameSite cookie parameters: https://github.com/ignetic/ee-samesite-cookies

obfuscode (approved):

PR #472, opened 2020-07-17
Increase Model Saving Speed by Adding Config Override to Ignore Channel Stats

dougblackjr:
## Overview

The current model set up becomes very slow, especially around channels and channel entries. This doesn't make a huge impact for some, but for those who are doing big imports and using channels with a lot of fields, this causes speed issues.

The majority of this issue appears to be in how stats are are processed when channels/entries are created or updated. The stats calls within the models here adds between 8 - 12 database calls.

This PR adds two things:
1. A config override (`ignore_entry_stats`) to stop the saving of stats during create/update calls.
2. An action endpoint for the Stats module that will process stats when called. This will allow for stats to stay up to date, if the user would like to do so, without causing issues during the save process.

## Nature of This Change

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No


litzinger:
Doug, I _think_ if you're seeing Jace's commits in the PR it means that you didn't rebase correctly. Try rebasing your changes with the latest stability branch.

litzinger:
FWIW I think this is an interesting PR. I've never used EE's entry stats (TBH I think it's outdated when you can use something like GA), so having an option to disable the stats for increased performance is a plus for me.

obfuscode (approved):

Issue #379, opened 2020-03-02
Comment module file missing?


iammattmartin:
When trying to use the "unsubscribe" link in emails for comment updates, we've noticed this generates blank pages and does not remove the email.

Turning on debug we get:

```
Fatal error: require_once(): Failed opening required '/var/www/xxxx/system/ee/EllisLab/Addons/comment/mcp.comment.php' (include_path='.:/usr/share/php') in /var/www/xxxx/system/ee/EllisLab/ExpressionEngine/Service/Addon/Addon.php on line 688
```

On checking, that files does not exist, as the error does indicate. Also checking the latest release it is missing too.

Is this an oversight and it should be there or another issue entirely?

iammattmartin:
Finding an old version of the file at https://raw.githubusercontent.com/ExpressionEngine/ExpressionEngine/43882902677986751066048580c9045deaade693/system/expressionengine/modules/comment/mcp.comment.php and testing that, seems to now work.

dougblackjr:
@robinsowell Do you have context as to why this was removed? I can pull this back in to the PHP7.4 branch, or adjust accordingly.

robinsowell:
@dougblackjr I'm not sure, though I suspect it was part of moving the comment module more into 'core' and not an optional module.  Something similar was done with Channels.  Both have always been integrated into a lot of the core controllers displays.

Instead of managing comments in a comment control panel, it was moved into core.  But it looks like we lost some functionality doing it.

PR #478, opened 2020-07-27
Bug/comment unsubscribe

robinsowell:
## Overview

 Fixes a bug where comment subscription emails contained an invalid unsubscribe link and added some system error messages for invalid attempts to unsubscribe.

Resolves [#379](https://github.com/ExpressionEngine/ExpressionEngine/issues/379).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


system/ee/EllisLab/Addons/comment/mod.comment.php
```diff
@@ -2251,23 +2258,38 @@ function comment_subscribe()
 		}
 
 		// They check out- let them through
-		ee()->subscription->$type();
+		if ($type == 'unsubscribe') {
+			ee()->subscription->$type(FALSE, $hash);
```
obfuscode, line 2262: According to a comment above, you must have the `$hash` if you are not logged in - which means you can be logged in and NOT have the hash, yet this line requires it - so do we need error checking around if a user is logged in but for some reason doesn't have a `$hash` in their URL (which would bypass the checks on 2215-2224)?

robinsowell:
Bugger- I'll poke as soon as I wrap up this other thing.  Thanks for the catch.  They shouldn't have to have the hash as far as I recall.

system/ee/EllisLab/Addons/comment/mod.comment.php
robinsowell: I think it's ok- $hash can be false and if it's not set in get, it should be false.

In legacy/libraries/Subscription.php:

```	function unsubscribe($identifiers = FALSE, $hash = FALSE)
	{
		if ($hash != '')
		{
			ee()->db->where('hash', $hash);
		}
		else
		{
			$user = $this->_prep($identifiers);

```
So if $hash is false, it would be like just doing ee()->subscription->$type(); which was what the old code was doing.

Was I looking in the right spot re: the question?

obfuscode (approved):

Issue #499, opened 2020-08-25
Hide a category group in a layout and lose stored data


robinsowell:
**Description of the problem**
If you had a category group via a layout, saving the entry can lose any stored category data.

**How To Reproduce**
Create 2 category groups.  Go make an entry.  Pick a category or two in each of the groups.

Then create a layout and hide one of the groups- leave the other one showing.

Edit your entry, resaving it.  Now the category group that was hidden has no categories selected.  You've lost that data.

Note- if you only have 1 category group, you can't replicate the issue.  It's having 2 and hiding one that's a problem.  I'm pretty sure the check on visibility isn't taking into account that you can have multiple groups, some showing and others not.

**Environment Details:**
 - Version: [e.g. 5.3.2]
 - PHP Version [e.g. 7.1]

Harold-git:
Hi Robin,
I've upgraded from 5.3 to 5.4 especially to be able to hide categories in the layout.
The issue is that members added specific categories to entries and we want to keep the category added to it. But we don't want them to tag new entries to those categories so we would like hide those categories.
Only when clicking on 'Hide' or 'Collapse' it still won't stick when saving to the layout.
Best regards,
Harold

PR #502, opened 2020-08-27
Bug/loosing categories hidden in layout

intoeetive:
## Overview

Fixed a bug where categories hidden from channel layout might get lost upon saving the entry

Resolves [#499](https://github.com/ExpressionEngine/ExpressionEngine/issues/499).

## Nature of This Change

- [x] 🐛 Fixes a bug

- [x] Yes

## Documentation


robinsowell:
Late to the party, but user just confirmed the patch works.

robinsowell:
And just confirming, I did some tests as well.

2 cat groups, put content in each.  Hide one group.  Edit.  In the visible group, uncheck all cats.  Save.  They remain unchecked, the hidden one retains content.  Show both, uncheck the remaining one.  Save- the save as unchecked.

I couldn't break it.

PR #504, opened 2020-08-28
Bug/like query construction when search in content

intoeetive:
## Overview

Fixed a bug where searching entries in CP in content only could produce SQL error

## Nature of This Change

- [x] 🐛 Fixes a bug

- [x] Yes


robinsowell:
Thanks!  This tested out fine on the user's site.

PR #530, opened 2020-09-22
Bug/pre 5.4 fixes

intoeetive:
Fixes that were in `cypress-testing` branch

Affecting installer/updater process for compatibility with post-upgrade advisor feature.

### DIFF
--- a/system/ee/installer/controllers/wizard.php
+++ b/system/ee/installer/controllers/wizard.php
@@ -13,7 +13,7 @@
  */
 class Wizard extends CI_Controller {
 
-	public $version           = '5.3.2';	// The version being installed
+	public $version           = '5.4.0';	// The version being installed
 	public $installed_version = ''; 		// The version the user is currently running (assuming they are running EE)
 	public $schema            = NULL;		// This will contain the schema object with our queries
 	public $languages         = array(); 	// Available languages the installer supports (set dynamically based on what is in the "languages" folder)
@@ -253,6 +253,8 @@ class Wizard extends CI_Controller {
 		{
 			if ($this->is_installed)
 			{
+				//remove the update notices from previous installations
+				$this->update_notices->clear();
 				return $this->update_form();
 			}
 			else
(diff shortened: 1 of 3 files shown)


system/ee/installer/controllers/wizard.php
```diff
@@ -2249,7 +2249,7 @@ public function canRenameAutomatically()
 	 * Rename the installer
 	 * @return void
 	 */
-	private function rename_installer()
+	private function rename_installer($template_variables)
 	{
 		if (TRUE || ! $this->canRenameAutomatically())
```
TomJaeger, line 2254: @intoeetive I believe this will always pass since it's TRUE OR?

system/ee/installer/views/success.php
```diff
@@ -1,4 +1,4 @@
-<div class="box <?=empty($update_notices) ? 'success' : 'warn'?>">
+<div class="box <?=empty($update_notices) || count($update_notices)==1 ? 'success' : 'warn'?>">
```
TomJaeger, line 1: @intoeetive should this be greater then or equal >= 1 ?

intoeetive: It should not. We DO WANT to apply success style if there's just one notice, otherwise we apply 'warn' style

system/ee/installer/controllers/wizard.php
intoeetive: Good catch! I should put more attention into what I commit :(

system/ee/installer/views/success.php
TomJaeger: What if there are 2 update notices?

intoeetive: Previously, "please delete your install folder" was not part of `$update_notices` so the message was always shown as success. It is now part of `$update_notices`, so we want to make sure the screen is still green if there is just one message. If there's something more, we apply 'warn' style to indicate there are more thing to look into - just the way it was before

TomJaeger: Ah, ok cool, I think I follow now.

obfuscode (approved):

PR #633, opened 2020-11-12
Major version update confirmation

obfuscode:
## Overview

Adds a password confirmation to the one-click update for major versions.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [X] Yes
- [ ] No


system/ee/EllisLab/ExpressionEngine/View/_shared/footer.php
```diff
@@ -6,20 +6,22 @@
...
-						{
-							if ($vital_update)
-							{
-								$version_class .= ' app-about__version--update-vital';
+						} elseif ($update_available) {
+							$version_major = explode('.', APP_VER, 2)[0];
+							$update_version_major = explode('.', $new_version['version'], 2)[0];
+
+							if (version_compare($version_major, $update_version_major, '<')) {
+								$update_is_major_version = true;
```
intoeetive, line 19: I think we need to make sure to not apply "vital" class even if update is vital, as that is only relevant per-major-release

obfuscode: I've modified this to be an elseif so it won't show the vital update styles if there's already a major update available.

ops-andy:
@obfuscode link for docs: `/installation/updating.html#upgrading-major-versions`

system/ee/EllisLab/ExpressionEngine/View/_shared/footer.php
intoeetive: I think this is good. Let's package and test it!

intoeetive (approved):

PR #643, opened 2020-11-17
IMPORTANT Update to ft.fluid_field.php

RK311y:
Currently, postSave() is called before the field_data_id is updated. I have moved this method call until after the fluid_field_data table is appropriately updated.

**Background**
I am making a custom field type. Instead of making/maintaining my own database tables, I have chosen to utilize the install() method. This allows me to add new columns to the standard 'channel_data_field_[field_id]' table.

**Problem**
The method post_save($data) in my custom field type class needs to make such changes to the new table columns. In order for this to happen, I need to know the field_data_id from the fluid_field_data table which corresponds to the $this->settings['fluid_field_data_id']. 

Currently, when the post_save($data) method is called, this value is '0'.

**Solution**
I moved the postSave() call to the field on the Facade until after the fluid_field_data table has been updated accordingly. This is how it should be.

### DIFF
--- a/system/ee/EllisLab/Addons/fluid_field/ft.fluid_field.php
+++ b/system/ee/EllisLab/Addons/fluid_field/ft.fluid_field.php
@@ -250,8 +250,6 @@ class Fluid_field_ft extends EE_Fieldtype {
 
 		$values['field_id_' . $field->getId()] = $field->getData();
 
-		$field->postSave();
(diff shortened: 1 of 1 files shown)


RK311y:
I just realized that call $field->postSave() on line 253, which I removed is necessary for existing fields.

PR #756, opened 2021-01-05
add github workflows back EECORE-948

intoeetive:


.github/workflows/tests-cypress-minimal-content.yml
```diff
@@ -0,0 +1,88 @@
+name: Run Cypress Content Tests
+
+on:
+  push:
+    branches:
+      - design/6.0.0/cp
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - release/6.0.0
```
matthewjohns0n: Just wanted to check - are we using these branch names or does this need to change to `6.x`?

.github/workflows/tests-cypress-minimal-installer.yml
```diff
@@ -0,0 +1,88 @@
+name: Run Cypress Installer Tests
+
+on:
+  push:
+    branches:
+      - design/6.0.0/cp
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - release/6.0.0
```
matthewjohns0n: Should be `6.x`

.github/workflows/tests-cypress-minimal-admin.yml
```diff
@@ -0,0 +1,88 @@
+name: Run Cypress Admin Tests
+
+on:
+  push:
+    branches:
+      - design/6.0.0/cp
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - release/6.0.0
```
matthewjohns0n: Should be `6.x`

.github/workflows/release.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - uses: olegtarasov/get-tag@v2
+        id: tagName
+
+      - name: Checkout build tools
+        uses: actions/checkout@v2
+        with:
+          repository: EllisLab/ExpressionEngine-Build
```
matthewjohns0n, line 22: We might need to do something about this, it is a private repo.

matthewjohns0n (changes requested):

.github/workflows/release.yml
intoeetive: Having it listed here just exposes its name, it does not provide any kind of access to external people or machines

matthewjohns0n: Okay - I wasn't sure if this would run into problems accessing that repo since it is a private repo in a different organization. From outside conversation, we determined that this does appear to work, and if any permissions would need to be adjusted, that would be on the github administration side fo things.

PR #761, opened 2021-01-06
Feature/option fieldtype

### LINKED ISSUE #746: Custom Fieldtype - {options:my_field_name} not available in Channel Form
This is an issue dating back to EE3 I believe, and I am curious about why this functionality was disabled. 

In system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php line 2834, there is reference to a $this->custom_option_fields array. This section is commented out in EE3 - EE6. 

We are updating one of our addons for EE6, and my idea was to enable the {option:FIELD_NAME} loop for Channel Form to allow more customization and styling of the form inputs. 

I would prefer to use the core ee functionality similar to 'select', but this appears to be impossible. 

I can get it to work if I add my fieldname to this array: (Line 2771)
`        $this->native_option_fields = array(
            'multi_select',
            'select',
            'radio',
            'checkboxes', 'my_fieldtype_name'
        );
`

Obviously this is not going to work for an addon, so I am looking for another way to do this. 

I would like to be able to register a 'custom_option_field' through my addon code. 

Can someone advise me and let me know if this is possible? If it isn't, and there is a good reason, maybe you can illuminate as to why this chunk of code (line 2834) was abandonned? 

Thanks so much!

Rowan Sentesy
Hop Studios

gilbert-hop:
## Overview

Adding capability for addon developers to be able to declare an option fieldtype so that `{options:field_name}` works. With this feature, you can declare an option fieldtype by setting property `public $is_option_fieldtype = true;` in ft.*.php. It will then pull from the field_list_items like select, checkbox, and radio button when `{options:field_name}` tag is called in the channel form tag. 

Resolves [#746](https://github.com/ExpressionEngine/ExpressionEngine/issues/746).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No

## Documentation


system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -2705,7 +2705,12 @@ public function initialize($reinitialize = false)
         );
 
         $this->custom_fields = array();
-        $this->custom_option_fields = array();
+
+        if ($custom_option_fields = ee('Model')->get('Config')->filter('key', 'custom_option_fields')->filter('site_id', $this->site_id)->first()) {
```
intoeetive: Is there a reason why this is being restricted to current site ID and not just taken from general system configuration?

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -2705,7 +2705,12 @@ public function initialize($reinitialize = false)
...
+            $this->custom_option_fields = json_decode($custom_option_fields->value);
```
intoeetive: Will this work correctly if `$custom_option_fields->value` is empty, or not defined property?

system/ee/legacy/libraries/addons/Addons_installer.php
```diff
@@ -156,6 +156,26 @@ function install_fieldtype($fieldtype)
 
 			$default_settings = $FT->install();
 
+			if (!empty($FT->is_option_fieldtype)) {
```
intoeetive: I think this does not account case where `is_option_fieldtype` is not set at all.
Plus probably should check for boolean and not just if empty

system/ee/legacy/libraries/addons/Addons_installer.php
```diff
@@ -156,6 +156,26 @@ function install_fieldtype($fieldtype)
...
+						]
+					);
+
+					$custom_option_fieldtypes = [$fieldtype];
+				}
+
+				$custom_option_fields->value = json_encode($custom_option_fieldtypes);
+				$custom_option_fields->save();
+			}
+
```
intoeetive: What about already installed fieldtypes, that are being updated?

intoeetive (changes requested):
I have left bunch of comments, please have a look.

But overall, I'm not very confident with this approach - I feel like using config file here leaves space for user error (user here being the site admin with write access to config file). Maybe it is better to check the loaded fieldtype directly?

gilbert-hop:
Hey @intoeetive I used this method because I think it's the most efficient and it allows me to see a list of fieldtypes loaded in the db. Before I move ahead, I'd like to get a sense of which method you want me to proceed:
1. continue with this route
2. ditch my current changes and modify the OptionFieldtype.php

Ideally I'd like to do something like
`ee('Model')->get('Fieldtype')->is('OptionFieldtype')->all();`
Then all the developer needs to do is to extend the OptionFieldtype
Is this possible?

I can also save the option fieldtype declaration inside the settings of the exp_fieldtypes but then I'd have to loop through the fieldtypes to get the option fieldtype since there's no way to query them directly (or is there?).

Any other suggestions?

intoeetive:
I think checking whether fieldtype extends `OptionFieldtype` would be more sane solution.
But this would require looping through fieldtypes. This might be not very resource-effective, but you can cache the results.
We are using similar approach in `getCompatibleFieldtypes()` in `system\ee\ExpressionEngine\Library\CP\EntryManager\ColumnFactory.php`

gilbert-hop:
@intoeetive I didn't want to loop through the fieldtypes and use ReflectionClass before but caching seems to be working just fine. Please have a look at my implementation. I think this is what you were talking about. Please let me know if this code should be moved to a more reusable location.

Also, I've removed the custom_option_fields after confirming that it's only used in this file. The commented out safecracker code can probably go too but I'll leave that to you.

gilbert-hop:
Scratch that.. not actually using the ReflectionClass for this instance.

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -2826,8 +2825,31 @@ public function initialize($reinitialize = false)
...
 
-        $this->option_fields = $this->native_option_fields;
+        if (empty($this->option_fields)) {
+
+            $cache_key = '/Fieldtype/OptionFieldtypes';
+            $this->option_fields = ee()->cache->get($cache_key);
+
+            if (empty($this->option_fields)) {
+                $fieldtypes = ee('Model')->get('Fieldtype')->all()->pluck('name');
+                ee()->legacy_api->instantiate('channel_fields');
```
intoeetive, line 2829: I think you might need to add `ee()->load->library('api');` call as this function seems to be executed before API library is loaded.

intoeetive (commented):
Ok, just added one comment - other than that it is probably good to go (I'll have someone on the team review it as well before releasing though)

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
gilbert-hop: Line added. However, for our instance it was working as is.

intoeetive (dismissed):

gilbert-hop:
Hello @intoeetive,

Just checking in on this. When do you think this will get merged? We are actually kind of waiting on this for a few projects. I also noticed that since we are caching the fieldtypes list, do we need to have a function to clear the cache when user uninstall a fieldtype? Or does the cache name prefix "Fieldtype/" makes it clear itself whenever something happens in fieldtypes?

TomJaeger:
Hi @gilbert-hop 

Currently we're planning on having this one ship in EE 6.1 (due to semantic versioning new features have to fall in line with a minor release.) 

We'll also need you to agree to the contributors agreement (which is forthcoming).

I'll let @intoeetive reply on the cache cleanup end of things.

obfuscode:
@gilbert-hop So it looks like we're finally ready to include this in the next minor version of EE. Would you be able to resolve the conflict on the Channel_form_lib file and then I think we'll be good to go!

gilbert-hop:
@obfuscode done. I asked earlier if we need to worry about clearing the `/Fieldtype/OptionFieldtypes` cache when a fieldtype is installed/removed. What do you think?

obfuscode:
I'll touch base with Yuri to discuss this and get back to you. Thank you for making those changes so quickly!

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -2822,37 +2821,62 @@ public function initialize($reinitialize = false)
...
+                $this->option_fields = array_filter($fieldtypes, function($fieldtype) {
+                    ee()->api_channel_fields->include_handler($fieldtype);
+                    $class = ucfirst($fieldtype) . '_ft';
+
+                    return is_subclass_of($class, 'OptionFieldtype');
+                });
+
+                $this->option_fields = array_values($this->option_fields);
+
+                ee()->cache->save($cache_key, $this->option_fields);
```
intoeetive: @gilbert-hop I have expected EE code for similar cases, and it apears that `ee()->session->cache` is being used instead. I believe session cache does not need invalidation.

If you have strong reasons to use `ee()->cache` this would need to be invalidated when fieldtypes are removed, 'clear all caches' is clicked etc.

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -2822,37 +2821,62 @@ public function initialize($reinitialize = false)
...
+            }
+        }
+
         /*
-                ee()->config->load('config');
+            TODO: I think the following code can be removed
```
intoeetive, line 2850: @gilbert-hop I don't think we can afford to remove this code without prior deprecation, as that might break existing functionality for the sites that are using those config overrides.

intoeetive (changes requested):

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
gilbert-hop: I suggested just because the code block was commented out.

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
gilbert-hop: I can definitely change this to be using the session cache. However, reviewing the entire logic, do we really need to cache the option fieldtypes at all? Usually, you'd only have 1 channel form tag pairs right?

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
intoeetive: Oh, you're right. Let's keep it as is then

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
intoeetive: I'd say to cache is better than not to cache. It is true that usually there is just one channel form, but sometimes there are several, and they can be used multiple times during session

gilbert-hop: @intoeetive sorry I missed the last message. The session cache is implemented now.

intoeetive (approved):

obfuscode (approved):

PR #771, opened 2021-01-11
Fix bug where the color picker fields would not init in the grid field settings page

### LINKED ISSUE #714: Color Picker events are not fired on Grid settings page
When trying to define a Color Picker field as a Grid column, the React component's events are not triggered when you add a new column. The same thing happens in Bloqs (since it's very similar to Grid). This is the JS I added to Bloqs so I can add a Color Picker field:

```
$('input[name^="grid[cols]"]').change(function () {
        if($(this).val() !== 'colorpicker') {
            return;
        }

        // (╯°□°）╯︵ ┻━┻
        setTimeout(function () {
            $('.fields-keyvalue', '[data-fieldtype="colorpicker"]').miniGrid({grid_min_rows:0,grid_max_rows:''});
            ColorPicker.renderFields()
        }, 100);
    })
```

Once the Grid field is saved with a Color Picker column, and the settings page reloads, the Color Picker kicks in.

I had someone else confirm this was an issue as well.

Jordan-Ellis:
This fixes the bug where when you create a new grid field, then add a color picker fieldtype column, the color picker settings fields would not initialize their color picker. 

Fixes #714

### DIFF
--- a/themes/ee/cp/js/build/components/colorpicker.js
+++ b/themes/ee/cp/js/build/components/colorpicker.js
@@ -249,6 +249,9 @@ $(document).ready(function () {
         ColorPicker.renderFields();
     });
 });
+Grid.bind('colorpicker', 'displaySettings', function (el) {
+    ColorPicker.renderFields(el[0]);
+});
 Grid.bind('colorpicker', 'display', function (cell) {
     ColorPicker.renderFields(cell[0]);
 });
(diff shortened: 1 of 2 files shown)


intoeetive (approved):
Works well, thanks!

obfuscode (approved):

PR #773, opened 2021-01-11
Fixes #687 where no valid channels were available in the channel fiel…

### LINKED ISSUE #687: Cannot Create New Entries in 2 Channels (losing the Channel in the Entry Form)

I've been creating entries in the two affected Channels for several months without a problem. Several days ago I began hitting an error on Submit with two Channels. Apparently the "Channel" parameter is lost. On the Options tab there is no Channel below the words "Channel this entry should be submitted to." 

**How To Reproduce**
1. In the CP choose 'Create' and then select one of the two affected Channels
2. Click on the Options tab to see the missing Channel 
3. Additionally, you can fill in the required fields and click Submit, at which point the Options tab turns red and reads "This field is required." below "Channel this entry should be submitted to."

**Error Messages**
<pre><code>This field is required.</code></pre>

**Environment Details:**
 - Version: 5.1.2
 - PHP Version 7.1
 - MySQL Version 5.6
 - OS: Linux
 - Web Server: nginx

**Possible Solution**
n/a

**Additional context**
At the same time these Channels lost their 'Create' ability, two Statuses went missing. I've since manually added-back the Statuses. I have confirmed that no Member/Super Admin deleted the Statuses.

error
missing-channel

robinsowell:
…d on the publish page.

For some reason the order of the status ids returned for $my_statuses = $this->Channel->Statuses->getIds(); was not the same as the order they're returned in $channel_statuses = $channel->Statuses->getIds(); so there was no match of valid channels- not even the actual already existing channel.

They had had some database oddness where the statuses went missing and they added them back, which I'm pretty sure is the reason.  But I tend to think it makes sense to go ahead and sort anyway.  

And I do realize that opens the question of whether we want to sort the custom fields as well.


system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
```diff
@@ -1108,8 +1108,10 @@ public function populateChannels($field)
 
 		foreach ($channels as $channel)
 		{
+			$channel_statuses = $channel->Statuses->getIds();
+
 			if ($my_fields == $channel->getAllCustomFields()->pluck('field_id') &&
```
intoeetive, line 1113: Any idea why we need `$my_fields` here? It does not seem necessary here, but maybe I'm, missing something

intoeetive (commented):
The order of statuses for channel can be custom sorted. @robinsowell can you double check this order is displayed correctly on entry publish page after sorting?

system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
robinsowell: It's what it was named, so I left it.  But basically, it's the current channel's fields.  So you're comparing the current ($my_fields and $my_statuses) fields and statuses to the fields and statuses of every channel.  And only ones that match on both the fields and the statuses go into the channel dropdown.

The goobered up order had it so even the current channel's statuses weren't matching that channel's statuses when retrieved via $channel->Statuses->getIds(); so there were no options in the channel select.  And it's required, so you couldn't even save.  

But the names $my_fields and $my_statuses were already there, so I just left them.  I think in this case 'my' is 'the current entry's channel'.

robinsowell:
@intoeetive I just double checked, and the order is properly reflected on the entry publish page with the patch applied, so that bit looks good.

system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
intoeetive: Ok, I thought you might know - we're probably fine leaving it there now, though this seems like some place for optimization

intoeetive (approved):

obfuscode (approved):

PR #821, opened 2021-01-26
fix #817 where it was not possible to add new items in Simple Commerce

### LINKED ISSUE #192: On MacOS Safari, Entry Manager (5.2.2) won't accept keyword filter after any other filters have been changed
This is a minor but annoying UI glitch, which seems to be restricted to MacOS Safari. It doesn't occur in Firefox or Chrome.

After having changed any of the other filters, the Entry Manager won't accept any new (or different) keyword filters. No JS issues appear in the console. Clearing the other filters doesn't help; the page must be reloaded, and the keyword filter must be added as the first filter if it is to be accepted.

The effect is minor, but it is an annoyance when performing multiple or more complicated searches.

**How To Reproduce**
In MacOS Safari (tested with 12.1, running on MacOS 10.14.4), go to the Entry Manager. Add a keyword filter without first having changed any of the other filters. This will work as expected.

Reload the page. Change another filter — for example change "search in (Titles)" to "search in (Titles & Content)". It now won't be possible to add a keyword filter.

**Environment Details:**
 - Version: 5.2.2
 - MacOS 10.14.4
 - Safari 12.1

### LINKED ISSUE #817: Simpler commerce- you can't add new items.
Install the simple commerce add-on and hit 'Add Items'.  It pulls up all of the entries, you can select entries to be items.

Checkbox a few, scroll to the bottom and select 'Add Item' and submit- it just reloads the list of entries to select from.

intoeetive:
also closes https://github.com/ExpressionEngine/ExpressionEngine/issues/192

### DIFF
--- a/themes/ee/asset/javascript/src/cp/publish/entry-list.js
+++ b/themes/ee/asset/javascript/src/cp/publish/entry-list.js
@@ -62,6 +62,14 @@ $(document).ready(function () {
 	}
 
 	// Submitting the search form
+	$('body').on('click', 'button[name="bulk_action_submit"]', function(event) {
+
+		event.preventDefault();
+		$('body').off('submit', form_selector);
+		$(form_selector).submit();
+	});
+
+	// Submitting the search form
 	$('body').on('submit', form_selector, function(event) {
 
 		event.preventDefault();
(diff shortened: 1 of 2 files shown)


robinsowell (dismissed):
Tested both in Safari and all works well for me.

obfuscode (approved):

PR #938, opened 2021-03-08
change login extension hook placement fot CP logins

stephengalbraith:
## Overview

While working on a small extension to allow users to login from the front-end via their email instead of username, I noticed that unlike in the front-end login form the hook for the CP gets called after the username variable is set, effectively not allowing the hook to be used to its full potential.

For example in system/ee/ExpressionEngine/Addons/member/mod.member_auth.php where the 'member_member_login_start' is called before setting the variables that get checked.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Auth.php
+++ b/system/ee/legacy/libraries/Auth.php
@@ -176,6 +176,21 @@ class Auth
      */
     public function verify()
     {
+        // If this is being called from the CP, use the hook
+        if (REQ == 'CP') {
+            /* -------------------------------------------
+            /* 'login_authenticate_start' hook.
+            /*  - Take control of CP authentication routine
+            /*  - Added EE 1.4.2
+            */
+            ee()->extensions->call('login_authenticate_start');
+            if (ee()->extensions->end_script === true) {
+                return;
+            }
+            /*
+            /* -------------------------------------------*/
+        }
+        
         $username = (string) ee()->input->post('username');
 
         // No username/password?  Bounce them...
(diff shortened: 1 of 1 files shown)


intoeetive (commented):
I wonder if the existing code order was intentional, not allowing to change username on CP logins as a security measure.

@matthewjohns0n I'd like your input here as well

matthewjohns0n:
I think this change makes sense. @intoeetive can we run the tests again before we merge, since they failed?

intoeetive:
@matthewjohns0n we probably need to merge in latest 6.dev to run the tests, as the testing workflow has been updated

intoeetive:
@stephengalbraith can you pull in latest `ExpressionEngine:6.dev` and merged it into your branch? I tried to do it, but was not successful

stephengalbraith:
@intoeetive it looks like my branch is equal to 6.dev now. If necessary, I can submit a new PR just in case there's something else going wrong since this was done in March

intoeetive:
@stephengalbraith you need to pull 6.dev from EE repo origin (not your fork's origin) and then merged (so you'll need to add ExpressionEngine/ExpressionEngine as an extra remote)

stephengalbraith:
Closed - see #1515

PR #948, opened 2021-03-15
Re-adding missing .php_cs.dist file

obfuscode:
Looks like this file was accidentally deleted during the 6.0.3 update and needs to be re-added to the repo.


intoeetive (approved):
Since this file previosly existed in 6.0 and was mistakenly removed, it is ok to add it back in patch release

PR #980, opened 2021-03-31
fixes #233 where Template Route could be accidently erased

### LINKED ISSUE #233: Saving a template already attached to a Template Route with segment will cause the matching URL to fail
If you create a Template Route with a segment, eg. route = template_name/{type:alpha_dash}, everytime you save that particular template will cause the matching URL to fail.

Error: The page you requested was not found.

To solve the above problem, you have to click Update in the Template Routes setting page.

**How To Reproduce**
I have created a template as below:
Books (template group) / listing (template)

Inside the Template Routes settings:
Route = /listing/{type:alpha_dash}

After creating the route, browse to site.com/listing/my-segment

Page is shown as intended.

Now edit your listing template. Browse to site.com/listing/my-segment. Error is shown.

Update Template Routes settings to solve error.

**Error Messages**
Error: The page you requested was not found.

**Environment Details:**
 - Version: 5.2.3
 - PHP Version 7.1
 - MySQL Version 5.6
 - OS: Linux
 - Web Server: Apache

**Additional context**
Only template routes with segment will cause this error.

intoeetive:
fixes #233 where Template Route could be accidently erased when editing template

EECORE-1030


TomJaeger (approved):
Looks good.

PR #985, opened 2021-03-31
RTE field enhancements

### LINKED ISSUE #669: EE 6 RTE field enhancements
Two feature requests for the new RTE field in EE 6:

1. Add a "source" button to reveal the HTML source of the field contents to allow for custom HTML editing.

2. Add the ability to restrict the height of the field.  Otherwise, long text requires constant scrolling from mid/bottom back to top to be able to click the formatting buttons to apply formats.  It's virtually unusable like this.  Even with the height of the field set to 200 or 400, it does not limit to that height if the contents of the field are longer.

### LINKED ISSUE #957: RTE icon positions reverting when editing configurations
I have created a new RTE Config called "Standard" (which has some of the icons in) and mapped this to a content Field.
Part of this config was to edit the sequence of the icons, in the toolbar.

These seem to display fine when editing / creating an Entry.
But when editing the RTE Config again, the sequence has reverted back to the original layout.

EE 6.0.3
PHP 5.6

intoeetive:
fixes #669
fixes #957


obfuscode:
@intoeetive Can you resolve the conflicts with this PR?

intoeetive:
@obfuscode to my belief, I have resolved this correctly, but need to test the new and pre-existing features and fixes

system/ee/installer/updates/ud_6_01_00.php
```diff
@@ -39,6 +38,15 @@ public function do_update()
         return true;
     }
 
+    private function removeRteExtension()
+    {
+        ee()->db->where('name', 'Rte')->update('fieldtypes', ['version' => '2.0.1']);
+
+        ee()->db->where('module_name', 'Rte')->update('modules', ['module_version' => '2.0.1']);
```
obfuscode, line 46: If this is the remove function, shouldn't the version be updated to 2.0.0 as part of the rollback?

intoeetive: I don't think this qualifies as rollback. Removing extension is part of update, which removes the code that is not needed. Since there is change in DB, the version is being updated

obfuscode (dismissed):

obfuscode (approved):

PR #993, opened 2021-04-01
fix #936 where some data were erased when saving member role 

### LINKED ISSUE #936: Disabling "Access file manager" permissions removes the assigned custom menu set
v6.0.3

When editing a role, and I disable the "Access file manager" permission, which was previously enabled, it also sets the `menu_set_id` column in the `role_settings` table to 0. I have to run the following query to restore the assigned menu items to the role.

```
UPDATE `exp_role_settings` SET `menu_set_id`='2' WHERE `role_id` = 6
```

intoeetive:
fix #936 where some data were erased when saving member role EECORE-1082
fixed issue when roles list did not properly save for some menu sets


TomJaeger (approved):
Looks good

PR #994, opened 2021-04-01
display proper alerts if entry could not be deleted

### LINKED ISSUE #941: UI allows delete entries from other users
v6.0.3

This is a Slack convo with a co-worker. The were presented with the UI options to delete an entry in a channel that they did not have permissions to delete entries from other members. EE took them through the process of deleting the entry, but it does not actually delete the entry, which is good, but, this appears to be more of a presentational issue than backend issue. I investigated this a bit and I'm not exactly sure how to fix it, mostly b/c there could be multiple paths to fixing it. 

Since the entry manager lists so many entries from different channels and users, and if multiple entries are selected, then it's difficult to impossible to determine which of those should have the checkbox hidden, or remove the action from the dropdown list, b/c there are so many permutations of what actions could be taken.

My first thought is the dialog should inform the user that 1 or more of the entries selected do not belong to them, thus they can't be deleted. If 1 entry is selected, and they can't delete it, then it should say you can't delete this entry and not display the submit button. If 2-5 entries are selected, then it should list the entry titles (b/c I think the confirm dialogs list up to 5 entry titles, then after that it just says "Entry: 25 entries") and next to the entry that the user does not have permissions to delete it should say something about not able to delete that entry. If more than 5 entries are selected, then it should probably say "you don't have permissions to delete 1 or more of the entries you've selected".

For this reason, I'm not sure I can provide a fix for this b/c it appears to require some UI/UX decisions.

ee-bug-deleting-other-entries

The issue seems to be stemming from this check. I think `hasAny` will return true if any of the conditions are valid. Unless I'm misunderstanding this I don't think `hasAny` checks should used here because if delete self entries is valid, but delete all is invalid, then it's going to return true, when attempting to delete entries from other users.

```
$vars['can_delete'] = ee('Permission')->hasAny(
                'can_delete_all_entries_channel_id_' . $channel_id,
                'can_delete_self_entries_channel_id_' . $channel_id
            );
```

intoeetive:
fixes #941; EECORE-1079

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/Edit.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/Edit.php
@@ -549,12 +549,24 @@ class Edit extends AbstractPublishController
 
         $entry_names = array_merge($this->removeAllEntries($entry_ids), $this->removeSelfEntries($entry_ids));
 
-        ee('CP/Alert')->makeInline('entries-form')
-            ->asSuccess()
-            ->withTitle(lang('success'))
-            ->addToBody(lang('entries_deleted_desc'))
-            ->addToBody($entry_names)
-            ->defer();
+        if (!empty($entry_names)) {
+            ee('CP/Alert')->makeInline('entries-form')
+                ->asSuccess()
+                ->withTitle(lang('success'))
+                ->addToBody(lang('entries_deleted_desc'))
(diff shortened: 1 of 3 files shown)


TomJaeger (approved):
Pulled tested, and reviewed code.

PR #999, opened 2021-04-05
fix #998 where exp:member:edit_profile was throwing an error

### LINKED ISSUE #998: exp:member:edit_profile form error when user not logged in
In a front end members area, a page with `{exp:member:edit_profile}` throws an error when the user accessing this page is not logged in.
```
Error Caught
Call to a member function getValues() on null
mod.member_settings.php:764
```
This happens even if the page has a redirect to the login form if the user is not logged in.

The error can be reproduced by copying the example form from https://docs.expressionengine.com/latest/member/edit-profile.html#example and putting it in a template.

If you add a redirect, `{if logged_out}{redirect="/sign-in"}{/if}` the error will still happen.

Note that the error can be fixed by commenting out the line in `mod.member_settings.php:764`.  
`Line 764: $result_row = $this->member->getValues();`
It looks like the variable `$result_row` is never used and that line can be removed.

This is in a docker container that's run by [Lando](https://lando.dev/)
Its a lamp server with EE 6.0.3

intoeetive:
fix #998 where exp:member:edit_profile was throwing an error when logged out;

EECORE-1093

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member_settings.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member_settings.php
@@ -761,6 +761,14 @@ class Member_settings extends Member
 
         $this->member = ee('Model')->get('Member', ee()->session->userdata('member_id'))->first();
 
+        if (empty($this->member)) {
+            if (! empty($tagdata)) {
+                return ee()->TMPL->no_results();
+            } else {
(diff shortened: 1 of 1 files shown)


TomJaeger (approved):
reviewed, tested and it looks good

PR #1000, opened 2021-04-05
filepicker in Fluid textareas

### LINKED ISSUE #207: File/Image Upload Button fails after new textarea field is created w/in a fluid field
When a new Textarea field is added within a Fluid Field and someone tries to immediately use the HTML/Formatting Image Upload button, the button does not work. I don't see any errors in the console either. 

It doesn't seem to matter if the entry is brand new or a preexisting entry that's been saved and reopened. 

In either case, the entry must first be saved with its new Fluid > Textarea field in place. Once the entry is saved, the Image Upload button works as expected.

**How to Reproduce:**
1. Create a Textarea field and enable Formatting Buttons
2. Create a Fluid Field and select the new, aforementioned Textarea field from the Field Options.
3. Add the new Fluid Field to some channel's field settings.
4. Create a new entry in that channel, or open an existing one. 
5. Add a Textarea field within the Fluid Field and immediately try to upload an image or file into it using the Image Upload button (cloud w/ arrow icon). 
6. You shouldn't be able to. 
7. Try saving the entry with your new Fluid<Textarea field still in place.
8. The Image Upload button should now work for that field.

**Environment Details:**
 - Version: 5.2.2
 - PHP Version: 7.1
 - MySQL Version: 5.6.43
 - OS: macOS
 - Web Server: Apache

intoeetive:
fix #207 where file picker in textarea toolbar was not initialized for new fields in Fluid;

EECORE-1076


ops-andy (approved):
confirmed this fixes both #207 and #1020

PR #1002, opened 2021-04-06
fix issue when new post could not be published using Metaweblog API

intoeetive:
EECORE-878


robinsowell (approved):

robinsowell:
Worked for me in a test with MarsEdit.

PR #1009, opened 2021-04-09
cookies updates

### LINKED ISSUE #271: exp:consent:form does not log data from non-member users
{exp:consent:form consent='ee:cookies_functionality' form_id="cForm"}
          {consents}
              <p>{consent_request}</p>
              <input type="hidden" name="{consent_short_name}" value="y" {if consent_granted}checked{/if}>
          {/consents}
          <input style="margin-bottom: 25px;" class="button" type="submit" name="submit" value="Accept">
      {/exp:consent:form}

This seems to work fine with logged in users, creating a log verifying their consent. This isn't the case for someone who isn't logged in. The cookie is saved to their browser but fails to produce a log.
Recently upgraded to v5.2.2

### LINKED ISSUE #955: Update EE6 to remove all persistent cookies in default state
EE mostly uses session based cookies, but sets a number of "Strictly Necessary" cookies that are "persistent" beyond the session.  These are flash (1 day), last_activity (1 year), last_visit (1 year), remember (2 weeks),  visitor_consents (1 year).

The presence of these persistent cookies trips a wire in GDPR - for each of these cookies it is necessary for the developer to provide justification for the need for the persistent cookie and report that the cookie cannot be linked to user activity.  Sometimes these justifications are not heard sympathetically - if the 'strictly necessary' determination is not accepted then these cookies must only be set if a visitor to the site from a GDPR country actively consents.

There is no mechanism in EE either to adjust the duration of these cookies, or to link their creation to active consent.

An option in EE CP to set the duration of all cookies to be session based by default.

If there are functional benefits to be gained by making any of the cookies have a longer life, this should be a choice that can be made by users and set using the consent features of EE.

There are no alternative solutions right now.

How this gets documented depends on the solution that is adopted.
Provided switching to session duration cookies is not something that breaks EE, there should be no migration issues.

intoeetive:
add logging of all consent requests including guest ones; 
remove unused cookies;
change cookies lifetime;
reorganize cookies

closes #271; closes #955; 
EECORE-1096; EECORE-1095

https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/153

Improve EE's Ability to be compliant with GDPR EECORE-1153


robinsowell:
So here's what I have so far that needs checking:

** Can't change between MSM sites in cp. 
** I don't think Remember me worked this morning- but needs another test 
- I seem to get the modal to re-login a lot after logging in and just in general more than I did before.  Be curious if it's just me thinking that and if I say only have 1 browser open or some such, whether it will matter.  But still- I usually have a billion browsers open.
- Should we kill existing cookies if they switch their settings not to allow that type of cookie?

I still need to:
- check logging of all consents - including anon
- make sure anon can still edit their consents
- check all of the consent module tags and make sure it's all good.
- Play with forum and comments more, play logged out more.

Also- totally unrelated, I don't have a save button on the bottom of my entry page.  I'd swear I used to.  Safari? Cookie?  Hallucination?

Loose rundown of how I tested

Logged in as superadmin: In CP it was already set to functionality cookies only at the start, consent is required in settings.- no performance or targetting cookies alowed.  I opt out of Functionality too in cp.  So- no cookies allowed now?

Go to frontend I have following cookies still set:
csrf_token
flash
last_activity
last_visit
sessionid

Logging out killed sessionid and that's it as far as cookies go.  I'm going to manually kill all cookies and see what happens.

OK, refresh and still logged out, now I only have csrf_token cookie.  Login frontend- now I have csrf_token and sessionid. Random order channels still work across pages

Go to cp, I'm sill logged in, csrf_token and sessionid exist.  I save some stuff- flash cookie now exists.  It is showing alerts properly.  Bork around, play with MSM.  Play with Forum- it seems to work fine.

intoeetive:
@robinsowell 

Answering the questions/notes:
- Switching the sites in CP does not work if you have consents required, but you did not opt in for Functionality cookies. This is expected, however I would expect a banner in CP with the consent form. I'll need to see how difficult it is to implement one.
- 'Remember' cookie is being set. I also did not spot any differences in how the login modal shows up.
- *Should we kill existing cookies if they switch their settings not to allow that type of cookie?* - I'd wait to see if someone requests this. I don't think this should be done, as the cookie was set when it was allowed. If the cookie needs to be updated, and the consent has been revoked, it will not be updated. This would fall in line with my expectations
- *save button* - since we have sticky header with tabs and save button on edit page, the button has been removed from footer in 6.0

I will likely need to write some automated tests to cover Consent module, making sure cookies are set/not set depending on user's consent

obfuscode (approved):

pixiwooh:
@intoeetive 

_Should we kill existing cookies if they switch their settings not to allow that type of cookie?_ 
Hi Yuri, Cookiebot (https://www.cookiebot.com/fr/), the service I am currently using kills them on consent's revoke. As an european drilled to privacy, to me it make sense. When I revoke a consent I "say" to the website from now don't track me anymore.
If killing cookies is not added to EE I will (others europeans than me too I guess) still have to rely on third party services like cookiebot despite the fact that you added the 'add logging of all consent requests including guest ones;' functionality :'-/

robsonsobral:
@intoeetive , is this important to you?

https://expressionengine.com/forums/topic/253095/secure-cookies-form-failures

intoeetive:
@robsonsobral I believe this has been reported in https://github.com/ExpressionEngine/ExpressionEngine/issues/1135

iammattmartin:
Sorry for the duplication in the forum and on here but will give this a test.

iammattmartin:
@intoeetive @robsonsobral I've tested this but still setting secure in either the GUI or config file for the site does not generate any cookies when enabled.

Setting either to no or "n" generates them normally.

PR #1030, opened 2021-04-17
support image modifiers for category images

### LINKED ISSUE #942: Apply image manipulations to category images
My site has a relatively extensive but important use of categories, wherein each category appears in the navigation and has its own dedicated page wherein its entries are listed. On these pages and in the navigation, the category image is shown in various sizes. Currently when I upload an image to a category's category image field, it requests to know which upload directory to use, which implies that it would apply the standard image manipulations attached to that file upload dir and, correspondingly provides the ability to output those image manipulations in the template code using the standard syntax like {category_image:webp} or {category_image:thumbnail}. It does not. As such my site is causing mammoth 2MB to 5MB category image downloads for new visitors, causing a total page size of 22MBs on the home page www.spainexpat.com.

Can you please improve the category images so that the same image manipulations apply to the category images as are applied to any other file upload field on a per file upload directory basis?

intoeetive:
also covers the categories functionality with Cypress tests

Fixes #942 
EECORE-1072

### DIFF
--- a/tests/cypress/cypress/elements/pages/channel/Category.js
+++ b/tests/cypress/cypress/elements/pages/channel/Category.js
@@ -1,0 +1,15 @@
+import ControlPanel from '../ControlPanel'
+
+class Category extends ControlPanel {
+    constructor() {
+        super()
+        this.url = 'admin.php?/cp/categories';
+
+        this.selectors = Object.assign(this.selectors, {
+            
+        })
+    }
+
+
+}
+export default Category;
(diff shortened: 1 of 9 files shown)


tests/cypress/cypress/elements/pages/channel/Category.js
```diff
@@ -0,0 +1,15 @@
+import ControlPanel from '../ControlPanel'
+
+class Category extends ControlPanel {
+    constructor() {
+        super()
+        this.url = 'admin.php?/cp/categories';
```
obfuscode, line 6: Can we assume this is `admin.php` if the file has been renamed?

obfuscode (approved):

PR #1040, opened 2021-04-20
Remove CSS styling which sets status to uppercase

### LINKED ISSUE #246: Status is Case-Sensitive, UI is Uppercase
When the status is MyStatus:
<input name="status" value="MyStatus"> Works.
<input name="status" value="MYSTATUS">  Fails.

However in the CP all the statuses are uppercased "MYSTATUS".  Only way to know the status is to get source.

Solution is to remove uppercase styling in CP.

johnfuller:
## Overview

The status column in the list of entries shows the status items in
uppercase. This is caused by a text-transform property in the CSS for
the "status-tag" class.

The status column is useful as a reference to be used in channel entry
tags, but the CSS obscures the actual casing being used.

Change the text-transform property from uppercase to "none." This change
affects status items being used elsewhere, creating the same issue in
each of these instances.

Resolves [#246 ](https://github.com/ExpressionEngine/ExpressionEngine/issues/246).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

No documentation needed.

### DIFF
--- a/changelogs/major.md
+++ b/changelogs/major.md
@@ -56,6 +56,7 @@ Developers
 
 - Forgot Password emails will now respect your "Mail Format" preference (essentially enabling the ability to use HTML in Forgot Password emails).
 - Fixed a bug where table bulk selections can be saved by the browser on page reload, but don't show in the UI.
+- Control panel status column in entry list displays status text casing which matches the actual casing as entered into the system on creation. 
 
 EOF MARKER: This line helps prevent merge conflicts when things are
 added on the bottoms of lists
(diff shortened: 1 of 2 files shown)


obfuscode (approved):
Changelog entry is technically a patch fix but it doesn't matter for our current changelog system.

PR #1054, opened 2021-04-30
apply search and sort parameters for Grid fields in LivePreview mode

### LINKED ISSUE #1044: Search parameter applied to Grid loop doesn't get applied in Live Preview
I’m editing an entry that has a Grid field and in the template that grid loop has a `search=` parameter applied to it to limit its output. That same template is used as the template for Live Preview. In normal front-end conditions the loop displays exactly as intended, but in Live Preview the search parameter is ignored and all rows of the grid are output. This is on an entry editing page, of course, so URL is cp/publish/edit/entry/xxxxxxxxxxxxx.

Screenshot 2021-04-23 at 15 58 31

The screenshot attached shows the editing pane with the preview pane beside it, the template code generating the output, and lastly the output as it occurs (correctly) on the final front end.

This is with EE 6.0.3 (but I noticed the issue with EE5 as well), PHP 7.2.22 and MySQL 5.7.26 running on MAMP Pro, but it also occurs on the live server.

intoeetive:
fixes #1044
EECORE-1152

### DIFF
--- a/system/ee/legacy/models/grid_model.php
+++ b/system/ee/legacy/models/grid_model.php
@@ -427,11 +427,7 @@ class Grid_model extends CI_Model
 
         $entry_data = isset($this->_grid_data[$content_type][$field_id][$marker]) ? $this->_grid_data[$content_type][$field_id][$marker] : false;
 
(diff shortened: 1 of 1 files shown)


TomJaeger (approved):
Tested and confirmed it's working. also reviewed code

PR #1103, opened 2021-05-24
Display the entry_id along with the channel name in Relationship fields

litzinger:
Many large sites might have entries with similar or identical titles. Other sites use multilingual add-ons where an entry title might not be readable by a content editor. Adding the entry_id can help make identifying entries easier.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

Screen Shot 2021-05-24 at 7 48 17 AM


litzinger:
The only downside to this is that adding the entry_id in the description is that it isn't searchable. If the ID were prefixed or suffixed to the title, then a user could find an entry by ID instead of title in the search bar. Displaying the ID could also be hidden behind a hidden config variable too if you don't want to add additional info to the UI, but I think power users (as well as sites using Transcribe or Publisher) could benefit from such a feature.

litzinger:
Here is what it might look like if the ID was part of the title, which makes it more functional.

relationship-ids

ajohnson:
@litzinger @intoeetive I agree, this is a good use case for making this update.

I'm having a hard time pulling the branch, but based on your attached screenshots I'd recommend a couple styling tweaks:

1. I agree adding the entry ID to the end of the title to make the ID searchable is beneficial. But I'd recommend tweaking the style of the ID so it doesn't look like part of the actual title content. It could look like this (ID CSS also attached below):
Screen Shot 2021-09-14 at 12 52 44 PM
Screen Shot 2021-09-14 at 12 52 57 PM

2. When displaying the entry IDs on the selected entries I'd recommend putting the ID next to the channel name under the entry name where the channel name currently displays. This is a pattern we're already using when displaying IDs on the Channels and Fields listing screens (and if we break that pattern here we should consider breaking it elsewhere). It would look like this:
Screen Shot 2021-09-14 at 12 58 12 PM

3. Turning the display of these entry IDs on/off via a config variable sounds like a good idea. Personally I'd only want to turn these on for certain sites and certain clients, with the IDs being turned off by default.

These are just my gut reactions after looking at your screenshots, but feel free to push back with other thoughts! Thanks!

litzinger:
It'll take me some time to refactor this PR to include those design changes, and to make this work based on a config value.

shbchk:
Hello! Maybe, in addition to displaying the entry_id, we can also give it a link to the entry page? I mean, if we already have it in channel settings (Search results URL or Preview URL), why not use it in this case?

Like this: `<a href="{search_result_url}">{entry_id}</a>`

system/ee/ExpressionEngine/Addons/relationship/libraries/EntryList.php
```diff
@@ -204,6 +204,7 @@ public function ajaxFilter()
         $settings['channel_id'] = ee('Request')->get('channel_id');
         $settings['related'] = ee('Request')->get('related');
         $settings['selected'] = ee('Request')->get('selected');
+        $settings['selected'] = ee('Request')->get('selected');
```
obfuscode: Looks like these lines were duplicated.

system/ee/ExpressionEngine/Addons/relationship/libraries/EntryList.php
```diff
@@ -214,7 +215,7 @@ public function ajaxFilter()
             $response[] = [
                 'value' => $entry->getId(),
                 'label' => $entry->title,
-                'instructions' => $entry->Channel->channel_title
+                'instructions' => $entry->Channel->channel_title,
```
obfuscode, line 217: This comma was added. Was there something that needed to go after it here?

ajohnson:
@Yulyaswan Could you copy/paste here the correct `$config` setting I should add to my `config.php` file in order to display the entry IDs? I'm messing up the syntax somewhere. Thanks!

/cc @intoeetive

intoeetive:
@ajohnson it is not on config file; it's a toggle that should be turned on in field settings

ajohnson:
@intoeetive Thanks!

Looks good. This will be a nice option to have.

ajohnson (approved):

system/ee/ExpressionEngine/Addons/relationship/libraries/EntryList.php
litzinger: Nope, thats just good practice IMO ;) 

PHP isn't silly like JS and stumbles on trailing commas. It's a good idea to leave a trailing comma so if something is added to the next line, you don't have to worry about a missing comma. Also allows easy re-ordering or commenting out of array keys without having to worry about commas. Also helps with diffs.

Always use trailing commas :)

obfuscode:
@litzinger If you can resolve the conflicts and fix that one duplicate issue, this looks good to get merged into 6.2.

litzinger:
What conflict? This PR barely resembles the one I originally created, and the duplicate line isn't from my commit.

intoeetive:
I've resolved the conflicts, but there have been quite few of those, so might be worth to review once again.

TomJaeger:
Looks like tests are failing on this one.

intoeetive (approved):
I think we're good to merge it at this point. The automated tests have been added and they are passing

PR #1116, opened 2021-06-03
Bug/6.x/files perpage resets filters

xhezairbey:
1. Removed $reset_url variable, it breaks sorting filters when perPage is first second
2. Reordered `AbstractFiles::listingsPage()` method `$vars` initialization, now checks against the non-default `$view_type` first
3. Fixed ViewType filter class name, removed unused properties & refactored methods
4. Removed duplicate viewtype attribute where ‘list’ is the default

## Overview

If the Files table is sorted first by Title, Type or Date, and then perPage filter is set to a different value, the sorting is reset back to default, and `sort_col` & `sort_dir` URL params are wiped.

This worked fine on EE5 for Files, but is broken on EE6 due to the introduction of the `viewtype` filter. There's a `$reset_url` variable instead of `$base_url` used which does reset the URL query params but then breaks the sorting.

## Nature of This Change

- [x] 🐛  Fixes a bug
- [ ] 🚀  Implements a new feature
- [x] 🛁  Refactors existing code
- [ ] 💅  Fixes coding style
- [ ] ✅  Adds tests
- [ ] 👽  Adds new dependency
- [ ] 🔥  Removes unused files / code
- [ ] 🔒  Improves security 

- [ ] Yes
- [x] No with EE5


intoeetive:
@obfuscode you should know how the filters work the best - would you mind reviewing this?

obfuscode (approved):

PR #1132, opened 2021-06-10
Fix comment issues EECORE-1041 EECORE-1040

bakin1999:
## Overview

Fixing the bugs from EECORE-1040 and  EECORE-1041.

## Nature of This Change

- [Check ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Addons/comment/mod.comment.php
+++ b/system/ee/ExpressionEngine/Addons/comment/mod.comment.php
@@ -1801,8 +1801,8 @@ class Comment
 
         // Bleh- really need a conditional for if they are subscribed
 
-        $sub_link = ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . $action_id . '&entry_id=' . $entry_id . '&ret=' . ee()->uri->uri_string();
-        $unsub_link = ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . $action_id . '&entry_id=' . $entry_id . '&type=unsubscribe' . '&ret=' . ee()->uri->uri_string();
+        $sub_link = ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . $action_id . '&entry_id=' . $entry_id . '&ret=' . ee()->uri->uri_string() .'&csrf_token=' . CSRF_TOKEN;
+        $unsub_link = ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . $action_id . '&entry_id=' . $entry_id . '&type=unsubscribe' . '&ret=' . ee()->uri->uri_string() . '&csrf_token=' . CSRF_TOKEN;
 
         $data[] = array('subscribe_link' => $sub_link, 'unsubscribe_link' => $unsub_link, 'subscribed' => $subscribed);
 
--- a/system/ee/ExpressionEngine/Addons/comment/mod.comment.php
+++ b/system/ee/ExpressionEngine/Addons/comment/mod.comment.php
@@ -1886,7 +1886,7 @@ class Comment
      */
     public function comment_subscribe()
     {
-        if (ee()->input->get('csrf_token') != CSRF_TOKEN) {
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/comment/mod.comment.php
```diff
@@ -31,16 +31,18 @@ class Comment
...
-                if ($val == 'comment') {
-                    $_POST[$val] = ee('Security/XSS')->clean($_POST[$val]);
+                    if ($val == 'comment') {
+                    
+                        $_POST[$val] = ee('Security/XSS')->clean($_POST[$val]);
+                    }
                 }
             }
-        }
+
```
obfuscode: It doesn't look like anything was changed here except the indentation was increased and an additional line was added. Can this be reverted as the changes here are incorrect.

system/ee/ExpressionEngine/Addons/comment/mod.comment.php
```diff
@@ -1267,6 +1269,8 @@ public function preview_handler()
             exit('Preview template not specified in your comment form tag');
         }
 
+        
+
```
obfuscode: Please remove these additional lines that were added.

obfuscode (changes requested):

obfuscode (dismissed):

system/ee/ExpressionEngine/Addons/comment/mod.comment.php
```diff
@@ -1887,6 +1886,9 @@ public function subscriber_list()
      */
     public function comment_subscribe()
     {
+        if (ee()->input->get('token') != CSRF_TOKEN) {
```
intoeetive: This is checking for `'token'` GET variable, however I don't see where it's set - can you point me ? (again, it would be preferred to not pass that openly, but to pass that as header or POST variable)

intoeetive (changes requested):

system/ee/ExpressionEngine/Addons/comment/mod.comment.php
obfuscode: After taking another look at this, I think I'm fine with the CSRF being in the query string. Previously, these links were completely unsecured and changing them to a header or POST would require refactoring how the Subscribe and Unsubscribe mechanism works. I think the CSRF is a step in the right direction but we would need to decide to refactor this code entirely if we don't want to go with this modified approach. @matthewjohns0n Do you have concerns about this from a security stand-point?

system/ee/ExpressionEngine/Addons/comment/mod.comment.php
```diff
@@ -1887,6 +1886,9 @@ public function subscriber_list()
      */
     public function comment_subscribe()
     {
+        if (!bool_config_item('disable_csrf_protection') && ee()->input->get('csrf_token')!=CSRF_TOKEN) {
```
matthewjohns0n: I would change this comparison operator to !== instead of !=.

system/ee/ExpressionEngine/Addons/comment/mod.comment.php
```diff
@@ -1978,8 +1980,14 @@ public function comment_subscribe()
      */
     public function edit_comment($ajax_request = true)
     {
-        @header("Content-type: text/html; charset=UTF-8");
+        /*
+        This check is needed because otherwise, links could be created to CSRF and edit comments.
+        */
+        if (!bool_config_item('disable_csrf_protection') && ee()->input->get('csrf_token') != CSRF_TOKEN) {
```
matthewjohns0n: I would change this comparison operator to !== instead of !=.

matthewjohns0n (changes requested):

obfuscode (approved):

obfuscode:
Matt and Yuri verbally approved this so I'm admin merging it despite "Requested Changes" still being active.

PR #1142, opened 2021-06-17
EECORE-1172 Forum Closing Bug

bakin1999:
## Overview
Fixing 1172, now only certified users can change status of the forum.

## Nature of This Change

- [  ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
+++ b/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
@@ -5933,7 +5933,7 @@ class Forum_Core extends Forum
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
         $token = ee('Request')->post('csrf_token');
-        if (! $this->_mod_permission('can_change_status', $query->row('forum_id')) || (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
+        if (! $this->_mod_permission('can_change_status', $query->row('forum_id')) || (! bool_config_item('disable_csrf_protection') && $token !== CSRF_TOKEN)) {
             return ee()->output->show_user_error('general', array(lang('not_authorized')));
         }
 
--- a/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
+++ b/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
@@ -5931,7 +5931,8 @@ class Forum_Core extends Forum
         $viewpath = ($query->row('announcement') == 'n') ? 'viewthread/' : 'viewannounce/';
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
-        if (!array_key_exists('csrf_token',$_GET) ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) || $_GET['csrf_token'] != CSRF_TOKEN) {
+        $token = ee('Request')->get('csrf_token');
+        if ( (! bool_config_item('disable_csrf_protection') &&  !array_key_exists('csrf_token',$_GET)  )  ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) ||  (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
             return ee()->output->show_user_error('general', array(lang('not_authorized')));
         }
 
--- a/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
+++ b/system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
@@ -5932,7 +5932,7 @@ class Forum_Core extends Forum
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
         $token = ee('Request')->get('csrf_token');
-        if ( (! bool_config_item('disable_csrf_protection') &&  !array_key_exists('csrf_token',$_GET)  )  ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) ||  (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
+        if (! $this->_mod_permission('can_change_status', $query->row('forum_id')) || (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
             return ee()->output->show_user_error('general', array(lang('not_authorized')));
         }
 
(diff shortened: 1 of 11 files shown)


system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
```diff
@@ -5930,10 +5931,11 @@ public function change_status()
         $viewpath = ($query->row('announcement') == 'n') ? 'viewthread/' : 'viewannounce/';
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
-        if (! $this->_mod_permission('can_change_status', $query->row('forum_id'))) {
+        if (!array_key_exists('csrf_token',$_GET) ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) || $_GET['csrf_token'] != CSRF_TOKEN) {
```
intoeetive: Please use EE functions where available instead of raw PHP, this makes code more DRY.
In this case, `ee('Request')->get('csrf_token')` should be used

Also, this does not seem to respect `disable_csrf_protection` config setting.

system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
```diff
@@ -3305,7 +3305,7 @@ public function thread_rows($data, $is_announcement = false, $thread_review = fa
                         $temp,
                         array(
                             'lang:change_status' => ($row['status'] == 'o') ? lang('close_thread') : lang('activate_thread'),
-                            'path:change_status' => ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . ee()->functions->fetch_action_id('Forum', 'change_status') . '&amp;topic_id=' . $row['post_id'] . '&amp;board_id=' . $this->fetch_pref('board_id') . '&amp;trigger=' . $this->trigger,
+                            'path:change_status' => ee()->functions->fetch_site_index(0, 0) . QUERY_MARKER . 'ACT=' . ee()->functions->fetch_action_id('Forum', 'change_status') . '&amp;topic_id=' . $row['post_id'] . '&amp;board_id=' . $this->fetch_pref('board_id') . '&amp;trigger=' . $this->trigger .  '&amp;csrf_token=' . CSRF_TOKEN,
```
intoeetive: This does not seem to respect `disable_csrf_protection` config setting.

Besides that, we really should avoid passing CSRF token in GET request. It is supposed to be sent as header, so I guess you can just use that

intoeetive (changes requested):
@bakin1999 I have added some feedback

(and could you also sign the CLA here on GitHub)

system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
```diff
@@ -5931,7 +5931,8 @@ public function change_status()
         $viewpath = ($query->row('announcement') == 'n') ? 'viewthread/' : 'viewannounce/';
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
-        if (!array_key_exists('csrf_token',$_GET) ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) || $_GET['csrf_token'] != CSRF_TOKEN) {
+        $token = ee('Request')->get('csrf_token');
+        if ( (! bool_config_item('disable_csrf_protection') &&  !array_key_exists('csrf_token',$_GET)  )  ||  ! $this->_mod_permission('can_change_status', $query->row('forum_id')) ||  (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
```
intoeetive: I believe this does what it needs to do, however I see some ways how it can be improved:
- `disable_csrf_protection` can be checked only once
- `!array_key_exists('csrf_token',$_GET)` is like unnecessary - `ee('Request')->get('csrf_token')` would return empty value if the GET variable is not set

intoeetive (changes requested):

system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
bakin1999: Thanks, good catch. I have made the changes requested to make the logic more simplified.

intoeetive (commented):
I am still not sure about this. I believe there is a way so that CSRF token would be taken from request header, not from GET. 

@matthewjohns0n can you elaborate here?

system/ee/templates/_themes/forum/blue/forum_announcements/announcement.html
```diff
@@ -25,7 +25,12 @@
 <td>{lang:posted_on} {post_date format="%d %F %Y %h:%i %A"}</td>
 <td align="right">&nbsp;
 {if can_view_ip}<i>{lang:ip_address} {ip_address}</i>{/if}
-{if can_change_status}&nbsp;&nbsp;<span class="smallLinks">[ <a href="{path:change_status}">{lang:change_status}</a> ]</span>{/if}
+{if can_change_status}&nbsp;&nbsp;
+	<form action="{path:change_status}" method="POST">
+		<input type="hidden" name="csrf_token" value = {csrf_token}>
+		<input type="submit" name="submit" value = "{lang:change_status}">
```
intoeetive: I have noticed that in other places of forum all submit buttons also have `class="submit"` - should this be applied here as well?

Also, is there a reason for inconsistency of spaces around ` = `?

This comment is applicable to the other themes as well

intoeetive (changes requested):

system/ee/ExpressionEngine/Addons/forum/mod.forum_core.php
```diff
@@ -5930,7 +5932,8 @@ public function change_status()
         $viewpath = ($query->row('announcement') == 'n') ? 'viewthread/' : 'viewannounce/';
         $viewpath .= ($query->row('announcement') == 'n') ? $_GET['topic_id'] : $_GET['topic_id'] . '_' . $query->row('forum_id') ;
 
-        if (! $this->_mod_permission('can_change_status', $query->row('forum_id'))) {
+        $token = ee('Request')->post('csrf_token');
+        if (! $this->_mod_permission('can_change_status', $query->row('forum_id')) || (! bool_config_item('disable_csrf_protection') && $token != CSRF_TOKEN)) {
```
matthewjohns0n: I would change this to be a !== comparison instead of a != comparison.

matthewjohns0n (changes requested):
Only thing I would change is to make that comparison a !==. Otherwise this looks ready to merge.

obfuscode:
Matt and Yuri verbally approved this so I'm admin merging it despite "requested changes" still being active.

PR #1156, opened 2021-06-23
use Member service without instantiating to access constants 

intoeetive:
(required for PHP 5.6)


matthewjohns0n (approved):
Looks good 👍

PR #1201, opened 2021-07-20
delegate member for permission check when using file field

intoeetive:
delegate member for permission check when using file field on channel forms

EECORE-1203


robinsowell (approved):
Worked perfectly to fix an upgraded site that failed after getting on v6.

obfuscode (approved):

PR #1205, opened 2021-07-20
Issue no 1195

### LINKED ISSUE #1195: Add information on template processed to hook: template_post_parse
EE provides a hook [template_post_parse($final_template, $is_partial, $site_id)](https://docs.expressionengine.com/latest/development/extension-hooks/global/template.html#template_post_parsefinal_template-is_partial-site_id) that provides access to the parsed template output once processing is complete.
The hook includes a variable `$is_partial` described thusly: **TRUE if the current template is an embed or a layout**
This suggests that the output could relate to a partially processed template yet for that use case the hook does not provide any information about *which* template it is relating to... which makes it less useful.

Either amend the documentation to explain the purpose of the `$is_partial` parameter, or amend the hook to include information about which template the parse step that has completed was applied to.

The second option would be more helpful - as it would potentially allow the intercept of content after a partial template has been processed.

?

Depends on what solution is available.

- [X] I am capable and would like to work on implementation of this feature if it is considered.

jcogs-design:
## Overview

A change to code for `template_post_parse` hook in Template class.
Current code fires after on sub or layout templates are parsed (i.e. during parsing process if multiple templates involved) but provides no mechanism to access the partial template output.  The parsed template is only made available via the hook after the 'final' template parse.

The updated code makes two changes:

- first it adds an extra parameter to the `template_post_parse` hook (`$currentTemplateInfo = array()`) that contains the name of the most recently completed template parse and template group it is from;
- second it modifies the information provided via the $template parameter defined in the hook: when hook fires on a partial template parse it returns the current working version of template

Resolves [#1195](https://github.com/ExpressionEngine/ExpressionEngine/issues/1195).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [X] 🛁 Refactors existing code
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pulls/217


system/ee/legacy/libraries/Template.php
```diff
@@ -1889,7 +1914,7 @@ public function fetch_cache_file($cfile, $cache_type = 'tag', $args = array())
         $status = & $this->$status;
 
         // Bail out if this tag/template isn't set to cache
-        if (ee('LivePreview')->hasEntryData() or ! isset($args['cache']) or $args['cache'] != 'yes') {
+        if (! isset($args['cache']) or $args['cache'] != 'yes') {
```
intoeetive: what is the reason check for live preview was taken out?

system/ee/legacy/libraries/Template.php
```diff
@@ -232,15 +232,40 @@ public function fetch_and_parse($template_group = '', $template = '', $is_embed
...
+                );
+            }
+            else
+            {
+                $this->template = ee()->extensions->call(
+                    'template_post_parse',
+                    $this->template,
+                    ($is_embed || $is_layout), // $is_partial
+                    $site_id,
+                    $currentTemplateInfo
```
intoeetive: The code seems to be identical, just assigned to different variables. Should it be moved to the top and only the variable assigment included inside conditional?

system/ee/legacy/libraries/Template.php
```diff
@@ -232,15 +232,40 @@ public function fetch_and_parse($template_group = '', $template = '', $is_embed
...
+            $currentTemplateInfo = array();
+            if ($this->templates_loaded) // don't do this if we don't have any template info!
+            {
+                $currentTemplateInfo = array(
+                    'template_name' => $template,
+                    'template_group' => $template_group,
+                );
+            }
+            // Add a conditional to adjust what is returned by function depending on what kind of template we are dealing with
+            if ($this->final_template)
```
intoeetive: I think this might work wrong if `$this->final_template` is an empty string, no?

system/ee/legacy/libraries/Template.php
```diff
@@ -232,15 +232,40 @@ public function fetch_and_parse($template_group = '', $template = '', $is_embed
...
+                    $this->template,
+                    ($is_embed || $is_layout), // $is_partial
+                    $site_id,
+                    $currentTemplateInfo
+                );
+            }
         }
         //
         // -------------------------------------------
+    return;
```
intoeetive: What is the reason this `return` has been added any why it does not return any value?

intoeetive (changes requested):
@jcogs-design I have added few comments on your code.

In addition, we are trying to follow PSR-12 rules, which require opening brace after `if` to be on same line.

system/ee/legacy/libraries/Template.php
jcogs-design: Was not an intended change - perhaps it is a sync issue and I was working from an old version of code?

system/ee/legacy/libraries/Template.php
jcogs-design: Layout was simply for clarity, for sure it can be compacted.

system/ee/legacy/libraries/Template.php
jcogs-design: Simply to provide a point to add a debugging breakpoint.  Can be removed.

system/ee/legacy/libraries/Template.php
jcogs-design: On checking my copy of 6.dev the code is indeed missing - I seem to have got an old copy of the file somehow.  Will fix my local copy...

jcogs-design:
Have posted updated code that I hope addresses your concerns.

intoeetive:
@jcogs-design looks like your branch (or repo) was deleted before we could merged your PR. Is there any chance you could restore it?

jcogs-design:
I think it is back now.

intoeetive:
image
The tooltip says "this repository has been deleted"

jcogs-design:
No idea what is going on - so have created a new PR from the branch.

intoeetive:
Thank you!

Closing in favor of duplicate PR https://github.com/ExpressionEngine/ExpressionEngine/pull/1335

PR #1207, opened 2021-07-20
Add better support for the Toggle field as an Entry Manager column

litzinger:
Render a disabled toggle field in the column instead of a 1, 0, yes, or no. When editing an entry the user sees a Toggle field, so give them the same experience in the Entry Manager.

## Nature of This Change

- [x] Yes
- [ ] No

### DIFF
--- a/changelogs/patch.md
+++ b/changelogs/patch.md
@@ -7,6 +7,7 @@ ExpressionEngine uses semantic versioning. This file contains changes to Express
 Bullet list below, e.g.
 
    - Added <new feature>
+   - Changed how the Toggle field renders in a column in the Entry Manager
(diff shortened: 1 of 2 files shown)


litzinger:
Screen Shot 2021-07-20 at 12 28 08 PM

litzinger:
FYI, even without this fix, the Toggle field does not render the correct text value. In our case the value of the field was `"1"` or `"0"`, but the code is doing a strict type check of `=== 1` or `=== 0`, so even when the field is checked, it doesn't display "yes". 

I personally think this is a better fix all around.

intoeetive (approved):

PR #1213, opened 2021-07-25
Exclude EE comments from being processed with Template Debugger

### LINKED ISSUE #1086: Template Tag Debugger searches inside comment text
I had an alert from the Template Tag Debugger.
Upon clicking through I found it had flagged up a description of a template tag that was included within an EE comment  (i.e. text enclosed between `{!--` and `--}` constructs) within a template.
I don't know how easy it would be to implement, but probably the Template Tag Debugger should be skipping over the content of commented out text.

Using EE6.0.4

jcogs-design:
## Overview

Addresses issue 1086.
Remove any EE comments in template before scanning for tags (to prevent tags within comments being flagged as problems).
Removal can be destructive as we don't make any persistent changes to templates and getAllTags does not return line numbers of found tags

Resolves [#1086](https://github.com/ExpressionEngine/ExpressionEngine/issues/1086).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

No documentation implications.

### DIFF
--- a/system/ee/ExpressionEngine/Library/Advisor/TemplateAdvisor.php
+++ b/system/ee/ExpressionEngine/Library/Advisor/TemplateAdvisor.php
@@ -70,6 +70,12 @@ class TemplateAdvisor
                 ];
             }
 
+            // remove any EE comments in template first (to prevent tags within comments being flagged as problems)
+            // this can be destructive as we don't make any persistent changes to templates and getAllTags does not
+            // return line numbers of found tags
+            $regex_comments = '/(\{!--\s(?:.|\R)*?\s--\})/';
+            $template_data = preg_replace($regex_comments, '', $template_data);
+
             $tags_found = preg_match_all($regexp, $template_data, $keys, PREG_PATTERN_ORDER);
 
             $tmpl_info['details'][] = $keys;
--- a/changelogs/patch.md
+++ b/changelogs/patch.md
@@ -40,6 +40,7 @@ Bullet list below, e.g.
    - Fixed a bug with the error message when importing invalid channel sets
    - Fix a bug with debug variables not being filtered
    - Fix a bug with the importer converter filter
+   - Fix a bug ([#1086](https://github.com/ExpressionEngine/ExpressionEngine/issues/1086) where `{exp:` tags within EE comments would be found / reported by the tag debugger.
 
 EOF MARKER: This line helps prevent merge conflicts when things are
 added on the bottoms of lists


system/ee/ExpressionEngine/Library/Advisor/TemplateAdvisor.php
```diff
@@ -70,6 +70,12 @@ public function getAllTags()
                 ];
             }
 
+            // remove any EE comments in template first (to prevent tags within comments being flagged as problems)
+            // this can be destructive as we don't make any persistent changes to templates and getAllTags does not
+            // return line numbers of found tags
+            $regex_comments = '/(\{!--\s(?:.|\R)*?\s--\})/';
```
intoeetive, line 76: In `remove_ee_comments` function, EE is currently using simplier regex:
> "/\{!--.*?--\}/s"

Just wondering, why yours is more complex?

jcogs-design:
TBH I did not know that there was a function `remove_ee_comments` - it does not appear to be listed in documentation!
So regex was created to achieve aim of mod from scratch.  In essence the two versions are identical - in mine the multi-line is provided for by the non-capture group rather than by the 's' global modifier, the curly brackets are escaped (which should be done, as curly brackets can be used as functional delimiter in regex) and the white space characters made more generic.

Whether one is better than another is for you to decide I guess, but I would recommend that we raise an issue to 
- add some documentation for the `remove_ee_comments` function (unless it is documented and I just could not find it)
- update the regex in the `remove_ee_comments` function to escape use of curly brackets at the very least.

Separately there is also the issue of the `LD` and `RD` system constants (which are set to `{` and `}`).  I have seen some regex (e.g.  in the stash code) where these are used in preference to the characters themselves: I don't know if this is because in some universe the comment delimiter is changeable?  (in which case the constants should be used more widely) or if this is just legacy noise.

Anyhow - that's why.  Let me know what you want to do etc.

jcogs-design:
Just noticed that [this issue](https://github.com/ExpressionEngine/ExpressionEngine/issues/1086) is still present in 6.1.3... :(  
Is there anything I can do to nudge this along? 
Looks like the only issue is some conflict in the changelog.md file... which seems an odd reason to halt what would otherwise be a sensible enhancement.

intoeetive:
@jcogs-design - thank you for brining attention to this again!

@matthewjohns0n  - can you add your review here? I am mainly wondering - per the comments above - whether we should reuse existing function here (I think yes) and if yes, whether we should update the regex that it is using (I think no)

intoeetive:
Closing in favor of https://github.com/ExpressionEngine/ExpressionEngine/pull/1819

PR #1243, opened 2021-08-03
properly validate arrays when using Required rule

### LINKED ISSUE #1224: Required checkbox field submits without any options checked
I've been looking into an issue that I thought was Bloqs related, but apparently it affects normal checkbox fields too. If I set a checkbox field to required, I expect that in order for the field to be valid/required, at least 1 checkbox needs to be selected, however, it lets me submit a form without checking anything. This screen cap shows the Publisher add-on is also installed. To be sure Publisher wasn't causing any weird behavior I uninstalled it and the same issue persists.

I also would expect the client side validation to kick in, but it never seems to. I suspect it has something to do with the checkbox fields now being React components now?

checkbox-validation-not-working

intoeetive:
empty array and arrays that only consist of empty strings will not pass validation

@obfuscode / @matthewjohns0n / whoever will be reviewing this - I'm a little hestant whether removing empty strings from array is the right thing to do here. However our checkboxes UI have empty hidden field, which adds empty string to the submitted data array. That empty string is stripped when field is saved. So I think cleaning the posted array would be the right thing to do.

fixes #1224

EECORE-1240

### DIFF
--- a/system/ee/ExpressionEngine/Service/Validation/Rule/Required.php
+++ b/system/ee/ExpressionEngine/Service/Validation/Rule/Required.php
@@ -23,6 +23,15 @@ class Required extends ValidationRule
             $value = trim($value);
         }
 
+        if (is_array($value)) {
+            $value = array_filter($value, function($val) { 
+                return $val !== '';
+            });
+            if (empty($value)) {
+                return $this->stop();
+            }
+        }
+
         if ($value === '' or is_null($value)) {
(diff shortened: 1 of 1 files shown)


obfuscode (approved):

obfuscode:
I reviewed this - and I agree with Yuri that it's OK to clean empty values from an array when passing it to the validation / required service; I think this is the expected outcome.

PR #1257, opened 2021-08-06
Add extra parameters to `exp:channel:category_archive`

### LINKED ISSUE #1256: Extra parameters on `exp:channel:category_archive`
As I started to use `exp:channel:category_archive` for more than a simple archive, some parameters are missing, like `sticky` and order by most used categories.

robsonsobral:
## Overview

Add `sticky` parameter and `most_used_categories` value to `orderby` parameter to `exp:channel:category_archive`.

Resolves [#1256](https://github.com/ExpressionEngine/ExpressionEngine/issues/1256).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/231


intoeetive (changes requested):
I think it would be good to add support for other options, that we have with channel entries, for consistency
`sticky="yes"`
`sticky="no"`
`sticky="only"`
https://docs.expressionengine.com/latest/channels/entries.html#sticky

robsonsobral:
@intoeetive , being `no` the default value of `sticky`, what else do I need to add?! `yes` and `only` are already there. 😕

intoeetive:
You are absolutely right. Not sure how I missed that =(

I'll add some tests to it and then we'll bake it into some future release (likely 6.2.0)

Thank you very much!

intoeetive:
@robsonsobral I have added tests around `orderby="most_used_categories"` (did not add any for sticky parameter yet) and it looks like those are failing.
Can you have a look and fix? Or just remove this feature if this PR is mostly about `sticky` parameter (which still needs tests)

robsonsobral:
I've no idea which drug I used when I coded that. That query makes no sense. I gonna remove `most_used_categories`.

tests/cypress/support/templates/cats.group/archive-sorted.html
```diff
@@ -0,0 +1,29 @@
+<div class="default-linear">
+    {exp:channel:category_archive channel="about" style="linear" orderby="title"}
```
TomJaeger: Other then the div class name above.. I don't see a difference between this and whats on line 17... Should this be running a different test? @intoeetive

tests/cypress/support/templates/cats.group/archive-sorted.html
```diff
@@ -0,0 +1,29 @@
...
+<div class="default-nested">
+    {exp:channel:category_archive channel="about" style="nested" orderby="title"}
+        {categories}
+            <div>{category_name}</div>
+        {/categories}
+    {/exp:channel:category_archive}
+</div>
+
+<div class="title-linear">
+    {exp:channel:category_archive channel="about" style="linear" orderby="title"}
```
TomJaeger: Should this be different then whats on line 2?

TomJaeger (commented):
Left 2 comments for @intoeetive above. I also noticed that you made mention of adding tests around sticky above @intoeetive I think it would be good to get them in before we merge.

Issue #1220, opened 2021-07-28
Better CDN support


robsonsobral:
I had some discussion with a coworker about the throttling module. Today, it's something for an emergency only, because this kind of feature is something we manage at a different level of a site: in the front of the server. This also point to other anachronistic serving related things eecms has:

## Manipulated images always have the same name

The manipulated image files have the same name as the original filename, but on manipulation specific folders. Any modification can cause a cascade of stale content over the network. We could have partial MD5 hashes added to them to avoid that kind of issue.

## Assets management and cache busting

[Doug Black wrote an great article](https://u.expressionengine.com/article/compiling-web-assets-with-mix-and-webpack) about how to use [Laravel Mix](https://laravel-mix.com/) to compile assets, but, due to Mix own difficult to handle cache busting, the URLs are suffixed using a query string, which **isn't enough** for CDNs.

We are lucky! EECMS already has the solution for that: the [`{stylesheet="group/template}` path](https://docs.expressionengine.com/latest/templates/globals/stylesheet.html) tag! I'm going back to eecms after years of JS projects and I never imagined I would use `{stylesheet}` again. However, the build process become so complex that any help are welcomed. Over the time, I used Gulp, Parcel, webpack, pure NPM scripts and more tools than I can remember now. We can use Laravel Mix or whatever to compile/build the assets and just dump the final files at a "template.group" folder.

Could we have a `{script="group/template"}` path tag? Then, we could add the hash suffix to that!

## Search by POST and non cacheable URLs

Now, the part that I think it is the harder to improve: the search results templates.

- they aren't sharable;
- they aren't cacheable;

To fix that, the `{exp:search:search_results}` tag needs to be able to use parameters from URL. Using GET, the query string parameters also need to be alphabetically sorted, to make sure the full URL is always the same.

There's a legend that changes are being made to the search form. I'm curious to see them!

jcogs-design:
Re **Assets management and cache busting**

Adding a `{script}` tag to mirror the `{stylesheet}` tag is understandable: the `{stylesheet}` tag primarily offers a route to **avoid** template parsing of the content of the stylesheet and so some performance gain when serving the file from within EE, and I guess a similar benefit could be obtained for javascript files.

[...]

PR #1259, opened 2021-08-07
Feature/script tag

### LINKED ISSUE #1246: Site short name on `{stylesheet='template_group/css_template'}`
Sometimes, multiples sites share the same stylesheet. We could have a way to centralize it.

The `{stylesheet='template_group/css_template'}` could have a Site parameter like `{embed="site_short_name:template_group/template"}`.

For now, I'm using symbolic links.

Just like embeds and layouts can be loaded all from one website, we could have the same for stylesheets (and maybe a `{script='template_group/js_template'}` ❤️ )

robsonsobral:
## Overview

It implements:

- `{script="group/template"}`
- `{stylesheet="site:group/template"}`

Resolves [#1246](https://github.com/ExpressionEngine/ExpressionEngine/issues/1246).
Addresses [#1220](https://github.com/ExpressionEngine/ExpressionEngine/issues/1220).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/232


robsonsobral:
As use the **Stylesheet** library to serve JavaScript feels weird, I duplicated and renamed it to **Assets**. I guess, the original can be converted into a proxy to the new one, to keep it available to addons developers. However, I've no idea on the better path to take.

intoeetive:
I don't like duplicatig the code - I think we need to be as DRY as possible.

I think it's better to have something like `class Javascript extends Stylesheet` - however we already have Javascript library so this might bring extra confusion.

Maybe this would be good chance to take this code out of `legacy` folder and make it Library or Service. And then the legacy class would just inherit the new one

robsonsobral:
@intoeetive , do you mind help me a little more with the right places and naming conventions? Should I create an Assets Library? Is that fine?

intoeetive:
I think making this change is combination of adding new feature with code refactoring.

We generally try to avoid adding stuff into `legacy` folder (we still do that for fixes though or when full component rewrite would be required otherwise). I also don't like the name `Assets` (that's just a personal thing; I think it brings confusion with the image management add-on), I prefer something like `Resources`.

So if I would be you :) I'd probably do something similar to this:
1. Under `ExpressionEngine\Libraries` create `Resources/Stylesheet` and `Resources/Javascript` libraries, and have those registered in `app.setup.php`
2. I would also probably create something like `Resources/AbstractResource.php` that the libraries could inherit or partially reuse. I'd put there stuff like sending generic headers, or adding hashsum etc.
3. Keep the legacy `EE_Stylesheet`, but replace all functions code will calls to new library. Make it log deprecation message when called
4. Update EE code to use the new library
5. Make the other necessary changes to implement the new feature (in `Template` library etc)

Again, this is just how I personally view things here. I admit it can be approached differently as well

robsonsobral:
I prefer to follow your lead, as I'm a JS Developer and you'll be taking care of this code in the long run.

This way?
image

Here we go!

intoeetive:
Yes, something along those lines should be fine

(please also try to avoid force-pushing, that makes merging more difficult)

robsonsobral:
I'm sorry it have been taking so long, @intoeetive . Not even being an important feature for me, I've not found time enough for.

While working, I've found a bug which should not have to wait for me to be fixed! [The `Last-Modified` header](https://github.com/ExpressionEngine/ExpressionEngine/blob/72615a3deb991cb2cf103ef81ccbbf3447d38987/system/ee/legacy/libraries/Stylesheet.php#L80) submitted along CSS requests are wrong when using template files!

```php
/** -----------------------------------------
/**  Retrieve template file if necessary
/** -----------------------------------------*/
if (ee()->config->item('save_tmpl_files') == 'y') {
    ee()->load->helper('file');
    $basepath = PATH_TMPL . (
        isset($site) ? $site : (ee()->config->item('template_folder') ?: ee()->config->item('site_short_name'))
    ) . '/';
    $basepath .= $group . '.group/' . $row['template_name'] . '.' . $this->type;

    $str = read_file($basepath);
    $row['template_data'] = ($str !== false) ? $str : $row['template_data'];
    $row['edit_date'] = ($str !== false) ? filemtime($basepath) : $row['edit_date']; // this line is missing!
}
```

Without that last line, the submitted date is from database, not saved files.

robsonsobral:
I'm sorry, @intoeetive . I proved myself a total newbie PHP developer. I've saved my files in the right places, but I couldn't make them to work.

intoeetive:
I'll try to go with it. Can't promise when that will be happening though

robsonsobral:
I'm sorry, @intoeetive . And thank you!

robsonsobral:
I did it, @intoeetive ! It worked!

robsonsobral:
@intoeetive , 😥😥😥?

intoeetive:
I looked through the code, and I think you did awesome job here.

I did not check every tiny bit, but it does look very good to me.

However we need to have some automated tests written to accept this - do you happen to have sample template code, or add-on code, that's utilizing this new code and that we can use to test?

Also in few places there need to be spaces around `.` to make the code PSR-12 compatible.

Thank you very much!

robsonsobral:
Tests added!

The same tests can be used to test the [`integrity`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-integrity) attribute in the future.

robsonsobral:
I guess I should add tests for [trigger words](https://docs.expressionengine.com/latest/templates/globals/stylesheet.html#triggering-a-stylesheet-via-the-url) on URLs, like `https://example.com/css/channel/channel_css/`.

robsonsobral:
Now I've a **problem**. To add `js` as a [template trigger word](https://docs.expressionengine.com/latest/templates/globals/stylesheet.html#triggering-a-stylesheet-via-the-url) is a breaking change. May I ignore this? I never used this feature, so I don't know how important it is.

intoeetive:
Adding `js` as reserved word might be a problem if the system already has template or group named `js`. So need to think about how to handle this best - maybe have this as configurable value

robsonsobral:
Options I have:

1. to **not** add `js` as a template trigger word. I don't know how important trigger words are. For example, `xml` is a valid template type which doesn't have trigger word on URI. Also, there's an annoying dev [suggesting adding `json` as a valid template](https://github.com/ExpressionEngine/ExpressionEngine/issues/1755#issuecomment-1020138427). Does it mean it requires a trigger word as well? Who knows?
2. to put `{script}` tag behind a configuration. Which one?
3. to put `js` as a trigger word behind a configuration. Which one?;
4. to delay `{script}` tag until eecms v7 (and make me cry for my life).

Options 1 and 3 would make template type trigger words to be considered almost as a to-be-deprecated feature.

In any case, `/css/` needs to be added to tests and I'm going to do that.

intoeetive:
I've never used the "trigger word" feature myself, so don't know how important it is. My vote is option 1.

system/ee/ExpressionEngine/Library/Resource/Request.php
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            list($group, $name) = $ex;
+
+            if (false !== strpos($group, ':')) {
+                // if there's a site, let's get it and redefine $group
+                list($site, $group) = explode(':', $group, 2);
+            }
+
+            ee()->db->select('templates.template_data, templates.template_name,	templates.edit_date');
```
robsonsobral: Should I change this to use models, @intoeetive ?

system/ee/ExpressionEngine/Library/Resource/Request.php
```diff
@@ -0,0 +1,161 @@
...
+                ee()->db->join('sites', 'sites.site_id = templates.site_id');
+                ee()->db->where('sites.site_name', $site);
+            } else {
+                ee()->db->where('templates.site_id', ee()->config->item('site_id'));
+            }
+
+            $query = ee()->db->get();
+
+            if (0 == $query->num_rows()) {
+                var_dump('no rows');
```
intoeetive: Felling like var_dump should not be here

system/ee/ExpressionEngine/Library/Resource/Request.php
intoeetive: If this is pre-existing code, I would not bother and leave it as it is. As long as it's working, I don't feel like switching to models is strictly necessary

robsonsobral: May it be like this, @intoeetive ?

```php
$template = ee('Model')->get('Template')
    ->filter('template_name', $name)
    ->filter('template_type', $this->type)
    ->with('TemplateGroup')->filter('TemplateGroup.group_name', $group);

$template = isset($site)
    ? $template->with('Site')->filter('Site.site_name', $site)
    : $template->filter('site_id', ee()->config->item('site_id'));

$template = $template
    ->fields('template_data', 'template_name', 'edit_date')
    ->all()
    ->first();
```

intoeetive: This looks like valid code

robsonsobral: Thank you. I was afraid I was doing unnecessary queries or something like that. At first, I thought `->first() ` was enough to get data, but it looks like I need to use `->all()->first()`. Thank you!

robsonsobral:
~~Now, we're ready to merge!~~ Dammed! I had a dream about an issue! Let's check it!

robsonsobral:

- the new Resource Library is pretty DRY. It's using the Template Model and the Cache and Output classes;
- the Page Cache settings on CP never worked on CSS templates. The cache is enabled by default and can't be disabled. My suggestion is to enable it on template creation and make it work;
image
- CSS and JS files can be used across MSM sites using `site_name:` on path, in this case, the cache is shared among them;
- there's a public function to clear the cache of a resource;
- the cache is saved under nested directories like `resource_cache/site/group/template`. This is more like a suggestion/experience for how we can improve cache and isn't a blocking for this PR;
- the tests consider both CSS and JS templates from current and a secondary MSM site;
- the template files used for tests were moved to accommodate 2 sites;

tests/cypress/cypress/integration/design/template_of_resources.ee6.js
```diff
@@ -0,0 +1,71 @@
...
+		it('loads script resource template from the current site', function () {
+			cy.get("#second").should('have.css', 'background-color', 'rgb(255, 0, 255)')
+		})
+
+		it('loads stylesheet resource template from a different MSM site', function () {
+			cy.get("#third").should('have.css', 'background-color', 'rgb(255, 255, 0)')
+		})
+
+		it('loads script resource template from a different MSM site', function () {
+			cy.get("#forth").should('have.css', 'background-color', 'rgb(0, 0, 0)')
```
matthewjohns0n: Can you change all "forth" to "fourth"?

robsonsobral: I'm sorry!

system/ee/ExpressionEngine/app.setup.php
```diff
@@ -263,6 +264,18 @@
             );
         },
 
+        'Resource' => function () {
+            return new Resource\Request();
+        },
+
+        'Resource/Javascript' => function () {
+            return new Resource\Javascript();
```
matthewjohns0n, line 272: @robsonsobral This is where the `Javascript` class is instantiated and the "s" is lowercase. PHP class names are technically case insensitive, but they really should be consistent to avoid any confusion.

system/ee/legacy/libraries/Stylesheet.php
```diff
@@ -25,135 +27,7 @@ class EE_Stylesheet
...
-     * Send CSS
-     *
-     * Sends CSS with cache headers
-     *
-     * @access	public
-     * @param	string	stylesheet contents
-     * @param	int		Unix timestamp (GMT/UTC) of last modification
-     * @return	void
-     */
-    public function _send_css($stylesheet, $modified)
```
matthewjohns0n, line 124: I'm concerned about removing this function because it is public. Removing this will break any addon currently using it, so I'd like to see us get it back in there.

matthewjohns0n (changes requested):

system/ee/legacy/libraries/Stylesheet.php
robsonsobral: It's done, but I can't run tests:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'Site_sites.site_color' in 'field list':
SELECT Site_sites.site_id as Site__site_id, Site_sites.site_label as Site__site_label, Site_sites.site_name as Site__site_name, Site_sites.site_description as Site__site_description, Site_sites.site_color as Site__site_color, Site_sites.site_bootstrap_checksums as Site__site_bootstrap_checksums, Site_sites.site_pages as Site__site_pages FROM (`exp_sites` as Site_sites) WHERE ( `Site_sites`.`site_id` IN (1) ) LIMIT 18446744073709551615
```

intoeetive:
I have fixed the tests. Letting you guys handle everything else!

system/ee/legacy/libraries/Stylesheet.php
robsonsobral: Yuri fixed it! Should we add a unit test to it, @matthewjohns0n ?

matthewjohns0n: It would be good to have this unit tested, is that something you would be able to add?

robsonsobral: Not yet. I'm a PHP newbie.

matthewjohns0n: All good! While it would be good to add unit tests, I'm alright with approving it for now since we're trying to get 6.3.0 ready to release and would like to get this in that release.

matthewjohns0n (approved):

system/ee/legacy/libraries/Stylesheet.php
robsonsobral: So soon? Coooooooooool!

PR #1262, opened 2021-08-09
CLI updates

intoeetive:
remove dot from submitted data when generating add-on with CLI

EECORE-1258

add slashes when writing strings during add-on generation w/CLI

EECORE-1257


obfuscode (approved):

matthewjohns0n:
I believe this breaks a lot of generation commands. I'm seeing this now:
Screen Shot 2021-08-09 at 1 49 08 PM

intoeetive:
The fix for this is coming in https://github.com/EllisLab/ExpressionEngine-Private/pull/643

PR #1271, opened 2021-08-11
Fix issue with cross platform newline matching

matthewjohns0n:
## Overview

Fixes issue with  cross platform newline matching. Lines  in  the  addon setup  file were not getting  replaced properly when  using  windows. 

Resolves EEPRO-247

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
+++ b/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
@@ -425,7 +425,7 @@ class AddonGenerator
 
     private function clearLine($string, $contents)
     {
-        return str_replace($string . "\n", '', $contents);
+        return preg_replace("~" . preg_quote($string) . "~", '', $contents);
     }
 
     public function slug($word)


system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
```diff
@@ -425,7 +425,7 @@ private function erase($string, $contents)
 
     private function clearLine($string, $contents)
     {
-        return str_replace($string . "\n", '', $contents);
+        return preg_replace("~" . preg_quote($string) . "~", '', $contents);
```
intoeetive, line 428: I'm just being curious here. Is `~` serving just as delimiter here? In all other places we use `/`, why this one is different.

FWIW, in another place I've used PHP_EOL constant and that seems to have solved it.

Lastly, would you mind making this PR in EE-Private repo, as that's where we're working now

@matthewjohns0n

matthewjohns0n: Yeah we definitely don't need the delimiter to be `~`. That came from me testing several options. This issue came up because we are using PHP_EOL in the private repo. I'm not sure why the repos are out of sync though. 

I'm closing this and resubmitting in private repo.

PR #1281, opened 2021-08-17
EECORE-1312 Get Variable Modifiers working for comments again

### LINKED ISSUE #1264: {comment} variable within {exp:comment:entries} tag does not work with variable modifiers
EE 6.0.6

This code does not work:

`{exp:comment:entries}
   {comment:limit characters='200'}
{/exp:comment:entries}`

The problem: the output is raw EE template code instead of parsed HTML.

While this code does work:

`{exp:comment:entries}
   {comment}
{/exp:comment:entries}`

That is, I can display the fulltext of a {comment} on a frontend template, but can't apply a character limit.

bakin1999:
## Overview
EECORE-1312
issue id #1264 

Mod.comment had changes that has it use function parse_variables_row instead of function parse_variables.  

parse_variables_row's condition to run getModifiedVariables was not being met. (getModifiedVariables was called directly in parse_variables.)

Resolves: https://github.com/ExpressionEngine/ExpressionEngine/issues/1264
## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -3504,7 +3504,13 @@ class EE_Template
         }
 
         // same with modified, sometimes devs run this method themselves instead of a full parse_variables()
-        if ($this->modified_vars === false) {
+        /*
+        Reason for if condtion change:
+        mod.comment switched to using this function instead of parse_variables function and now no Variable Modifiers work in comments
+        i.e. {comment:limit characters="20"} will not parse correctly. The getModifiedVariables() is never run here and is run without any checks
+        in parse_variables().  Keep the $this->modified_vars === false condition because of comment above.
+        */
+        if (empty($this->modified_vars) || $this->modified_vars === false) {
             $this->modified_vars = $this->getModifiedVariables();
         }
 
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -3504,12 +3504,6 @@ class EE_Template
         }
 
         // same with modified, sometimes devs run this method themselves instead of a full parse_variables()
-        /*
-        Reason for if condtion change:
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/Template.php
```diff
@@ -3504,7 +3504,13 @@ public function parse_variables_row($tagdata, $variables, $solo = true)
...
 
         // same with modified, sometimes devs run this method themselves instead of a full parse_variables()
-        if ($this->modified_vars === false) {
+        /*
+        Reason for if condtion change:
+        mod.comment switched to using this function instead of parse_variables function and now no Variable Modifiers work in comments
+        i.e. {comment:limit characters="20"} will not parse correctly. The getModifiedVariables() is never run here and is run without any checks
+        in parse_variables().  Keep the $this->modified_vars === false condition because of comment above.
+        */
+        if (empty($this->modified_vars) || $this->modified_vars === false) {
```
intoeetive, line 3507: I think `$this->modified_vars === false` will always return true if the first condition is met (`empty($this->modified_vars)`). Or do I miss something here.

Also I don't think we need that comment in the code. It is descriving the change and not code behaviour, and probably would work better if converted to commit message

obfuscode: I think this is fine. We could get rid of the `|| $this->modified_vars === false` because I think it's the opposite, that `empty($this->modified_vars)` would be true if `$this->modified_vars === false`. Either way, I don't see an issue leaving this here.

obfuscode (approved):

PR #1300, opened 2021-08-25
prevent duplicate category_url_title accross category groups

### LINKED ISSUE #390: Inbuilt check to prevent duplicate category_url_title
Where you have multiple category groups assigned to a channel it's possible to use the same category_url_title in each group. This can then cause problems displaying the correct content in a category view.

Prevent a duplicate category_url_title from being used. If creating or renaming a category_url_title EE should flag this up, i.e. "The url title is already in use" and prevent saving until it has a unique value.

intoeetive:
fixes #390

EECORE-1336

### DIFF
--- a/system/ee/ExpressionEngine/Model/Category/Category.php
+++ b/system/ee/ExpressionEngine/Model/Category/Category.php
@@ -68,7 +68,7 @@ class Category extends ContentModel
 
     protected static $_validation_rules = array(
         'cat_name' => 'required|noHtml|xss',
-        'cat_url_title' => 'required|alphaDash|unique[group_id]',
+        'cat_url_title' => 'required|alphaDash|unique[site_id]',
         'cat_description' => 'xss',
         'cat_order' => 'isNaturalNoZero',
         'parent_id' => 'validateParentCategory'


ops-andy:
@intoeetive this needs to be attached to a milestone. Please also update the corresponding Jira ticket. thanks

system/ee/ExpressionEngine/Model/Category/Category.php
```diff
@@ -68,7 +68,7 @@ class Category extends ContentModel
 
     protected static $_validation_rules = array(
         'cat_name' => 'required|noHtml|xss',
-        'cat_url_title' => 'required|alphaDash|unique[group_id]',
+        'cat_url_title' => 'required|alphaDash|unique[site_id]',
```
TomJaeger, line 71: Unfortunately I think this is a breaking change.  It's been long standing that EE can have duplicate categories in different groups, but not the same.

TomJaeger:
Unfortunately it's been long standing that you can use the same category in multiple category groups in EE (just not the same group).  This would be a breaking change.  We should consider adding a small notification when it matches an existing one though, so people are aware that their creating a category in another category group.

PR #1307, opened 2021-08-26
fixed `$.browser` is undefined for channel form, EECORE-1299

Yulyaswan:
fixed `$.browser` is undefined when having textarea with buttons in channel form

### DIFF
--- a/themes/ee/asset/javascript/src/jquery/plugins/markitup.js
+++ b/themes/ee/asset/javascript/src/jquery/plugins/markitup.js
@@ -27,6 +27,47 @@
 // ----------------------------------------------------------------------------
 */
 (function($) {
+	// =============================================
+	// For backwards compatibility: adding $.browser
+	// from: https://github.com/jquery/jquery-migrate
+	// =============================================
+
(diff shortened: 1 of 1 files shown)


themes/ee/asset/javascript/src/jquery/plugins/markitup.js
```diff
@@ -27,6 +27,47 @@
 // ----------------------------------------------------------------------------
 */
 (function($) {
+	// =============================================
+	// For backwards compatibility: adding $.browser
+	// from: https://github.com/jquery/jquery-migrate
+	// =============================================
+
+	$.uaMatch = function( ua ) {
```
intoeetive, line 35: Do we need to add a check like `if ( !$.uaMatch ) {` here? (not sure, just asking)

Yulyaswan: no, we don't. This function will work if $.browser is undefined

intoeetive (approved):
Works good on my local test, no JS error

PR #1312, opened 2021-08-27
fixed JS error in ordering relationships fields, EECORE-1277

Yulyaswan:
the error appeared after:
1. the image has been upload into the grid field
2. were added 2 items to the relationship field 
3. were re-ordered these relationship fields

### DIFF
--- a/themes/ee/asset/javascript/src/jquery/ui/jquery.ui.sortable.js
+++ b/themes/ee/asset/javascript/src/jquery/ui/jquery.ui.sortable.js
@@ -426,8 +426,7 @@ return $.widget("ui.sortable", $.ui.mouse, {
 		}
 
 		//Call callbacks
-		// hide to prevent the script from breaking when the order is changed
-		// this._trigger("sort", event, this._uiHash()); 
+		this._trigger("sort", event, this._uiHash()); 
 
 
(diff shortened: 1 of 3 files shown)


themes/ee/asset/javascript/src/jquery/ui/jquery.ui.sortable.js
```diff
@@ -427,7 +426,9 @@ return $.widget("ui.sortable", $.ui.mouse, {
 		}
 
 		//Call callbacks
-		this._trigger("sort", event, this._uiHash());
+		// hide to prevent the script from breaking when the order is changed
+		// this._trigger("sort", event, this._uiHash());
```
intoeetive: I am not sure this code can be commented out. I believe we have some scripts that rely on "sort" event - unless this has been checked

intoeetive (changes requested):

themes/ee/asset/javascript/src/jquery/ui/jquery.ui.sortable.js
obfuscode: This was reverted by 3b15de4. Resolving conversation.

obfuscode (approved):

intoeetive (approved):

PR #1313, opened 2021-08-27
fixed js error that appeared after saving consent in cp profile, EECORE-1284

Yulyaswan:
fixed js error that appeared after saving consent in cp profile


intoeetive (approved):

intoeetive (approved):
No error shown anymore, approving

PR #1322, opened 2021-09-02
fixed add-ons view for mobile screen, EECORE-1362

Yulyaswan:
changed styles to prevent the block from going outside the parent element


intoeetive (approved):
Tested, is looking good

PR #1323, opened 2021-09-02
fixed issue with clipping dropdown in modal window, EECORE-1344

### LINKED ISSUE #1309: Dropdown clipped on modal
The dropdown menu of folders is clipped by modal wrapper when there's no files enough to make it greater.

We need to remove `overflow: hidden;` from `common.css`, or to add a `min-height` to it.

image

Yulyaswan:
resolves #1309


intoeetive:
Just noting that this needs to be tested on all kinds on modals that we have (including triggered from other fields like RTE etc) on desktops and mobiles

intoeetive (dismissed):

intoeetive (approved):

Issue #282, opened 2019-08-09
Required field error message not showing for a Grid inside a Fluid Field


pyrobob83:
**Description of the problem**
Required fields inside a Grid inside a Fluid Field do not show an indication of the error when the entry fails to save because of missing data.  This only happens with a Grid inside a Fluid Field - assigning the same Grid field directly to a channel does show the error messages:

Grid directly assigned to channel:
![grid-errors-when-assigned-directly-to-channel](https://user-images.githubusercontent.com/8193519/62781146-51f5f100-ba85-11e9-9b71-e44ed76b5280.jpg)

Grid assigned to Fluid Field:
![grid-errors-when-assigned-to-a-fluid-field](https://user-images.githubusercontent.com/8193519/62781147-51f5f100-ba85-11e9-94b6-4a9f00e88a4e.jpg)

**How To Reproduce**
1. Create a Grid field with required columns - examples I've tested are text, file, Wyvern
2. Create a Fluid Field, assign the Grid field from #1 and assign that Fluid Field to any channel
3. Add or edit a channel entry, add the Grid field to the entry and attempt to save without entering any data
4. Notice you see no indication of what fields are causing the error

**Environment Details:**
 - EE Version: 5.2.6
 - PHP Version 7.3.6
 - MySQL Version 5.5.5-10.3.9-MariaDB
 - OS:  Linux
 - Web Server:  Apache

intoeetive:
This is also an issue with EE6

PR #1329, opened 2021-09-07
fixed Required field error message for a Grid inside a Fluid field, EECORE-1032

Yulyaswan:
an error message is shown if the user clicks on a required field but leaves it empty


themes/ee/asset/javascript/src/cp/form_validation.js
```diff
@@ -236,7 +236,6 @@ EE.cp.formValidation = {
 	 * @param	{jQuery object}	form	jQuery object of form
 	 */
 	_errorsExist: function(form) {
-
```
intoeetive: @Yulyaswan when possible, try to avoid pushing changes like adding/removing empty line or space (if that does not improve code readability). This will help to mantain cleaner Git history

themes/ee/asset/javascript/src/cp/form_validation.js
```diff
@@ -375,7 +380,7 @@ EE.cp.formValidation = {
 
 		// Validation error
 		} else {
-
+			console.log("fieldset", fieldset);
```
intoeetive: I think this should not be here

themes/ee/asset/javascript/src/cp/form_validation.js
```diff
@@ -326,9 +326,14 @@ EE.cp.formValidation = {
 			grid = true;
 		}
 
+		if (fieldset.find('.fluid').size()>0)
```
intoeetive: While we don't have any coding convertions for javascript code, the existing style is to have spaces around `>` so please add those.

themes/ee/asset/javascript/src/cp/form_validation.js
```diff
@@ -326,9 +326,14 @@ EE.cp.formValidation = {
...
+		{
+			container = field.parents('td');
+			grid = true;
```
intoeetive, line 346: I did not test this, but it does not seem to be that the sole condition `fieldset.find('.fluid').size()` is enough to assume that `grid = true;`

intoeetive (changes requested):
@Yulyaswan please see some comments here.

I did not mark repeating issues, so please just walk through the changes yourself

Yulyaswan:
fixes #282
EECORE-1032, EECORE-1524

themes/ee/asset/javascript/src/cp/form_validation.js
obfuscode: I did test this and it works. I think the `grid = true` (which was there before) is a bit of a misnomer because I think this might work for other fieldtypes. @Yulyaswan There is a "grid-field" class in that tree that you can use to identify it instead.

obfuscode (changes requested):
Update the selector for the grid = true conditional (See inline comment)

obfuscode (approved):

PR #1330, opened 2021-09-07
fix the bug when CP URL was wrong sometimes after showing idle modal

intoeetive:
EECORE-1006


ops-andy:
@intoeetive can we move this to milestone 6.1.2 please.

obfuscode (approved):

PR #1334, opened 2021-09-08
add a way to update signaure and manage sig settings directly from cp…

### LINKED ISSUE #1266: After Enabling Signatures, No Settings for Signatures in Admin Present
### ISSUE
I have enabled member signatures as specified in documentation: https://docs.expressionengine.com/latest/general/system-configuration-overrides.html#allow_signatures

After enabling I do not see a way of editing the signature in the members profile settings or anywhere else in the admin settings panel.

### EXPECTED BEHAVIOR
I would expect that after enabling this feature there would be settings in the member profile area:
Screen Shot 2021-08-09 at 10 57 49 AM

### ENV
- PHP 7.4
- mysql 5.7
- ExpressionEngine 6.0.6

bakin1999:
…. For both text and image signatures. Add signature settings for image size ect

## Overview

Currently, there is no way to update signature or signature settings directly from the CP. Users have to go to the forum mod to do this which is not ideal since signatures can be used in other places (ex. They can be used in comments).  This will update the user profile (checking for if signatures are enabled).

Resolves [#1266](https://github.com/ExpressionEngine/ExpressionEngine/issues/1266).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/238

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Members/Profile/Settings.php
+++ b/system/ee/ExpressionEngine/Controller/Members/Profile/Settings.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This source file is part of the open source project
  * ExpressionEngine (https://expressionengine.com)
@@ -24,7 +25,7 @@ class Settings extends Profile
     {
         $id = ee()->input->get('id');
 
-        if ($id != $this->session->userdata['member_id'] && ! empty($id)) {
+        if ($id != $this->session->userdata['member_id'] && !empty($id)) {
             parent::permissionCheck();
         }
     }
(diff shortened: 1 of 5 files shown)


intoeetive:
Honestly I think that currently the signatures are only being used in forums. So we either need to make those available in template, or remove from CP as a whole.

I'll let @TomJaeger say

intoeetive:
Because the Signatures are currently only being used on Forums frontend, me and @TomJaeger agreed it is sufficient ho have the setting available there (and it is now). We do need however to update the documentation to say that Signatures are for Forums only - @ops-andy FYI

PR #1338, opened 2021-09-10
Bug/6.dev/drop column row size

robinsowell:
There is a very rare issue in the upgrade from v5-6 where you can get a MySQL error:

```
SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs:

ALTER TABLE `exp_member_groups` DROP `can_delete_all_comments`
```

I can't for the life of me figure out why dropping columns would cause a row size error.  In both cases I've seen it, it was on MariaDB and both cases threw the error in the same spot- ALTER TABLE `exp_member_groups` DROP `can_delete_all_comments`.

Google failed me.  BUT I figured we really should have a batch column drop, it would save a bit of memory and it makes sense to have, and by happy coincidence, it solves this weirdo bug.  

So- new batch column delete function added and implemented in the v5 upgrade that was hitting the row size error.

The one thing that may be missing is clean roll back functionality.  The automatic roll back failed in this spot anyway, but I did nothing to make this know how to add the columns back.  So it would probably fail here too.  I could use some thoughts on that issue.  So pinging @matthewjohns0n for review and eyeballs on the roll back issue.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/installer/updates/ud_6_00_00_b_1.php
+++ b/system/ee/installer/updates/ud_6_00_00_b_1.php
@@ -596,9 +596,7 @@ class Updater
             ee()->db->insert_batch('permissions', $insert);
         }
 
-        foreach ($permissions as $permission) {
-            ee()->smartforge->drop_column('member_groups', $permission);
-        }
+        ee()->smartforge->drop_column_batch('member_groups', $permissions);
     }
 
     private function reassignChannelsToRoles()
--- a/system/ee/legacy/database/DB_forge.php
+++ b/system/ee/legacy/database/DB_forge.php
@@ -269,10 +269,10 @@ class CI_DB_forge
     }
 
     /**
-     * Column Drop
+     * Batched Column Drop
      *
      * @access	public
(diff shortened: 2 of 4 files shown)


intoeetive (changes requested):
@robinsowell I think this needs to use our PSR-12 formatting in some palces (use spaces instead of tabs, spaces around '.' etc.

Also, is this needed for 6.1 or can go into 6.2?

robinsowell:
@intoeetive It can definitely go in 6.2.  It's a super rare issue.  And I'll hit the psr changes.

obfuscode (approved):

PR #1339, opened 2021-09-10
Return and after should never be an array

matthewjohns0n:
## Overview

Fixes a bug where return value wasnt working properly.

Resolves EECORE-1345

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


intoeetive (changes requested):
I'm not feeling confident about this fix. I think we have few cases where `return` IS array, and we're just throwing them out, forcing redirect to homepage (`''`). I don't think we could do it.

intoeetive (approved):

PR #1349, opened 2021-09-22
Feature/6.x/min max relationship field, EECORE-1326, EECORE-1125, EECORE-1393

### LINKED ISSUE #40: Relationship field should have a min/max option
It would be nice if the relationship field could be set to have a minimum and/or maximum number of selections.  That way, you could cap how many choices are made -- this would be useful for enforcing editorial control on slideshows, sidebars, related articles, etc. etc.

This would require error checking in the entry form, modification to the javascript, and changes to the field settings. But it's not a huge change, it doesn't change any underlying data structures, etc.

To get this effect currently, I've used grid limited to 5 rows, with a relationship field in it. But that's kind hack-y and creates complicated template code and additional load time on the entry page, as well as it being annoying to the site admins.  You can use multiple relationship fields, in some cases, to enforce a limit, but that's not always possible, and it makes re-ordering a pain.  You can instruct people not to choose too many options (but they never listen).

I don't think this will require much documentation change, if any.  It's not going to be confusing for users or site builders, it's just a nice additional sub-feature.

Yulyaswan:
added  minimum and maximum number of selections for relationship field
fixes #40 
EECORE-1326, EECORE-1125, EECORE-1393

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -531,7 +531,9 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
             'no_related' => ['text' => lang('no_entries_related')],
             'select_filters' => $select_filters,
             'channels' => $channel_choices,
-            'in_modal' => $this->get_setting('in_modal_context')
+            'in_modal' => $this->get_setting('in_modal_context'),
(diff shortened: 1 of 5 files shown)


intoeetive:
@ops-andy semanically, this is adding new feature (and a visible one with UI elements), so it can't go into patch release. Setting milestone to 6.2.0

matthewjohns0n (approved):

obfuscode (approved):

PR #1355, opened 2021-09-23
Sort sites menu

robsonsobral:
## Overview

I'm just sorting menu of sites. I don't think a single line deserves even a changelog.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/View/_shared/sidebar/navigation/navigation.php
+++ b/system/ee/ExpressionEngine/View/_shared/sidebar/navigation/navigation.php
@@ -10,6 +10,7 @@
 
 					<?php
                     if (!empty($sites)) :
+                    asort($sites);
                     foreach ($sites as $site_id => $site_name):
                         if ($site_id != ee()->config->item('site_id')) :
                     ?>
--- a/system/ee/ExpressionEngine/View/_shared/sidebar/navigation/navigation.php
+++ b/system/ee/ExpressionEngine/View/_shared/sidebar/navigation/navigation.php
@@ -10,7 +10,6 @@
 
 					<?php
                     if (!empty($sites)) :
-                    asort($sites);
                     foreach ($sites as $site_id => $site_name):
                         if ($site_id != ee()->config->item('site_id')) :
                     ?>
--- a/system/ee/legacy/libraries/Session.php
+++ b/system/ee/legacy/libraries/Session.php
@@ -1307,6 +1307,8 @@ class EE_Session
 
         $this->userdata['assigned_sites'] = $assigned_sites->all()
             ->getDictionary('site_id', 'site_label');
+
+        asort($this->userdata['assigned_sites']);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


system/ee/ExpressionEngine/View/_shared/sidebar/navigation/navigation.php
```diff
@@ -10,6 +10,7 @@
 
 					<?php
                     if (!empty($sites)) :
+                    asort($sites);
```
intoeetive: Could it be better to sort the sites when those are fetched? That would also give more freedom if we'd want to make sorting configurable (whether we want to have it by ID or name, etc)

robsonsobral:
Do you want me to change it [here](https://github.com/ExpressionEngine/ExpressionEngine/blob/49d282f67c4ff5e7a79830bc8b94550b31013383/system/ee/legacy/libraries/Session.php#L1284)?

intoeetive:
@robsonsobral I think that might be the best. It will make the list already sorted when it's being used elsewhere

robsonsobral:
It's done, @intoeetive . However, I didn't like the result. I guess the previous code is safer.

intoeetive:
Can you share why you think this code is not safe?

(also, would it be better if we sort before saving to `userdata`, not after?)

robsonsobral:
I was in doubt because at some point, somebody chose to [reverse sort](https://github.com/ExpressionEngine/ExpressionEngine/blob/49d282f67c4ff5e7a79830bc8b94550b31013383/system/ee/legacy/libraries/Session.php#L1289) the sites and I didn't know why. So, I took sometime digging the file history and found it's a regression caused on a [commit](https://github.com/ExpressionEngine/ExpressionEngine/commit/203b551c98917782351408fb5330a2ac7bfca5e7#diff-dd0cee0b8a116d64bba79702feed27ea0c571adc834c3f4ae338a06774b4cf76R1373) by @obfuscode .

Can you, please, check if you agree to change the sorting order of the site here, @obfuscode ? (it's nothing personal!)

obfuscode:
Do you have a link to a more specific commit @robsonsobral ? That commit link just goes to where Matt merged in 3,690 files? Nothing personal taken! I want to make sure we get the best version of the code in as well! :)

obfuscode:
So I do see where in that link the change was made but not any reason why `desc` was specifically added.

robsonsobral:
To me, it looks like a simple mistake when adapting the code to the PSR standard. If so, this fix can be rebased and merged.

(It's really weird to see the sites in reversed order. Why?!)

intoeetive (approved):
I have looked through the Git history and I was only able to track this down to merge commit by EllisLab back in 2019: https://github.com/EllisLab/ExpressionEngine-Private/commit/3b8aba2e586b4b5607b07760fe85507af706ec9d
It does not show specifically this change and it does not show the reasoning. I believe this was done my mistake.
Approving the PR

PR #1375, opened 2021-10-05
Add ability to use class/method to ACT requests

stephengalbraith:
## Overview

This PR modifies existing code to add the ability for developers to access ACT requests using the Class/Method reference instead of making calls to fetch the action id. It also moves the conditional to check if the module is installed outside the main conditional (the current ordering would only check if the ACT was not numeric, which would ignore many calls.

Essentially what I've done is allowed calls to be made by: ?ACT=class/method

If this is accepted, I can also submit a PR to the docs talking about this change.

I am open to discussions or any other ideas on this as well.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Actions.php
+++ b/system/ee/legacy/libraries/Actions.php
@@ -38,146 +38,156 @@
  */
 class EE_Actions
 {
-    /**
-     * Constructor
-     *
-     * Loads the class and calls the method associated with
-     * a particular action request
-     *
-     */
-    public function __construct($can_view_system = false, $callback = null)
-    {
-        // Define special actions
-        // These are actions that are triggered manually
-        // rather than doing a lookup in the actions table.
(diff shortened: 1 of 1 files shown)


intoeetive (changes requested):
@stephengalbraith can you update the file to use the previously existing code formatting? Because of formatting change (tabs vs spaces probably?) GitHub is essentially marking each line as changed, making it impossible to spot the real changes made

stephengalbraith:
@intoeetive I've done a new PR - the changes seem to reflect better. I did re-arrange quite a bit. See #1424  - I'll close this one.

PR #1387, opened 2021-10-06
Feature/6.x/select within grid, EECORE-1235

### LINKED ISSUE #1172: [EE6] Select dropdown fields within grid and file grid fields should default to unselected
When creating a new channel entry, standalone select dropdown fields default to being unselected — with an additional "Choose wisely" meta-option added to show this.

Select dropdown fields within grid and file grid fields, on the other hand, default to the first actual option. This behaviour is inconsistent with the above, and, in the absence of a prompt which directs users to make a selection, makes it more likely that default selections will be left in place, perhaps incorrectly.

For consistency, and to enforce a conscious selection, select dropdown fields within grid and file grid fields should also default to being unselected.

(A possible extension to this would be to allow a default selection to be specified for select dropdowns. But defaulting in practice to the first option is unhelpful UI.)

EE6.0.6

Yulyaswan:
removed restricted for select dropdown within grid field
for now dropdown always displayed above the grid

fixes #1172
EECORE-1235


intoeetive (approved):
Since it´s failing the tests, not including into 6.1.1

intoeetive (changes requested):
Please see details in Jira: https://packettide.atlassian.net/browse/EECORE-1235?focusedCommentId=14975

intoeetive (approved):

PR #1410, opened 2021-10-12
Patch/search bad queries

### LINKED ISSUE #1403: Unnecessary JOINs on search module
The performance of Search module isn't that by itself. However, there are some unnecessary `OR` and `JOIN`s. For example, the comments table is [left joined even with `search_in` not set to `everywhere`](https://github.com/ExpressionEngine/ExpressionEngine/blob/1b4080febb757f07edb7221edd15dde5ee42ae80/system/ee/ExpressionEngine/Addons/search/mod.search.php#L501) and the categories are also [always left joined](https://github.com/ExpressionEngine/ExpressionEngine/blob/1b4080febb757f07edb7221edd15dde5ee42ae80/system/ee/ExpressionEngine/Addons/search/mod.search.php#L504).

I did little tweaks which fix that, but I think the best is to rewrite everything using [ActiveRecord](https://docs.expressionengine.com/latest/development/legacy/database/active-record.html#selecting-data).

The [sql_andor_string()](https://docs.expressionengine.com/latest/development/legacy/libraries/functions.html#sql_andor_stringstr-field-prefix---null--false) function also needs a variant which uses `IN` instead of `OR`.

I already started to refactor, but I'm going to need some help on tests.

- [ ] I am capable and would like to work on implementation of this feature if it is considered.

### LINKED ISSUE #1549: Undefined index: search_flood_control
EE 6.1.5

When debug is on and a non-superuser uses a search-form on the frontend PHP throws a notice:

`Notice
Undefined index: search_flood_control
ee/ExpressionEngine/Addons/search/mod.search.php, line 97

Severity: E_NOTICE
Warning
Cannot modify header information - headers already sent by (output started at ee/legacy/core/Exceptions.php:120)
ee/legacy/libraries/Functions.php, line 458

Severity: E_WARNING`

robsonsobral:
## Overview

Fix performance issues on Search Module

Resolves [#1403](https://github.com/ExpressionEngine/ExpressionEngine/issues/1403).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


robsonsobral:
@intoeetive , I don't understand the reason the Build Testing is falling. I didn't touch the directory `build-tools/builds`. Any tips?

intoeetive:
Looks like that's because it's forked repo and it does not have token to access Pro-specific tests. We'll need to work on workaround

robsonsobral:
Oh! There's something I need to ask! Which value shoud I use for all sites? I used `site="not none"` for development, but I know it's ugly. Any suggestions, @intoeetive ? @TomJaeger ?

intoeetive:
What is the intended usage of this parameter / value?

robsonsobral:
(Ops! I fixed a typo!)

It's intended to disable site filtering on `exp:channel:entries` and `exp:search:results`. `not some_unsed_site_name` works. However, shouldn't we have an official value for it?

intoeetive:
so is that like "all sites", or "current site"? I don't think `none` is good here, since technically you can set your site short name to `none`

robsonsobral:
Yeah! Bad name. So, what do you suggest? `ALL SITES`? I think it's safer than `ALL_SITES` as the name of sites can't have spaces.

intoeetive:
`ALL_SITES` might be not bad idea. It might be even worth making it available globally.

You may also consider using another param here, like `all_sites="yes"` (but personally I like `ALL_SITES` more)

robsonsobral:
I agree, but... Is `ALL_SITES` backward compatible?

image

Is it possible to check for conflicting names and report them on update?

intoeetive:
It looks like `site` parameter here is designed to accept site IDs, not short names. So particularly here we're fine. But I can see your concerns

Maybe we need to looks for more of catch-all solution here. Like with `channel:entries` if you pass `entry_id="not -1"` or `entry_id="not 08976786535478734546"` etc it will return you list of all your entries. I think inernally it just converts this to query and executes. So you can pass literally anything after `not`

So I think it would be nice if the user could provide anything that does not exist after `not` and that would give him full list

robsonsobral:
That's the trick I'm using: `not none`. However, I think a clearer option should be used. Have you ever coded some language with `not unless`? Jesus, it's confusing.

I took a look at channel module. There isn't a standard:

- `CURRENT_USER`;
- `NOT_CURRENT_USER`
- `\W`
- `IS_EMPTY`
- `not IS_EMPTY`

This need to be standardized:

- `CURRENT_USER`;
- `NOT_CURRENT_USER`
- `IS_EMPTY`
- `NOT_IS_EMPTY`
- `CURRENT_SITE`
- `NOT_CURRENT_SITE`
- `ALL_SITES`

(The work never ends. 😞 )

---

I had an idea! What happens when you add something to the reserved words list? Does the update process check for it and shows a warning?

intoeetive:
What you suggest is forcing the people to use certain keyword (e.g. `none`). What I suggest is that they can use anything that can't be populated into list of site ID's. So we don't force them to use `not none`, but they can use `not something` or `not something_else` or `not anything` etc. etc. and every word would work

robsonsobral:
Oh, no! Let me try to explain better!

We already can use `not foobar` or whatever. I've been using it! My point is that it's awkward and confusing. And it doesn't prevent unnecessary JOIN's.

intoeetive:
I see your current code has `not none` hardcoded, so that confused me. That's rather deceiving ;)
What if you replace it with regex and that's it?

robsonsobral:
I'm sorry! I was sure you didn't read my code so carefully. Next time, I'll write `// placeholder` at the end of line! 😉 

I can write a RegEx, but... Would it be so better to have a global value, like... Hmmmmnnnn... `ALL_OF_THEM`? `NOT`? `ROSEBUD`? `BYPASS`? `IGNORE`? `SOMETHING WITH SPACES`?

Does somebody at PackedTide has a better idea?

ops-andy:
Personally, I don't think hacking the `site=` param is the best (as shown in this discussion). If we're going to address this, I think we should add a new param, something very clear like `search_all_sites="yes"`

robsonsobral:
I'm sorry, @ops-andy , but I've to disagree with you. To add a new parameter looks much more like a hack to me: a hack to another parameter. Plus, see how weird would be to use it:

```html
{exp:MODULE site="default_site" search_all_sites="yes"}
```

Better to just close the fixes and keep the discussion on the parameter going apart.

robsonsobral:
~~And... What about `BYPASS` as a default parameter for these cases? Is it safe? Would somebody have a channel already called `BYPASS`?!~~

Jesus! I've already suggested that!

robsonsobral:
Besides the blocking discussion here, there is a huge improvement on search performance on this PR.

I gonna remove my suggestion of a way to bypass the unnecessary `JOIN`. I'm sorry. I didn't expect to make such a mess. The performance of the search module is so bad that I tried to improve it the most I can in a simple PR.

intoeetive:
I think if you remove that block for the `site`, it would still be improvement, right?

Maybe just move the contradictive part into separate PR, then we could easier review/approve this one?

robsonsobral:
I moved the value to a protected variable with value of `null`, waiting for our decision on a way to bypass default values.

---

I consider this a bug, because the unnecessary [loop of queries on fields](https://github.com/ExpressionEngine/ExpressionEngine/pull/1410/commits/744db63ad400faab9bcb71cce1082694401ebd7d) can kick the database down, depending on how many fields the site has.

---

Just to clarify things, my idea is to have a common value which bypasses default values on parameters.

---

Do you want me to rebase this branch, **Mister** @intoeetive ? 😉

intoeetive:
@robsonsobral I have read through the rest of your code, and all looks good and solid to me. I think this would be huge performance boost for sites that use search.

So the only disagreement we have is your update to this row:
```
$site_ids = (ee()->TMPL->fetch_param('site')) ? ee()->TMPL->site_ids : array(ee()->config->item('site_id'));
```
What if we just look at it differently? All custom fields are either assigned to a site, or are shared through the sites (`site_id = 0`)
(I think shared fields might by actually currently being omited from search, which is not correct)
So generally, you would always want to use either current site fields + shared fields, or specified site field + shared fields.
I think we could keep the current row as it is, and add a second row after it, like
```
$site_ids = array_merge($site_ids, [0]);
```
What do you think?

robsonsobral:
This time, I disagree, @intoeetive .

The ternary you mentioned was already there. I didn't change it.

The custom fields are being get using the channel model, around line 369:

```php
$custom_fields = array();
foreach ($channels as $channel) {
    $channel_array[] = $channel->channel_id;
    $custom_fields = array_merge($custom_fields, $channel->getAllCustomFields()->asArray());
}
$this->custom_fields = array_chunk($custom_fields, 50);
```

Does it make sense?

---


(Now, I realize that the `array_chunk()` limiting the fields to 50 is wrong! The fields should be limited **after** we check for `$field->field_search`. We can do it at line 369 or count the iterations after line 377.)

intoeetive:
Yes, you're actually right there

So, is there anything that prevents us from leaving the code that sets `$site_id` the way it is now?

(I know I probably am just missing something that's obvious for you... 😐)

robsonsobral:
There's just one extra thing; for all of us to have a beer and decide on the bypass value, around version 6.34 or so. The line 34 will be wainting for it:

```php
private $param_bypass = null; // we need to define a standard here, something like `BYPASS PARAM`
```

There's a conditional around that ternary using this constant:

```php
if (ee()->TMPL->fetch_param('site') != $this->param_bypass) {
```

Soooooooooooooooo, there's nothing more to do! You're free to merge this PR, which will also close issue #1549.

robsonsobral:
(except if you want to write some tests for the search...)

intoeetive:
Ok. I think I'm good including it as it is :) And will certainly need to get some beer after we merge it :) Maybe at next bEErConf.

But I'd like also an extra approval from team member

intoeetive (approved):

robinsowell:
@robsonsobral @intoeetive 

Doing some testing  using the default advanced search form- just dropped the new mod.search.php into a 6.1.5 install.  MSM is installed and I do have 2 sites.

If I search for an exact match on member, I get a PHP error

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'exp_channels.site_id' in 'where clause':
SELECT DISTINCT(exp_channel_titles.entry_id), exp_channel_titles.channel_id FROM exp_channel_titles LEFT JOIN exp_channel_data ON exp_channel_titles.entry_id = exp_channel_data.entry_id WHERE exp_channels.site_id IS NOT NULL AND exp_channel_titles.entry_date < 1639508656 AND (exp_channel_titles.expiration_date = 0 OR exp_channel_titles.expiration_date > 1639508656) AND exp_channel_titles.status = 'open' AND exp_channel_titles.channel_id IN (1, 2, 3, 4, 5) AND (exp_channel_titles.author_id IN (1) )
```

Screen Shot 2021-12-14 at 2 13 24 PM

<strike>I won't swear it's a new issue, but it is an issue.</strike>

NOTE- retested with the live mod.search.php and this worked without error.

robinsowell:
Hit another one:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'exp_categories.cat_id' in 'where clause':
SELECT DISTINCT(exp_channel_titles.entry_id), exp_channel_titles.channel_id FROM exp_channel_titles LEFT JOIN exp_channels ON exp_channel_titles.channel_id = exp_channels.channel_id LEFT JOIN exp_channel_data ON exp_channel_titles.entry_id = exp_channel_data.entry_id LEFT JOIN exp_channel_data_field_9 ON exp_channel_data_field_9.entry_id = exp_channel_titles.entry_id LEFT JOIN exp_channel_data_field_6 ON exp_channel_data_field_6.entry_id = exp_channel_titles.entry_id LEFT JOIN exp_channel_data_field_21 ON exp_channel_data_field_21.entry_id = exp_channel_titles.entry_id WHERE exp_channels.site_id IS NOT NULL AND exp_channel_titles.entry_date < 1639509470 AND (exp_channel_titles.expiration_date = 0 OR exp_channel_titles.expiration_date > 1639509470) AND exp_channel_titles.status = 'open' AND exp_channel_titles.channel_id IN (2) AND ( exp_categories.cat_id = '2' ) AND ( (((exp_channel_titles.title LIKE '%dude%' OR exp_channel_titles.title LIKE '%booterpie%' )) ) OR (((exp_channel_data_field_9.field_id_9 LIKE '%dude%' OR exp_channel_data_field_9.field_id_9 LIKE '%booterpie%' )) ) OR (((exp_channel_data_field_6.field_id_6 LIKE '%dude%' OR exp_channel_data_field_6.field_id_6 LIKE '%booterpie%' )) ) OR (((exp_channel_data_field_21.field_id_21 LIKE '%dude%' OR exp_channel_data_field_21.field_id_21 LIKE '%booterpie%' )) ) )
```
Screen Shot 2021-12-14 at 2 19 31 PM

booterpie does exist in the content field in a blog entry in the 'Personal' category.

NOTE- retested with the live mod.search.php and this worked without error.

obfuscode (changes requested):
Per Robin's comments, I'm going to mark this as changes requested so it can't be merged yet.

obfuscode:
@robinsowell Just so I follow, you edited your comments - does this still need to be looked into or is this good to go?

robsonsobral:
I need to check them, @obfuscode . I'm sorry!

robsonsobral:
Fixed, @robinsowell !

robinsowell:
> @robinsowell Just so I follow, you edited your comments - does this still need to be looked into or is this good to go?

I need to test now that it's been fixed again.  I haven't yet, so no, it's not good to go yet.  I'll pull and test tomorrow.

robsonsobral:
I don't think @robinsowell will ever believe me, but I had a dream of this code running and there's a bug. I'm fixing it right now.

robsonsobral:
DONE!

robinsowell:
@robsonsobral I'm still getting the issue when I pick a category.  The member search is all good.

But if I search Titles and entries and have a category selected on the default advanced search form, I get an sql error
Screen Shot 2021-12-15 at 12 31 26 PM

Flipped back to the old mod.search.php and it worked.

robsonsobral:
@robinsowell , have you tested with more than one category, please?

robsonsobral:
FIXED

robinsowell:
Thumbs up from me- member and categories are looking good now.

robinsowell (approved):

TomJaeger:
Looks like there's a few tests failing on this one.... If we get those resolved, we'll aim to get this into the 6.2.0 release.

robsonsobral:
> Looks like there's a few tests failing on this one.... If we get those resolved, we'll aim to get this into the 6.2.0 release.

Maybe the code needs to be rebased. I have no idea why tests of consent module and user registration are failing.

PR #1424, opened 2021-10-13
Add ability to use class/method to ACT requests

stephengalbraith:
## Overview

This PR modifies existing code to add the ability for developers to access ACT requests using the Class/Method reference instead of making calls to fetch the action id. It also moves the conditional to check if the module is installed outside the main conditional (the current ordering would only check if the ACT was not numeric, which would ignore many calls.

Essentially what I've done is allowed calls to be made by: ?ACT=class/method

If this is accepted, I can also submit a PR to the docs talking about this change.

I am open to discussions or any other ideas on this as well.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


intoeetive:
@TomJaeger what are your thoughts on this one?

matthewjohns0n:
Chatting internally about this one, I think we all tend to agree that this is not something we're comfortable including in the ExpressionEngine core. We're afraid it may open a potential security hole. For a few reasons, we prefer the action IDs to be obfuscated rather than referenced by name. 

Leaving them to be referenced by ID makes it difficult for an attacker to figure out which action ID belongs to a function that they might want to use as an attack vector. 

This also makes it hard to determine what other add-ons might be installed. For example, an attacker might try to post to `eCommerce:ProcessPayment` or something along those lines, to figure out more information about an add-on running in the core. 

We do appreciate the contribution, but at this time we've decided not to include this PR.

Issue #1467, opened 2021-10-20
Member Roles (#) - Count of members made more accurate or descriptive


mcfarlandonline:
EE Version 6.1.4
PHP Version 7.3.30

Right now, the (#) displayed after a Role only reflects how many members have that role assigned as their Primary Role. 

This feels a little deceiving and dangerous, as I could be tempted to delete a Role, not realizing it is a secondary role assigned to a plethora of members. 

I would rather have the (#) reflect both Primary and Secondary role assignments, or have it broken out. For example: 
> Role Name (# primary, # additional)
or
> Role Name (# *star icon, #)

See example screenshots where the Staff Portal role reflects 0 members, but in reality, there are 5.

![Screen Shot 2021-10-20 at 11 04 22 AM](https://user-images.githubusercontent.com/20441129/138148280-073d83d4-78bc-4776-a42d-70fdab84e87c.jpeg)

![Screen Shot 2021-10-20 at 11 04 11 AM](https://user-images.githubusercontent.com/20441129/138148229-0ace8b20-f620-48f0-bae0-1099309f7b84.jpeg)

PaulBaileyAtChan:
A couple of additions to this.

- The misleading member count shown after each role is especially dangerous because there's no check or prompt for existing secondary users if a role is deleted. The user is asked to confirm the deletion, but there's no indication of whether there are existing secondary users who would lose that role. (Including a prompt like that with the deletion confirmation would be a very good thing.)
- Also, in EE<=5, the member count shown after each role acted as a link to display a list of those members, which was useful. In EE6, that seems to be broken. It does seem to work as a link, but just lists all users, not users with that role attached. This looks like a bug to me, because linking to all users from here has no practical use.

Honestly, there's a lot that could be tightened up here. Not totally sure the new role management interface was properly tested.

intoeetive:
@ajohnson what are your thoughts on this?

ajohnson:
@intoeetive @mcfarlandonline @PaulBaileyAtChan All great points!

It looks like a fix has since been implemented for the link to display a list of members. Good deal.

The quickest improvement to the count display would be to simply make the count reflect the _combined_ number of primary _and_ additional within the _single_ count number. Clicking on that count number links to a list of all members who have that role assigned as either primary or additional (the link seems to be working properly after the recent fix).

[...]

PR #1469, opened 2021-10-21
improve UX of "Login as" page

### LINKED ISSUE #1466: Glitchy password masking when using "Login as..."
EE Version 6.1.4
PHP Version 7.3.30

Steps to recreate: 
1. Login as a Super Admin
2. Edit a member's roles and provide password to Authenticate & Save
3. Visit Administration > "Login as..." page of any member profile
4. Type in password to Login as... and see that the password masking is broken

_Note: 
I'm guessing there's a cookie that is caching the Super Admin's password, since after editing once member's roles, authenticating & saving, I'm not required to type my password in again to perform the action again or on another member. I find this inconsistent presence of the password field a little more alarming than helpful, and wonder if it's causing the glitchy password masking issue._

Screen Shot 2021-10-20 at 10 33 27 AM

intoeetive:
where URL input could be confused with password input

fixes #1466

EECORE-1464

fix the faded URL to Role Members on Roles page
related #1467

### DIFF
--- a/system/ee/language/english/myaccount_lang.php
+++ b/system/ee/language/english/myaccount_lang.php
@@ -523,10 +523,6 @@ $lang = array(
 
     'upload_avatar' => 'Upload — .gif, .jpg, .png (max %skb)',
 
-    'url' => 'Personal website',
-
-    'url_desc' => 'URL location of your personal website',
-
     'user_ip_address' => 'IP Address',
 
     'username' => 'Username',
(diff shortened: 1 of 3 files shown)


system/ee/language/english/myaccount_lang.php
```diff
@@ -523,10 +523,6 @@
 
     'upload_avatar' => 'Upload — .gif, .jpg, .png (max %skb)',
 
-    'url' => 'Personal website',
-
-    'url_desc' => 'URL location of your personal website',
-
```
obfuscode, line 529: Is this a conflict with the `Profile/Login.php` `lang('url')` field? Or is this related to this PR?

intoeetive: Keeping these in place made the wrong string being output in profile/login (so yes, conflict)
I checked and could not find where it is used (including legacy profile pages and forum), I believe it has been removed from EE6

obfuscode (approved):

PR #1471, opened 2021-10-21
Feature/6.dev/sql in empty

### LINKED ISSUE #1460: All Members filter-by-role results in SQL error if selected role is empty/has no members assigned
EE Version 6.1.4
PHP Version 7.3.3

When logged in as a Super Admin, and viewing "All Members", I'm trying to sort by Role. 

If the Member Role has no members assigned to it, I get a SQLSTATE error, (screenshots attached).

Screen Shot 2021-10-19 at 1 35 53 PM

Screen Shot 2021-10-19 at 1 34 29 PM

intoeetive:
fix the error when searching by role in CP if no members have this role 

The general reason for the error was that ActiveRecord
did not treat empty arrays well when passing those to `IN` filter.
The fix is to use random negative integer instead,
which assures the resulting query returns no results.

fixes #1460

EECORE-1462
EECORE-1465

-----------------

If null is provided as value to IN filter, use WHERE IS NULL instead

EECORE-1465

### DIFF
--- a/system/ee/legacy/database/DB_active_rec.php
+++ b/system/ee/legacy/database/DB_active_rec.php
@@ -660,9 +660,7 @@ class CI_DB_active_record extends CI_DB_driver
             return;
         }
 
-        if (is_array($values) && empty($values)) {
-            $values = 0 - floor(rand(1, 100) * 10000);
-        }
+        $emptyArray = is_array($values) && empty($values);
 
         if (! is_array($values)) {
             $values = array($values);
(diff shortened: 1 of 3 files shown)


obfuscode (dismissed):
I've approved this but I don't have a setup to test this currently.

system/ee/legacy/database/DB_active_rec.php
```diff
@@ -660,6 +660,10 @@ public function _where_in($key = null, $values = null, $not = false, $boolean_op
             return;
         }
 
+        if (is_array($values) && empty($values)) {
+            $values = 0 - floor(rand(1, 100) * 10000);
```
matthewjohns0n: Why is `floor()` being used? The result of `rand(1, 100) * 10000` will always already be an integer.

intoeetive: You are right. I did not read the PHP docs well :)
I can make the change, just waiting to see if there are more comments

matthewjohns0n: Also, to achieve a random negative int you could also just do something like this `rand(-200000, -100000);`

TomJaeger (approved):

PR #1486, opened 2021-10-24
fix icon image upload issue

### LINKED ISSUE #1361: Icon files fail the image check
When trying to upload an icon file into an image upload directory, one receives the error
```
Cannot Upload File
File not allowed.
```

### Steps to reproduce
* Create an upload directory that can only take images
* Attempt to upload an icon file
* See the error

### I know exactly what needs to be done and can submit a PR.
* In `mimes.php` add the following two lines
```
    'image/vnd.microsoft.icon', //.ico
    'image/x-icon', // .ico
```
* In `MimeType.php`, in the `fileIsImage` function, bypass the `imagecreatefromstring` check for icon files, as they are [not supported](https://www.php.net/manual/en/function.imagecreatefromstring.php) by that function.

Again, I am ready to submit a PR if a project lead can confirm that this would be desired. If there is a contributing guide, just send me a link.

The big argument for needing this is really straightforward: being able to manage favicons via the control panel.

acalvino4:
## Overview

Add the other common mime for .ico files to the list of allowed mime types, and bypass the convert-to-image check for icons since the php function doesn't support them.

Resolves [#1361](https://github.com/ExpressionEngine/ExpressionEngine/issues/1361).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Library/Mime/MimeType.php
+++ b/system/ee/ExpressionEngine/Library/Mime/MimeType.php
@@ -199,7 +199,17 @@ class MimeType
 
         // If the reported mime-type is an icon, we won't do the next validation step because imagecreatefromstring does not support .ico files
         if ($mime === 'image/vnd.microsoft.icon' || $mime == 'image/x-icon') {
-            return true;
+            try {
+                $file = fopen($path, 'r');
+                $first = fread($file, 4);
+                fclose($file);
+                return $first === "\x00\x00\x01\x00";
+            } catch (\Exception $e) {
+                if (DEBUG) {
+                    show_error($e->getMessage());
+                }
+                return false;
+            }
         }
 
         // If the reported mime-type is an image we'll do an extra validation
--- a/system/ee/ExpressionEngine/Library/Mime/MimeType.php
+++ b/system/ee/ExpressionEngine/Library/Mime/MimeType.php
@@ -197,6 +197,11 @@ class MimeType
             return true;
         }
 
+        // If the reported mime-type is an icon, we won't do the next validation step because imagecreatefromstring does not support .ico files
+        if ($mime === 'image/vnd.microsoft.icon' || $mime == 'image/x-icon') {
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Library/Mime/MimeType.php
```diff
@@ -197,6 +197,11 @@ public function fileIsImage($path)
             return true;
         }
 
+        // If the reported mime-type is an icon, we won't do the next validation step because imagecreatefromstring does not support .ico files
+        if ($mime === 'image/vnd.microsoft.icon' || $mime == 'image/x-icon') {
+            return true;
+        }
+
```
intoeetive: Is there a way to ensure that the file is icon by checking its content? For SVG files we check whether those contain something that looks like svg markup, however for .ico files here we always return true based on mime - is there a way we can make it more secure?

acalvino4: According to [this](https://stackoverflow.com/a/3382348/2410203) stack overflow, you can check the first few bytes. I can look into that and test it out a little - just need to figure out how to compare file content against a byte string.

acalvino4: Validation added, and icon images successfully upload. I have a hunch that whatever detects the mime type does this validation already, since when I deleted those bytes from the beginning of the icon file, it didn't even reach my validation since it already wasn't recognized as an icon mime type.

intoeetive (approved):
Thank you for this contribution!

It looks good to me and we will likely include it into next release (or soon after).

TomJaeger:
@matthewjohns0n can you review and make sure we can't get a PHP file to upload as a .ico or something along those lines before we merge this one?

matthewjohns0n (approved):
I tested this locally and it seems like a good change to me.

PR #1489, opened 2021-10-25
fix the issue where using exact keyword search on exp:channel:entries…

### LINKED ISSUE #1434: Live Preview giving notices when using exact keyword search on `exp:channel:entries`
Notice
Undefined offset: 1
ee/ExpressionEngine/Addons/channel/mod.channel.php, line 2293

Severity: E_NOTICE
Notice
Undefined index:
ee/ExpressionEngine/Addons/channel/mod.channel.php, line 2295

Severity: E_NOTICE

EE 6.1.1

intoeetive:
… was causing PHP notices in Live Preview

fixes #1434

EECORE-1468

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
+++ b/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
@@ -663,7 +663,7 @@ class Channel
                 $fields_sql .= ee()->channel_model->field_search_sql($terms, $search_column_name, $site_id);
             } // foreach($sites as $site_id)
             if (! empty($fields_sql)) {
-                $sql .= 'AND (' . $fields_sql . ')';
+                $sql .= ' AND (' . $fields_sql . ')';
             }
         }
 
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/channel/mod.channel.php
```diff
@@ -663,7 +663,7 @@ private function _generate_field_search_sql($search_fields, $legacy_fields, $sit
                 $fields_sql .= ee()->channel_model->field_search_sql($terms, $search_column_name, $site_id);
             } // foreach($sites as $site_id)
             if (! empty($fields_sql)) {
-                $sql .= 'AND (' . $fields_sql . ')';
+                $sql .= ' AND (' . $fields_sql . ')';
```
TomJaeger, line 666: Space at start of AND here shouldn't be needed, as it's added in before this method is called... Looks like most of the SQL in mod.channel has space at end of appended string as it's building.

TomJaeger (changes requested):

system/ee/ExpressionEngine/Addons/channel/mod.channel.php
intoeetive: It is exactly the space that was causing the issue (together with the brackets). But if you have strong objections why it should not be here, please let me know

obfuscode: Because this is in a `foreach`, it does need a space before it. Having extra spaces shouldn't be an issue but it does need one either before or after. I'm going to resolve this as I'm fine with this fix.

obfuscode (approved):

PR #1505, opened 2021-10-28
Fix for subtitle on Auto Save tab

Caleydon:
The title and subtitle at Auto Save tab in the entry publish are the same - "Auto-saved versions". This fixed subtitle lang('autosaved_versions_desc') translation string.

ee-entry-auto-save-tab


intoeetive (dismissed):
@Caleydon good catch! Thank you for this contribution, really appreciate it.

Would you mind signing the CLA agreement linked in the comment above, then we can include this into next release

intoeetive (approved):

TomJaeger (approved):

TomJaeger (approved):

PR #1507, opened 2021-10-28
Resolved #1429, #2157 where parts of Grid field could be not fully visible

### LINKED ISSUE #1429: Grid Field within Fluid Field gets cut off.
When having a grid field within a Fluid Field, the editing pane get's cut off on smaller screens, as below.
image

This occurs at `/admin.php?/cp/publish/edit/entry/#`

The interface is somewhat responsive in that when it gets small enough, it does snap to a vertical arrangement that works well.

However, the in-between screen width is at times unusable since I can't edit the content on the far right. The easiest (and probably best) solution would be to give that portion of the editor a horizontal scroll bar, rather than it overflowing the parent container.

- [x] I am available, able, and willing to work on this fix if it is approved.

### LINKED ISSUE #2157: Color picker field panel, hidden inside of grid field
When using the color picker as a field in the grid field, the actual color picker panel is hidden by the grid.

**How To Reproduce**
1. Create a grid field
2. Add a color picker as a grid field
3. Create a new entry and add a row to your grid
4. See that the color picker is hidden by the grid field

Screenshot 2022-07-31 at 11 20 34

**Environment Details:**
 - Version: 6.3.4
 - PHP Version 8
 - MySQL Version [e.g. 5.7]
 - OS: Linux
 - Web Server:  nginx

**Possible Solution**
Obviously Im not aware of the full impact. But on the element: .table-responsive removing the css property: overflow-x: auto; allows the color picker panel to be fully visible.

Yulyaswan:
Added `min-width = 0`  for fieldset to prevent content from appearing outside the block
Added scroll inside the Fluid filed if content more then block width

fixed  #1429 
fixed #2157 
EECORE-1494
EECORE-2085


ops-andy:
@intoeetive please attach this to a milestone and update the corresponding Jira ticket. thanks!

intoeetive (dismissed):
Looks good to me

image

TomJaeger (dismissed):
Running into some issues with Relationship in grid, and grid taking over whole preview screen space.... Posted video in Jira.  Not 100% sure if it's all related or not.

Yulyaswan:
@TomJaeger I fixed preview mode (posted video in Jira task)
I think we need to add a new ticket to fix the `relationship` dropdown position. It's a different issue and not related to this one

intoeetive:
I don't think we can release this without addressing the issue with dropdown position for the relationship field. The position is fine before this change (even though the UI is good)
We probably should not fix one thing and break another one.
I'll mark this PR as draft and set milestone to 6.2.0
@Yulyaswan please create Jira ticket for the relationship dropdown if needed

Yulyaswan:
I changed position: relative to static for the `main fluid container`. It fixed the dropdown issue because this dropdown should be attached to the `body` for the correct construction of coordinates.
I checked all styles related to the `Fluid` filed and didn't find anything with `position: absolute` that can be broken

intoeetive:
It does not look right still:
image

Yulyaswan:
I changed the dropdown position for Relationship field from bottom-start to bottom-end
on my end now it looks like this - https://share.getcloudapp.com/OAuG2DGL
Please let me know what do you think about this way?

intoeetive:
@Yulyaswan this PR is still draft, should it be reopened?

Yulyaswan:
@intoeetive it's still a draftб not ready to reopen

intoeetive (dismissed):

intoeetive (commented):
@Yulyaswan should https://github.com/ExpressionEngine/ExpressionEngine/pull/1774 be closed in favor of this one?

Yulyaswan:
@intoeetive yes

intoeetive (commented):
@Yulyaswan I have found an issue with this - however it also exists in EE7, so maybe worth merging this one first and then work on fix (as we need the change in both versions)? Not sure, looking for your opinion.

https://packettide.atlassian.net/browse/EECORE-1494?focusedCommentId=16533

The color picker change works great

Yulyaswan:
@intoeetive I added code to fix this issue. Let's double-check here and if everything ok (how it should be now), I'll move these changes to EE7

intoeetive:
@Yulyaswan functionality-wise it's working as intended now, but it looks like the code that's executed on load and on resize is same, or very similar. I think it needs to be abstracted to a function

intoeetive (approved):
This has been tested before and the changes are already in EE7. Approving

PR #1508, opened 2021-10-28
Fix the model generation

matthewjohns0n:
EECORE-1483

This includes better examples of how to use the model in the model that is generated.


system/ee/ExpressionEngine/Service/Generator/stubs/addon_model.php
```diff
@@ -1 +1 @@
-        '{{class}}'    => '{{namespace}}\Models\{{class}}',
+        '{{class}}'    => 'Models\{{class}}',
```
intoeetive: I believe the common convention used by EE is `Model` folder. And this might be needed to change in other places as well

system/ee/ExpressionEngine/Service/Generator/stubs/model.php
```diff
@@ -7,17 +7,29 @@
...
 
     // Add your properties as protected variables here
     protected $id;
 
-    public function __construct()
+    // Example of a property getter
+    protected function get__name()
     {
-        // Make magic, my friend
+        return $this->first_name.' '.$this->last_name;
```
intoeetive: This is not following PSR-12 formatting (needs spaces)

intoeetive (changes requested):
I think we need some changes here - see the comments

system/ee/ExpressionEngine/Service/Generator/stubs/addon_model.php
matthewjohns0n: I've updated this to use the `Model` folder.

intoeetive (changes requested):
@matthewjohns0n this worked fine, however I spotted related issue which I think would be nice to address together with this one.

I used single quote in author's name (which is not uncommon with European names) and it was used as part of namespace for model, so all code was broken. I think we need to strip everything that's not letter from the namespace

EDIT:

I actually don't think this was done right
I created model named `Two` in was placed directly in `Model` folder. And the code looks like this:
```
namespace JustMeHere\One\Model;

use ExpressionEngine\Service\Model\Model;

class Two extends Model
```
I think this might cause issues as it's not clear what `extends Model` represents.
I believe the common convention would be to have the model file placed in folder with same name, e.g. `Model/Two`

matthewjohns0n:
@intoeetive I like the suggestion of filtering the namespace. I've added that to this PR.

I don't agree with your second comment though regarding this code:

```
namespace JustMeHere\One\Model;

use ExpressionEngine\Service\Model\Model;

class Two extends Model
```

I think it is clear what the code is doing. In this example, the fully qualified name of the class is `JustMeHere\One\Model\Two` and it is using (and extending) the class `ExpressionEngine\Service\Model\Model`. There will be no conflicts here, since we are clearly using that model. 

If we wanted to be more clear for people just glancing at the code, we could change the use line to something like this:

```
namespace JustMeHere\One\Model;

use ExpressionEngine\Service\Model\Model as EEModel;

class Two extends EEModel
```

I'm happy with either solution, although this is unnecessary from an OOP point of view.

intoeetive (approved):

matthewjohns0n:
I have been looking into the tests failing and I just can't explain it. I set up cypress locally and ran all the tests on this branch and it passed all the tests. The tests that fail have absolutely nothing to do with the code I changed, so I'm convinced this is an issue with cookie consents/ the tests surrounding them. 

@intoeetive I see that you did recently change these tests and cookie-related things here:
https://github.com/ExpressionEngine/ExpressionEngine/commit/10a7091519395eb757dffc9fed9ec3826be2ed5d

The tests failing on this branch seem directly related to this. I'm not sure where else/ how far to dig. I'll try merging anything recent into this branch and pushing to retrigger the tests, but beyond that, I'm not sure how else to fix this.

matthewjohns0n:
Looks like that worked and now the tests passed! Excited to see that, I was very confused at this.

obfuscode (approved):

PR #1521, opened 2021-11-01
moved only login button to the modal footer part, EECORE-1180

Yulyaswan:
@intoeetive sure.
this is Jira ticket https://packettide.atlassian.net/browse/EECORE-1180
the whole form was in the block `dialog__actions`  that is responsible for the footer. I moved this wrapper, to wrap only login button and added an extra class `.dialog-input-wrap`  for input, to save the previous view for the modal window and prevent global styles from being overwritten


intoeetive:
@Yulyaswan can you add some explanation of what this does; or add a reference to JIRA or GH issue

ops-andy:
@intoeetive please attach this to a milestone and update the corresponding Jira ticket. thanks!

ajohnson:
The resulting screen shot in the issue looks good to me!

TomJaeger (dismissed):

intoeetive (approved):

PR #1524, opened 2021-11-02
Revised main Entries screen title, EECORE-1058

Yulyaswan:
EECORE-1058
Removed main title block, moved the button to the white block `panel-heading `, made title statiс


system/ee/ExpressionEngine/Controller/Publish/Edit.php
```diff
@@ -224,7 +224,8 @@ function ($key) {
                 'href' => ee('CP/URL', 'publish/create/' . $channel_id)->compile(),
                 'filter_placeholder' => lang('filter_channels'),
                 'choices' => $channel_id ? null : $choices
-            ] : null
+            ] : null,
+            'class' => strtolower(lang('entry_manager'))
```
obfuscode: I think this just needs to be `entries` without the call to the lang file otherwise people with language translations would end up with a different class name.

obfuscode (changes requested):
See line comment

system/ee/ExpressionEngine/Controller/Publish/Edit.php
Yulyaswan: changed

obfuscode (changes requested):
Added comment on jira ticket.

intoeetive:
@Yulyaswan the tests are failing. I believe they need to be updated to reflect the changes you made

obfuscode (approved):

PR #1525, opened 2021-11-02
Modify behavior of "Entries" flyout menu, EECORE-1059

Yulyaswan:
EECORE-1059

Added conditional to show `View All` only if the channel has at least 1 item
Added conditional to show the "Entries" flyout menu if it is not empty


ajohnson:
Can't speak to the code (I'll let @intoeetive check that), but the end result looks great. Thanks!

intoeetive (approved):

obfuscode (approved):

PR #1529, opened 2021-11-03
add password security policies, EECORE-1431 EECORE-1506

intoeetive:
improve UI of Security Settings controller
validate dictionary file name and password min length
validate passwords against selected password security policy

EECORE-1431

Docs PR: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/304


TomJaeger (changes requested):
We should add some tests on this front... Specifically to confirm the different PW settings are enforced correctly... That they pass when they should, and that they fail when they should.

system/ee/ExpressionEngine/Controller/Settings/SecurityPrivacy.php
```diff
@@ -163,7 +163,15 @@ public function index()
                     'title' => 'require_secure_passwords',
                     'desc' => 'require_secure_passwords_desc',
                     'fields' => array(
-                        'require_secure_passwords' => array('type' => 'yes_no')
+                        'require_secure_passwords' => array(
+                            'type' => 'radio',
+                            'choices' => array(
+                                'n' => lang('password_security_none'),
```
TomJaeger: For readability I think we should change these from letters to words.

TomJaeger: Tagging @intoeetive

system/ee/ExpressionEngine/Controller/Settings/SecurityPrivacy.php
```diff
@@ -288,6 +310,40 @@ public function index()
...
+        return true;
+    }
+
+    /**
+     * Custom validator to make sure min password length matches the selected policy
+     **/
+    public function _validatePwLen($length)
+    {
+        $rules = [
+            'g' => 8,
```
TomJaeger: For readability I think we should change these from letters to words. @intoeetive

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordSecure.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace ExpressionEngine\Service\Validation\Rule;
+
+use ExpressionEngine\Service\Validation\ValidationRule;
+
+/**
+ * Password Validation Rule
+ */
+class PasswordSecure extends ValidationRule
```
TomJaeger: I'm wondering if the name of this class is misleading... with Password secure.. It's a bit more of password enforce or something along those lines..  I'm thinking we should rename this to be a bit clearer.

enforce password level or something along those liens.

As a validation rule, I feel like we're validating that a password is secure, based on the name.. but they could have the "no" security option chosen and it would pass... 

Any gut reaction on this one @matthewjohns0n or @intoeetive Any good naming ideas?

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordSecure.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    public function validate($key, $value)
+    {
+        switch (ee()->config->item('require_secure_passwords')) {
+            case 'n':
```
TomJaeger: I think we should switch from letters to names for better code clarity (especially across files etc)

system/ee/language/english/settings_lang.php
```diff
@@ -756,9 +760,17 @@
 
     'require_ip_for_posting_desc' => 'When enabled, members will be unable to post without a valid user agent and <abbr title="Internet Protocol">IP</abbr> address.',
 
-    'require_secure_passwords' => 'Require secure passwords?',
+    'require_secure_passwords' => 'Password security policy',
```
TomJaeger: Should use a different array key

system/ee/legacy/controllers/cp/login.php
```diff
@@ -795,6 +799,31 @@ private function _return_to_login($lang_key)
 
         $this->functions->redirect(BASE . AMP . $redirect);
     }
+
+    /**
+     * AJAX endpoint for password strength meter
+     *
+     * @return JSON
+     */
+    public function validate_password()
```
TomJaeger, line 1057: @intoeetive can we use the validation service here instead of having this function here?

system/ee/legacy/core/Config.php
```diff
@@ -547,7 +547,7 @@ public function divination($which)
             'password_lockout_interval',
             'require_ip_for_login',
             'require_ip_for_posting',
-            'require_secure_passwords',
+            'secure_passwords',
             'allow_dictionary_pw',
```
TomJaeger, line 551: @intoeetive if someone currently has require_secure_passwords turned on, are we migrating them to basic?

TomJaeger (changes requested):
This is looking super solid!  Cleans up a lot of the old strong password checks etc.  Testing in the CP, install, etc we're all great.

Requested a few changes

system/ee/ExpressionEngine/Controller/Settings/SecurityPrivacy.php
intoeetive: We have labels for the readability. The options are letters for historical reasons - in previous versions the possible values for this setting were `y` and `n`, so we're adding two more here following the same pattern

system/ee/ExpressionEngine/Controller/Settings/SecurityPrivacy.php
intoeetive: See the comment above

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordSecure.php
intoeetive: I have no preference here (but also no other ideas).

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordSecure.php
intoeetive: The reason for using letters is described in comment above

system/ee/legacy/controllers/cp/login.php
intoeetive: This is the controller function providing the AJAX endpoint to show password strength meter. It is not possible to use validation service directly as a controller - and also the validation service can only return boolean value (valid or not valid), not the rank number/text

system/ee/legacy/core/Config.php
intoeetive: This change was made by mistake, reverted it. Thank you for bringing it to my attention

system/ee/language/english/settings_lang.php
intoeetive: What would be the reason for that? (we're not adding new setting here; we're upgrading the pre-existing one)

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordSecure.php
TomJaeger: I'd say lets go with enforce password level

system/ee/language/english/settings_lang.php
TomJaeger: Thought is that, this key used to line up with a setting called "Require Secure Passwords" Now it's really choosing a policy level.. so even though it's an upgrade, the settings has changed as a whole, no need to keep a legacy array key.

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordMatchesSecurityPolicy.php
```diff
@@ -0,0 +1,45 @@
...
+                break;
+            case 's':
+                return ee('Member')->calculatePasswordComplexity($value) >= 60;
+                break;
+            case 'y':
+            case 'g':
+            default:
+                return ee('Member')->calculatePasswordComplexity($value) >= 40;
+                break;
+        }
```
TomJaeger, line 35: Just to make sure.. the single characters match whats used elsewhere? @intoeetive

TomJaeger: I ask because in this file it looks like they have changed... https://github.com/ExpressionEngine/ExpressionEngine/pull/1529/files#diff-af1051c66cc328e3cf1b9d76141062401eace8e41651bd6fc2accde6bf17f30eR178

system/ee/ExpressionEngine/View/_templates/login.php
```diff
@@ -20,16 +20,20 @@
 		<?=ee()->view->script_tag('jquery/jquery.js')?>
 		<?=ee()->view->script_tag('common.min.js')?>
 		<?=ee()->view->script_tag('vendor/focus-visible.js')?>
-		<script>
-			var eyeOpen = '<?=URL_THEMES_GLOBAL_ASSET.'img/eye-open.svg'?>',
-				eyeClosed = '<?=URL_THEMES_GLOBAL_ASSET.'img/eye-closed.svg'?>'
-		</script>
-		<?=ee()->view->script_tag('cp/login.js')?>
+
```
TomJaeger, line 23: Just to confirm @intoeetive was the removal of the eye here on purpose?

intoeetive: Yes - it was moved to external JS file

system/ee/ExpressionEngine/Service/Validation/Rule/PasswordMatchesSecurityPolicy.php
intoeetive: This has been updated

TomJaeger (approved):

PR #1538, opened 2021-11-05
Pro Entries prolet: prefer preview based on Pages/channel settings

intoeetive:
for edits triggered from Entries prolet

EEPRO-295

Pro PR: https://github.com/EllisLab/ExpressionEngine-Pro/pull/130

### DIFF
--- a/system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
+++ b/system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
@@ -74,7 +74,7 @@ class LivePreview
     /**
      * generate and display the live preview
      */
-    public function preview($channel_id, $entry_id = null, $preview_url = null)
+    public function preview($channel_id, $entry_id = null, $preview_url = null, $prefer_system_preview = false)
     {
         if (empty($_POST)) {
             return;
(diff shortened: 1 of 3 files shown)


system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
```diff
@@ -137,20 +122,41 @@ public function preview($channel_id, $entry_id = null, $preview_url = null)
...
-            // We want to avoid replacing `{url_title}` with an empty string since that
-            // can cause the wrong thing to render (like 404s).
-            if (empty($entry->url_title)) {
-                $entry->url_title = $entry->entry_id;
-            }
+        }
 
-            $uri = str_replace(['{url_title}', '{entry_id}'], [$entry->url_title, $entry->entry_id], $channel->preview_url);
+        if (empty($preview_url) || $prefer_system_preview === true) {
+            if (! empty($_POST['pages__pages_uri'])
```
TomJaeger, line 128: @intoeetive This should be using the input class and not accessing the POST variable directly. ee()->input->post

system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
```diff
@@ -137,20 +122,41 @@ public function preview($channel_id, $entry_id = null, $preview_url = null)
...
+                //pages data passed with POST
+                $values = [
+                    'pages_uri' => $_POST['pages__pages_uri'],
+                    'pages_template_id' => $_POST['pages__pages_template_id'],
+                ];
+
+                $page_tab = new \Pages_tab();
+                $site_pages = $page_tab->prepareSitePagesData($entry, $values);
+
+                ee()->config->set_item('site_pages', $site_pages);
```
TomJaeger, line 139: I think this would be going a little contrary to what Live preview is.  I believe right now it's really just a preview of your changes before their saved... This is actually saving the URL for the page before it may be saved as a whole?  (Which could also result in a page going live in the navigation before it's been saved, as such it would 404) 
Thoughts? @intoeetive

TomJaeger (changes requested):

system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
intoeetive: This is pre-existing piece of code that has been moved to different position. I did not modify the actual code

system/ee/ExpressionEngine/Service/LivePreview/LivePreview.php
intoeetive: I think there's truth in what you're saying. But this code has always been there (I only moved it to different position). And I'm afraid there is currently no way to set the Page URL without actually setting it to config (the approach to Pages needs to be reworked really!)

TomJaeger: After reviewing this, and going back in git history to where it was added, I agree... This has actually been here for years... we will address when we revamp pages.

TomJaeger (approved):

PR #1542, opened 2021-11-06
Bring back default order field

### LINKED ISSUE #1258: Fallback order of fields to "old way"
The layouts of channels are powerful to use, but hard to customize.

I would like to be able to set the order of the fields on the fields themselves, like the old way. It don't, of course, need to use the old number fields, but a drag and drop within the groups. They would be the default value to be used in case of no layouts.

robsonsobral:
## Overview

Because to have to edit layouts one by one is too much work. It isn't a fancy drag and reorder thing, but is much better than to have to open layouts again, again and again for every channel.

Maybe the field description needs an update.

Resolves [#1258](https://github.com/ExpressionEngine/ExpressionEngine/issues/1258).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/changelogs/minor.md
+++ b/changelogs/minor.md
@@ -8,6 +8,7 @@ Bullet list below, e.g.
    - Added Google reCAPTCHA as an option for system wide CAPTCHA
    - Added ability to disable Live Preview on a per channel basis
    - Added extension hooks for Fluid Fields
+   - Added back default field order, because to have to edit layouts one by one is too much work
 
 
 EOF MARKER: This line helps prevent merge conflicts when things are
--- a/system/ee/ExpressionEngine/Controller/Fields/Fields.php
+++ b/system/ee/ExpressionEngine/Controller/Fields/Fields.php
@@ -541,6 +541,16 @@ class Fields extends AbstractFieldsController
                         )
                     )
                 ),
+                array(
+                    'title' => 'field_order',
+                    'desc' => 'global_sort_order',
+                    'fields' => array(
+                        'field_order' => array(
+                            'type' => 'text',
+                            'value' => $field->field_order,
+                        )
+                    )
+                ),
             ),
         );
 


intoeetive:
@ajohnson can you review this? 
It's adding back the functionality that was removed at some point - however I wonder if there has been specific reasoning on removing it. (Maybe intended to replace with drag&drop sorting? Which is not there - yet - however)

robsonsobral:
@ajohnson , to make your job easier:

image

ajohnson:
@robsonsobral @intoeetive It seems innocent enough to bring back the "Field Order" field to the New/Edit Field form. It was removed in EE 3 with the (then) new Channel Layouts taking its place.

But I'm honestly not sure if there are unintended consequences to bringing it back? Not sure why they removed it to begin with. Maybe some kind of nested logic edge case conflict they were trying to avoid with Channel Layouts?

Also if we bring back the Field Order will users expect to be able to view and sort their list of fields by field order like was available in EE 2?

Screen Shot 2021-11-10 at 2 34 53 PM

All this to say I'm fine with bringing back the "Field Order" field, but I don't trust myself alone to test this for potential errors/conflicts. We'd probably want to the team to jump into testing a bunch of random scenarios on this one since custom fields are such a high profile feature of EE.

@TomJaeger Do you recall any history here as to why the Field Order field was removed in the first place?

robsonsobral:
I don't see how it could be currently to sort the fields on CP. However, they should never be removed as they provide a starting point to the Channel Layouts (which should have names on their buttons).

One more thing! They were never removed from database or fields ordering. Just from the interface.

intoeetive:
We had discussion on this internally.

The "sort order" field was removed with introduction of Channel Layouts, which is intended to serve as replacement.

We have agreed that the Channels UI would benefit from some improvements, however we would rather want to address those in complex. Input for field order is a legacy thing and we don't want to reintroduce it.

robsonsobral:
I'm disappointed that the team which runs the project doesn't use it. Otherwise, it would know my reasons to bring this back aren't frivolous.😒

Issue #1472, opened 2021-10-21
UX improvements for Relationship field


intoeetive:
Please use this issue to add (and discuss) the ideas on how the UX of Relationship field can be improved

aasun:
I'm working on an extremely large and complex information/research site.  There are hundreds of thousands of entries with multiple relationship fields.  When more than a few relationship fields exist in a single entry, performance in the CP is greatly impacted.  So, some thoughts on how I'd like relationship fields to be able to be configured....

With tens of thousands of entries that can be related, it's not practical to set the view limit to all entries (it takes too long for the the entry edit screen to load).  Setting the limit to a small number (5 or 10) provides dramatic improvements for the entry edit page to load.  And the AJAX search in the relationship field will find matching entries from the tens of thousands, not just what is initially visible.  Great!  BUT, the  number of matching results that are displayed are limited by the display limit of the dropdown. Not great.  With tens of thousands of entries, there may be hundreds of entries that match a search.  As you continue to type your search term/phrase, the number of matches reduce. Great!  But, as a content editor, you won't know if your seeing all of the possible search matches until you see fewer results than what the view limit is on the dropdown.  Not great!

So, our current options are to increase the view limit of the relationship dropdowns to display more than what we think the search results may show.. In a site with tens of thousands of entries per relationship field, that takes us to showing hundreds in the dropdown. Which, completely bogs down the load time of the entry edit form. (with 10 or more relationship fields, it's nearly unusable).  Or, set the relationship field display limit to 5 or 10 and tell the content editors to open up the channel listing of the relationships to "pre-search" so they know exactly the title of the entry they want to relate so they have a very long search term they can type in the relationship field to find the entry to relate.  Which, is a completely impractical workflow.

So, Ideally, I'd like to be able to set the relationship drop down (initial) display limit to 0. (there is no practical reason to show any, when the relationship field pulls from thousands of entries..  Browsing a dropdown will never be the workflow.). Instead, rely fully on the AJAX search of the relationship field, and show ALL results that match the search as it's being typed.  (could probably place a max search result display to 99, as scrolling through 99 titles is as impractical as scrolling through thousands).  That way, there is no load/display/performance impact on the entry edit form from relationship fields, and the content editor then searches for all relationships.

[...]

PR #1545, opened 2021-11-08
Added MSM site info to Relationship Fields

robsonsobral:
## Overview

Add MSM necessary info to relationship fields.

Addresses [#1472](https://github.com/ExpressionEngine/ExpressionEngine/issues/1472).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/changelogs/minor.md
+++ b/changelogs/minor.md
@@ -8,7 +8,7 @@ Bullet list below, e.g.
    - Added Google reCAPTCHA as an option for system wide CAPTCHA
    - Added ability to disable Live Preview on a per channel basis
    - Added extension hooks for Fluid Fields
-
+   - Added MSM site info to Relationship Fields
 
 EOF MARKER: This line helps prevent merge conflicts when things are
 added on the bottoms of lists
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -523,7 +523,7 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
         if (ee()->config->item('multiple_sites_enabled') === 'y') {
             foreach ($channels as $channel) {
                 $channel_choices[] = [
-                    'title' => $channel->channel_title . ' <small>(' . $sites_by_id[$channel->site_id] . ')</small>',
+                    'title' => $channel->channel_title . ' (' . $sites_by_id[$channel->site_id] . ')',
                     'id' => $channel->getId(),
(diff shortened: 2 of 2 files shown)


robsonsobral:
To make @ajohnson work easier:

image

The channel dropdown deserves more love:

image

robsonsobral:
Any comments, @ajohnson ?

ajohnson:
@robsonsobral Just looking at your screen shots, and it seems like you have the site names rendering in parenthesis after the channel names. But I think we'd want the styling and/or placement of the site names to be unique so they can't be confused as being part of the channel names.

You can see how we did something similar here to avoid confusion: https://github.com/ExpressionEngine/ExpressionEngine/pull/1103

Was going to play with the styling, so I pulled this branch and have it running locally with a few MSM sites, but I'm not seeing the site names show up in the Relationship dropdown UI like your screen shots. Is there a config I need to tweak in order to see the site names show up?

A different thought...since we're starting to cram a lot of information (i.e. entry title, entry ID, channel name, site name) into the Relationship dropdown UI I'm wondering if it makes more sense to simply add a "Site" filter instead of trying to display the site name with every entry:
Screen Shot 2022-01-18 at 4 11 39 PM
This would allow you to restrict your entry searches by site. And the "Site" filter would only display when there are multiple sites in use.

Would this address the original problem? Or is displaying the site name with each entry a requirement for your use case?

Thanks!

robsonsobral:
I loved the idea of the site filter! Is there a component I could use there? Even better with an autocomplete in case of more than X sites. Being X whatever you think is right. I've almost 60 right now.

---

I don't remember of any required config. I did this one vacation ago. So, some info is lost among my memories of beach and good times. I gonna take a look.

robsonsobral:
I'm sorry it to so long to write back to you, @ajohnson .

I'm having a terrible hard time to build the styles and scripts. The dependencies hate me.

Anyway... I loved the idea of the select.

ajohnson:
@robsonsobral I'm seeing what the core team can do with this and maybe we can get it into our normal release schedule. No promises, but we'll see how it goes!

robsonsobral:
@ajohnson , hi!

Did you add site info to the relationship field?

Thank you!

TomJaeger:
was going through older EE 6 PR's and saw this... It's a good idea, so instead of closing I changed the base of the PR to 7.dev :)

PR #1546, opened 2021-11-09
prioritized rte styles over CP styles, EECORE-1489

### LINKED ISSUE #1485: RTE Redactor styles issue
When using the RTE with the new Redactor editor, the styles that are applied when viewing content in the control panel uses the styles of the control panel. For example a H2 is formatted using all capitals and an underline like within the control panel
Screenshot 2021-10-23 at 16 26 43
.

Yulyaswan:
Added parent class to the H2, to make styles prioritized.

fixes #1485 
EECORE-1489


intoeetive:
@Yulyaswan would it be worth to make this kind of update for other tags as well, not just H2?

Yulyaswan:
@intoeetive I have checked other headers, they work correctly

intoeetive (approved):

TomJaeger:
I believe this CSS needs to be moved out of the Redactor style sheet and into the main EE one, so when we update redactor we don't lose our changes.

PR #1547, opened 2021-11-09
Add setting for turning off the CLI

matthewjohns0n:
## Overview

Adds a setting for disabling the CLI.

EECORE-1388

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


TomJaeger:
@matthewjohns0n It looks like we have some failing tests on this one.  Any chance you can look into those?

matthewjohns0n:
@TomJaeger I pulled the failed archives, but it doesn't look to be related to these changes at all. Maybe @intoeetive has more insight into this?

I thought maybe it had something to do with the tests, so I merged the latest `6.dev` branch and pushed that up, since it would trigger the tests. Unfortunately, the tests only run when the PR is into `6.dev` so no tests ran. I think to trigger the tests again we need the PR to change the PR to 6.dev.

intoeetive:
@matthewjohns0n I've updated the workflow file to run tests on this branch as well

intoeetive:
So the installer tests failed, because there is mismatch on the version - the updater file for 6.2.0 is present in change, however the version in all files is still kept as 6.1.6. This is not correct, if the version is being updated, it should be changed in multiple places.

Maybe for this particular case we can ignore this however, as we know these changes already exist in other branches (and the rest of the tests succeeded)

TomJaeger:
Re-ran the tests... still failing.

matthewjohns0n:
I merged the latest `6.2.0` branch and now the tests are passing.

obfuscode (approved):

PR #1564, opened 2021-11-17
mobile view for Grid, EECORE-1499

Yulyaswan:
EECORE-1499


cp-styles/app/styles/components/_grid-field.scss
```diff
@@ -17,7 +17,40 @@
...
+  		@include m-tablet-down {
+  			min-width: 0 !important;
+  			width: 100% !important;
+  			max-width: 100% !important;
+  		}
+
+  		&.grid-rte,
+  		&.grid-file-upload,
+  		&.grid-multi-relate,
+  		&[data-fieldtype="wygwam"] {
```
intoeetive: While we own `wygwam`, technically it's third-party add-on, so if possible we should not reference it in EE stylesheets and use something more generic instead. If it's something specific for Wygwam, it should be added to Wygwam own styles instead.
Also, does this code mean we're limiting max width to 600 px even if the Grid has only one column?

Yulyaswan: no, as you can see on the screenshot it takes full space
Edit Entry Sub Page Two ExpressionEngine 2022-01-10 at 1 20 52 PM

Edit Entry Sub Page Two ExpressionEngine 2022-01-10 at 1 23 16 PM

intoeetive (approved):

PR #1568, opened 2021-11-19
fixed button text after search input loses focus, EECORE-1510

Yulyaswan:
fixed issue if the search input is empty and loses focus, the dropdown buttons become `undefined`


intoeetive (approved):
Tested, works well

PR #1574, opened 2021-11-22
fixed "URI too long" issue, EECORE-1493

### LINKED ISSUE #1504: "URI too long" error when using Relationships field
When using the "keyword" search in the relationshipsfield in the cp, you get an "URI too long" error from Apache, when relating a large channel with long titles.

This is because some settings (as like to current selection) are stored in the request URI (GET) but also in the POST.
Quick fix: increase: increase LimitRequestLine config in Apache

Might be a better fix to drop the current selection from URI and only use the POST values.

Yulyaswan:
Changed `get` ajax to `post` for relationship field, added conditional to check is it get or post a request

fixes #1504
EECORE-1493


intoeetive (approved):
Tested, works well

TomJaeger (approved):

PR #1576, opened 2021-11-23
define `PATH_PRO_ADDONS` on installer

intoeetive:


system/ee/ExpressionEngine/Core/Installer.php
```diff
@@ -23,6 +23,7 @@ public function boot()
         define('APPPATH', SYSPATH . 'ee/installer/');
         define('EE_APPPATH', BASEPATH);
 
+        define('PATH_PRO_ADDONS', SYSPATH . 'ee/ExpressionEngine/Addons/pro/levelups/');
```
TomJaeger, line 26: @intoeetive Are we still using the levelups directory for this?  Asking as I don't see this path in my local install (not 100% sure if it's part of the install process or something along those lines though).

TomJaeger (commented):
Awaiting feedback.

system/ee/ExpressionEngine/Core/Installer.php
intoeetive: We don't use it, but the constant is still defined in other places - and was missing here

TomJaeger (approved):

PR #1577, opened 2021-11-24
entry cloning

intoeetive:
EEPRO-311
Pro code PR: https://github.com/EllisLab/ExpressionEngine-Pro/pull/132
https://github.com/EllisLab/ExpressionEngine-Pro/tree/feature/entry-cloning
Core code PR: https://github.com/ExpressionEngine/ExpressionEngine/pull/1577

Docs PR: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/329


system/ee/ExpressionEngine/Addons/pages/tab.pages.php
```diff
@@ -294,6 +333,8 @@ public function prepareSitePagesData($entry, $values)
 
                 $page = '/' . trim($page, '/');
 
+                //if we got here, and page URI is not unique, generate new one
+
```
TomJaeger: @intoeetive I see this comment added... should there be additional code here?

system/ee/ExpressionEngine/Addons/pages/tab.pages.php
```diff
@@ -118,6 +120,37 @@ public function display($channel_id, $entry_id = '')
...
+        //exclude current page from check
+        if (isset($uris[$entry->entry_id])) {
+            unset($uris[$entry->entry_id]);
+        }
+        //ensure leading slash is present
+        $value = '/' . trim($values['pages_uri'], '/');
+
+        while (in_array($value, $uris)) {
+            $value .= '_1';
+        }
```
TomJaeger, line 148: I'm thinking we should check to confirm this URL isn't in the array already, and increment the number if it is... if people use the same base entry for cloning, we would then have duplicate urls ending in _1.
Tagging @intoeetive

system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
```diff
@@ -372,7 +372,17 @@ protected function validateEntry(ChannelEntry $entry, $layout)
             }
         }
 
-        $entry->set($_POST);
+        if (defined('CLONING_MODE') && CLONING_MODE === true && $this->entryCloningEnabled($entry)) {
+            $entry->setId(null);
+            while (true !== $entry->validateUniqueUrlTitle('url_title', $_POST['url_title'], ['channel_id'], null)) {
+                $_POST['url_title'] .= '_1';
```
TomJaeger: I think we need to check to confirm this is actually unique... Thinking people will likely use the same entry as a base multiple times.. tagging @intoeetive

system/ee/ExpressionEngine/Model/Content/ContentModel.php
```diff
@@ -503,6 +503,28 @@ protected function deleteFieldData()
...
+        } else {
+            foreach ($this->getCustomFieldNames() as $field) {
+                $this->setBackup($field, $this->getRawProperty($field));
+            }
+        }
+
+        parent::markAsDirty($name);
+
+        return $this;
+    }
```
TomJaeger, line 527: This looks like substantially the same code that was added in system/ee/ExpressionEngine/Library/Data/Entity.php in this PR. Can we reuse that? tagging @intoeetive

system/ee/installer/updates/ud_3_01_00.php
```diff
@@ -379,6 +379,7 @@ private function synchronize_layouts()
 
         $layouts = ee('Model')->get('ChannelLayout')
             ->with('Channel')
+            ->fields('field_layout', 'Channel.field_group')
             ->all();
```
TomJaeger, line 383: Is this change unrelated to cloning?

system/ee/language/english/channel_lang.php
```diff
@@ -302,6 +302,10 @@
 
     'enable_sticky_desc' => 'When enabled, entries can be marked as sticky and have preference in sorted listings.',
 
+    'enable_entry_cloning' => 'Enable entry cloning?',
+
+    'enable_entry_cloning_desc' => 'When enabled, entries can be duplicated using "Save as New Entry" option in saving options dropdown.',
+
```
TomJaeger, line 308: I think we changed this to "Clone to New Entry"

system/ee/legacy/libraries/api/Api_channel_fields.php
```diff
@@ -682,7 +682,7 @@ public function get_module_fields($channel_id, $entry_id = '')
     {
         $tab_modules = $this->get_modules();
 
-        $set = false;
+        $set = [];
 
         if ($tab_modules == false) {
```
TomJaeger, line 687: What is this change for?

system/ee/ExpressionEngine/Addons/pages/tab.pages.php
intoeetive: Parsing the string to get number would be more resource consuming then appending a string, which we do here. We might as well use `_copy` or `_clone` instead of `_1`

system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
intoeetive: Notice we have this in `while` loop. It is assuring the url_title to be unique

system/ee/installer/updates/ud_3_01_00.php
intoeetive: I am not 100% sure at this point. It is related in general to the line of changes we've agreed on where we only fetch the data that we need during the update process

system/ee/ExpressionEngine/Model/Content/ContentModel.php
intoeetive: It is not exactly same. This code is also marking entry's custom fields as "dirty" which parent function (also called here) can't do

system/ee/ExpressionEngine/Addons/pages/tab.pages.php
intoeetive: Removed this comment, it's not relevant anymore - the necessary code is in `clone` method

system/ee/installer/updates/ud_3_05_05.php
```diff
@@ -49,7 +49,7 @@ public function do_update()
      */
     private function normalizeFieldLayoutData()
     {
-        $layouts = ee('Model')->get('ChannelLayout')->all();
+        $layouts = ee('Model')->get('ChannelLayout')->fields('field_layout')->all();
```
TomJaeger, line 52: I assume this is also unrelated to this change? @intoeetive

TomJaeger (commented):
Left a few comments

system/ee/legacy/libraries/api/Api_channel_fields.php
intoeetive: In PHP 8.1, you can't convert `bool` to `array` by trying to add array elements to it

system/ee/installer/updates/ud_3_05_05.php
intoeetive: Correct

system/ee/language/english/channel_lang.php
intoeetive: Updated

system/ee/ExpressionEngine/Controller/Channels/Channels.php
```diff
@@ -1255,6 +1255,19 @@ private function renderSettingsTab($channel, $errors)
             )
         );
 
+        if (IS_PRO && ee('pro:Access')->hasValidLicense() && (ee()->config->item('enable_entry_cloning') === false || ee()->config->item('enable_entry_cloning') === 'y')) {
```
matthewjohns0n, line 1258: In this section:

`ee()->config->item('enable_entry_cloning') === false`

I believe the idea is that we're checking that the config item is not set. Is there any chance that it would return `false` if it were set to `n` or something like that? I don't think the config helper would do that, but this did seem like an odd check until I realized what it probably was for. 

At the very least, a comment would help with understanding the logic a bit.

intoeetive: `=== false` is specifically checking if the value is not set at all

intoeetive: There is no way for `n` to convert to `false` here

matthewjohns0n: Okay, that's what I figured, but I wanted to make sure.

TomJaeger (approved):

PR #1587, opened 2021-11-26
Issue #1586

### LINKED ISSUE #1586: Filesystem->findAndReplace() method bug
The findAndReplace method in [Filesystem.php](https://github.com/ExpressionEngine/ExpressionEngine/blob/6.dev/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php) has a bug that prevents its operation.

The method opens with a conditional that checks (presumably) to see if the file identifier provided exists as a file or directory, and exiting the path does not exist.

The conditional for this is missing a 'not' operator, and so the method exits when a valid path is provided before any operation is carried out.

jcogs-design:
## Overview

Correct conditional to address issue #1586

Resolves [#1586](https://github.com/ExpressionEngine/ExpressionEngine/issues/1586).

## Nature of This Change

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No


intoeetive (dismissed):
Good catch, thanks!

TomJaeger (dismissed):

TomJaeger (approved):

PR #1599, opened 2021-12-01
add counter for roles

### LINKED ISSUE #1467: Member Roles (#) - Count of members made more accurate or descriptive

### LINKED ISSUE #1533: Can't access Communicate and Roles on sites with ~ 50k members
Summary: 
When an ee site gets too many members, an inefficient query in system/ee/ExpressionEngine/Controller/Utilities/Communicate.php prevents admins from accessing the ee utilities in the control panel. Clicking on tools > utilities in the control panel loads system/ee/ExpressionEngine/Controller/Utilities/Utilities.php which in turn automatically loads Communicate.php. 

admin.php?/cp/utilities/communicate 
This page isn’t working domain is currently unable to handle this request.
HTTP ERROR 500
Can't reproduce on staging ( Our staging database doesn't have member data )

The snippet of code which is causing the issue is in the file 
```
system/ee/ExpressionEngine/Controller/Utilities/Communicate.php
```
on line 87

```
if ($role->getAllMembers()->count() == 0) {
     $disabled_roles[] = $role->role_id;
}
```
This is a foreach loop which iterates through each member role and checks to see if there are any members in that group. 

When $role->getAllMembers() gets called groups with large numbers of member 10k - 50k (the exact number changes with your infrastructure), you are not able to load the page. 

I have this conditional commented out for now on my site. Let me know if I diagnosed this issue incorrectly. 

Possible Solution: 

I could see this being replaced with a more verbose method such as
```
$role->hasMemebers()
```
which would hopefully be more efficient than returning a collection of all members.

### LINKED ISSUE #1582: Author listing for default Channel Form author does not account for > 100 authors
When an author is selected in a channel's "default author" field of it's Channel Form settings, and that author is past the 100 default limit of the Member service's `getAuthors() `function (e.g., found by searching for the author), subsequent loads of that channel's settings screen will not reflect the actually chosen author. Instead it will default to selecting the first author in the list.

intoeetive:
docs PR: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/333

partially solves the issues with some CP parts not loading when too many members are present

add `total_members` property for Roles
add `ignore_member_stats` config override

fix display of selected author when a lot of authors exist in system
fixes #1582
EECORE-1548

fix issue with Communicate page not loading with many members being present
EECORE-1513
fixes #1533

display correct number of members for each role
EECORE-1516
fixes #1467


system/ee/ExpressionEngine/Model/Member/Member.php
```diff
@@ -481,6 +503,13 @@ public function onBeforeDelete()
 
     public static function onAfterBulkDelete()
     {
+        if (ee()->config->item('ignore_member_stats') != 'y') {
+            foreach (ee('Model')->get('Role')->all() as $role) {
+                $role->total_members = null;
+                $role->save();
```
TomJaeger: @intoeetive does this already basically run from the onAfterDelete (assuming that gets run for each member?  Or does it get skipped in bulk delete?)

TomJaeger (commented):
Waiting for feedback before approving.

system/ee/ExpressionEngine/Model/Member/Member.php
intoeetive: Your are correct. Removed

TomJaeger (approved):

PR #1600, opened 2021-12-01
add PHP 8.1 support

intoeetive:
EECORE-1511


robinsowell (approved):

robinsowell:
Ran great for me logging errors and using it to replicate issues for support.

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_session.php
```diff
@@ -52,10 +52,10 @@ public function __construct($config)
     /**
      * userdata
      *
-     * @param	mixed $str
-     * @return	void
+     * @param mixed $str
+     * @return void
      */
-    public function userdata($str)
+    public function userdata($str, $default = false)
```
TomJaeger, line 58: @intoeetive I'm wondering why this $default value was added, but doesn't appear to be used?

system/ee/ExpressionEngine/Addons/filepicker/Service/FilePicker/Factory.php
```diff
@@ -53,26 +53,6 @@ public function make($dirs = 'all')
 
         return $fp;
     }
-
-    /**
-     * Handle a filepicker request. Does all the default stuff.
```
TomJaeger, line 58: @intoeetive I'm wondering why these were removed?

system/ee/ExpressionEngine/Controller/Files/AbstractFiles.php
```diff
@@ -227,13 +227,7 @@ protected function buildTable($files, $limit, $offset)
                 ),
             );
 
-            if (! ee('Permission')->can('edit_files')) {
-                unset($toolbar['view']);
```
TomJaeger, line 231: Didn't see this unset get reimplemented... does it need to be? @intoeetive

system/ee/legacy/libraries/Messages.php
```diff
@@ -1409,7 +1409,7 @@ public function do_member_search()
                         ->first();
 
                     if ($role) {
-                        $search_query[] = " member_id IN (" . implode($role->Members->pluck('member_id'), ',') . ")";
+                        $search_query[] = " member_id IN (" . implode(',', $role->Members->pluck('member_id')) . ")";
```
TomJaeger, line 1412: Just to confirm.. this one is more of a bug that PHP 8.1 support?

TomJaeger (commented):
Few questions opened on this one.  For now I'm going to move this to a 6.2.x release so we can finalize the 6.2 release

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_session.php
intoeetive: This is needed because the class
`class Channel_form_session extends EE_Session`
and the function definition needs to be compatible with definition in parent class, where `$default` exists

system/ee/ExpressionEngine/Addons/filepicker/Service/FilePicker/Factory.php
intoeetive: These were never called and also referenced non-existing class `Endpoint`

system/ee/ExpressionEngine/Controller/Files/AbstractFiles.php
intoeetive: The `$toolbar` array (created just above this code) does not have `view` key

system/ee/legacy/libraries/Messages.php
intoeetive: This **is** PHP 8.1 support fix. In older versions of PHP, `implode` could accept parameters in any order. PHP 8.1 is strict here

intoeetive:
All questions answered. Changing the milestone to 6.2 back again

system/ee/ExpressionEngine/Addons/filepicker/Service/FilePicker/Factory.php
TomJaeger: I guess my concern with this one is if a third party is using it?  (thinking handle request more then from request that's using a class that doesn't exist.)

Thoughts @intoeetive

intoeetive: If anyone was using it, it would not work - because it's referencing unexisting class. Basically it is broken code.

TomJaeger: I don't believe the handleRequest method was referencing a non-existing class?

FromRequest was though?

intoeetive: `handleRequest` is calling `$this->fromRequest()` which calls non-existing `Endpoint`

TomJaeger (approved):

PR #1603, opened 2021-12-01
Feature/1050/icon dates

### LINKED ISSUE #1050: Provide a method to quickly distinguish published entries from scheduled entries
Because of the way statuses interact with entry dates, looking at a list of entries means you need to scan the column of dates before you can tell which entries are open and published and which entries are open but unpublished.

One thing to do could be to bring Status and Date closer together in the table, making it clearer they interact in an important way. Another option would be to append "pending" to any dates in the future.

Screen Shot 2021-04-26 at 5 10 39 PM

It would also be great if the Date filter provided a dedicated option for seeing all entries with a date in the future, which I described in #513.

eric-lamb62:
## Overview

Adds 2 new columns to the Channel Entries list output that adds icon support for Entry and Expiration dates. This is based on the discussion here (#1050). Note that, as I have a terrible design sense, and the icon style was being discussed pretty heavily in the original issue, this ONLY outputs a div based on criteria. 

There are 3 options:
- col-date-closed
- col-date-future
- col-date-already-published

A UI dev should poke if able. 

Resolves [#1050](https://github.com/ExpressionEngine/ExpressionEngine/issues/1050).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


intoeetive (commented):
I like the feature, but I don't like the idea of having 2 columns for the same thing ("entry date" and "icon entry date")
I would prefer having a setting or config override instead, something like "use icons for dates in entry manager"
Tagging @ajohnson as he's the UX man here

eric-lamb62:
@intoeetive i see; makes sense. Will update ASAP using a config parameter to enable.

eric-lamb62:
@intoeetive made the changes. I think it hits all the notes but let me know if there's anything else on this. 

Went with a config value of `$config['entry_icon_dates'] = 'y';`  with enabling the feature defaulting to 'nope'

system/ee/ExpressionEngine/Library/CP/EntryManager/ColumnIconDate.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace ExpressionEngine\Library\CP\EntryManager;
+
+use ExpressionEngine\Library\CP\EntryManager\Columns\Column;
+
+abstract class ColumnIconDate extends Column
```
intoeetive: Maybe more generic `DateColumn` instead of `ColumnIconDate` would work better?

system/ee/language/english/content_lang.php
```diff
@@ -1013,6 +1013,10 @@
 
     'entry_manager_view_not_updated_desc' => 'We were unable to update this view, please review and fix errors below.',
 
+    'icon_expiration_date' => 'Icon Expiration Date',
+
+    'column_icon_entry_date' => 'Icon Entry Date',
+
```
intoeetive: I assume this is not needed anymore?

intoeetive:
I've added couple of comment, plus need to add some spaces around dots to make the code PSR compatible.

Also, do I understand correctly that we need to add CSS for icons to make this actually work?

system/ee/ExpressionEngine/Library/CP/EntryManager/ColumnIconDate.php
eric-lamb62: Agreed. Will update.

system/ee/language/english/content_lang.php
eric-lamb62: No, you're right. Good find. Will remove.

eric-lamb62:
> Also, do I understand correctly that we need to add CSS for icons to make this actually work?

@intoeetive yes please. I don't any that skill set where things look "good". Did attach some generic class properties though they need the pretty.

system/ee/ExpressionEngine/Library/CP/EntryManager/ColumnIconDate.php
eric-lamb62: What do you think about ColumnDate? That way it'll be consistent with all the other objects in the directory (being prefixed with "Column")? It itched the other way...

ajohnson:
@intoeetive @mithra62 Trying to catch up with some of these issues. This icon approach could work, but after reading some of the other discussions on this topic, I think there are additional details we'll need to consider when mixing the meaning of these icons with the way statuses & dates work and avoiding potential confusion. Also want to make sure we're not painting ourselves into a corner with any future features we might have planned for the Entries listing screen.

Can you give me a bit of time to work out the details? Thanks!

eric-lamb62:
@ajohnson no worries; this'll be here when/if you need it :) take your time, I don't think there's a rush

intoeetive:
@mithra62 I tried to bring this PR up-to-date with the current codebase, but appears there have been to much changes since you've submitted it.

I am going to close it, pick the necessary lines of your code and open another one. Thank you!

PR #1605, opened 2021-12-02
fix the issue when `{member_profile_link}` variable could be incorrect

### LINKED ISSUE #1604: {member profile_link} ignores "Member profile URL segment"
EE 6.1.6
PHP 7.3

I have a random hash of letters and numbers set for my site's Member profile URL segment (under System Settings > URL and Path Settings), but when I try using the Single Global Variable "{member_profile_link}" it defaults to using /member/ in that segment. 

So, for example, the link should be:
>` <a href="https://asitedomain.com/22asdkf239487asdkjh138a4b97jasd/2">Screen Name</a>`
But I get: 
> `<a href="https://asitedomain.com/member/2">Screen Name</a>`

Is there a different place to update the Profile URL segment, or is this a bug/legacy documentation that needs removing? 

bug?

intoeetive:
fixes #1604
EECORE-1557

Pro branch: https://github.com/EllisLab/ExpressionEngine-Pro/tree/feature/1.dev/cypress


system/ee/ExpressionEngine/Service/Template/Variables/StandardGlobals.php
```diff
@@ -111,7 +111,7 @@ private function getMemberProfileLink()
         if (ee()->session->userdata('member_id') != 0) {
             $name = (ee()->session->userdata['screen_name'] == '') ? ee()->session->userdata['username'] : ee()->session->userdata['screen_name'];
 
-            $path = "<a href='" . ee()->functions->create_url('/member/' . ee()->session->userdata('member_id')) . "'>" . $name . "</a>";
+            $path = "<a href='" . ee()->functions->create_url(ee()->config->item('profile_trigger') . '/' . ee()->session->userdata('member_id')) . "'>" . $name . "</a>";
```
TomJaeger, line 114: @intoeetive it looks like the starting slash got dropped, was this intentional?

intoeetive: It is redundant there

TomJaeger (approved):

PR #1613, opened 2021-12-05
Fix/1609/time limit cli

### LINKED ISSUE #1609: set_time_limit called on Command (CLI) Requests
When running ExpressionEngine through the Command routing, the a hard time limit of 90 seconds is imposed. This being a Command Line program, there should be no limits on execution time. 

This is on all platforms and systems as is the EE Core code. To get around this, the third party Commands have to set their own time limit which is counter to Command line as an environment. 

The offending code is here:
https://github.com/ExpressionEngine/ExpressionEngine/blob/05f4f24e81ec4b6590d2ffeab13fb361acad0f8f/system/ee/legacy/libraries/Core.php#L43

It should be a simple change to check for `php_sapi_name() === 'cli'` on that call.

eric-lamb62:
## Overview

Adds check for PHP on CLI before setting the execution time default.

Resolves [#1609](https://github.com/ExpressionEngine/ExpressionEngine/issues/1609).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


system/ee/legacy/libraries/Core.php
```diff
@@ -39,7 +39,7 @@ public function bootstrap()
         }
 
         // Set a liberal script execution time limit, making it shorter for front-end requests than CI's default
-        if (function_exists("set_time_limit") == true) {
+        if (function_exists("set_time_limit") == true && php_sapi_name() !== 'cli') {
```
TomJaeger, line 42: Just to confirm... I assume this then defaults to no limit?  (Any chance you ran a quick dump to confirm or anything along those lines?  If not I can run a quick test.)

Tagging @mithra62

eric-lamb62: Yes, no limit for CLI requests is what this does. I did test it to ensure the code for setting the memory limit doesn't execute on my local with the PR branch. I'm not following "a quick dump to confirm" though so you might wanna test just to be safe.

TomJaeger: Yeah, that's what I was wondering... basically just confirmation that it took effect as intended (time limit wasn't being set somewhere else or something along those lines) 

Right on!

eric-lamb62: You bring up a good point. I've been kicking around on whether to submit a ticket for this, but EE does set the execution time in numerous places across the code. 

image

Only one of those is engaged by default when used on Commands though abstracting that may help solve future Command based tasks.

intoeetive (dismissed):

TomJaeger (dismissed):

TomJaeger (approved):
Letting tests run.. but looks good to me.

PR #1614, opened 2021-12-05
Feature/1531/dynamic cli columns

### LINKED ISSUE #1531: CLI list cuts off longer names of commands

When creating a command with a longer name, it gets cut off in `eecli:list`

**How To Reproduce**

1. Create an addon with "Hey This Is a Great Module".
2. Create a command for that addon titled "Oh Wow What a Command" with a signature `do:amazing-things-all-the-time`
3. Run `php eecli.php list`

**Environment Details:**
 - Version: 6.1.5
 - PHP Version 8.0 and 7.3
 - MySQL Version: MariaDB 10.5.12
 - OS: Linux
 - Web Server: nginx

**Additional context**

Screenshot from 2021-11-03 10-48-09

Screenshot from 2021-11-03 10-48-14

eric-lamb62:
## Overview

Changes the Column output on the Command List to allow for dynamic widths (longer commands and/or Command descriptions)

Resolves [#1531](https://github.com/ExpressionEngine/ExpressionEngine/issues/1531).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


dougblackjr:
@mithra62 I think this gets closer, but still doesn't show the entirety of the command.

Screenshot from 2021-12-05 14-45-02

eric-lamb62:
That brings up an interesting question; should EE just use whatever's there without constraint or consideration for UI?

I ask, 'cause I added a sanity check on this so the UI/X wouldn't get munged on long names. Otherwise, a single command could cause the entire output to get screwy looking. 

Right now, it's locked to 40+/- on command column and 100+/- on the command description column. 

FWIW, wrapping lines doesn't look to be supported. 

Thoughts @dougblackjr?

dougblackjr:
@mithra62 I definitely think constraints make sense. My only concern would be how one would process long names.

Another framework handles this by showing the long name, and wrapping the text.

Screenshot from 2021-12-06 06-39-15

It _appears_ that in this framework, they are not obeying a table, simply setting up rows. Here is a small window displayed with that long command:

Screenshot from 2021-12-06 06-42-25


What might be good IMO would be to show the entire command name, and then limit the amount of characters in the description, or remove the table entirely.

eric-lamb62:
@dougblackjr guess I'm saying, EE could, in its authority as the platform, dictate the commands can't be longer than 40/50/100 characters then the UI becomes a non issue. I mean, EVERY platform has a limit so I see no problem with putting a max on it. 

Once we know the max THEN we can figure out the UI.

eric-lamb62:
So, to recap, currently, there is no limit on Command length. So Command strings gets cut off due to the existing UI being a hard limit on width of 20 characters. The PR I added allows for fluid width and adjusts the List Commands UI accordingly. 

This PR though, opens up the possibility for long commands, well over 40 characters (which is kinda rude IMHO to expect of users to input) to munge up the UI as a whole. The entire thing can now look turrible if a single dev decides to yolo. Which probably isn't good.  

So, the way I see it, the way around this is to put a max limit (which the PR includes). So we just have to agree upon what that limit is. 

Anyone else wanna chime in?

matthewjohns0n:
Thanks for hopping into this one! Overall, I think this is a good addition. I agree it would be best for us to take a stance on command length and enforce it since one add-on can spoil the whole UX. I'm not sure what the length should be, but I'll play around with it in the CLI and see what seems reasonable. Any suggestions on character length?

A few things we will also need to add for enforcing the max length:

- Document our chosen length in the [Docs](https://docs.expressionengine.com/latest/cli/creating-a-command.html)
- Have the list command cut off the command name after that number of characters if a developer doesn't follow the max length. We could even throw something in the developer logs, which might encourage them to change it
- Validate command name length in the command generator

eric-lamb62:
Any downside to putting the value for max command string length as a config value? That way it can be overridden and we can avoid hard coding it across the codebase.  Plus, dev on the 2 code based items could proceed while the exact number is determined. 

Speaking of, did you wanna do those as a new Issue/PR or keep attached here? I'm down to do the work FWIW if you guys wanna tag me in.

dougblackjr:
@mithra62 @matthewjohns0n Having a set character limit would be good when creating the command would be good. I think, too, about those who are manually creating commands that this may not impact, since we couldn't enforce what would be written in the code (i.e. existing add-on devs adding commands to their add-on).

I'd almost rather see us adjust the UI view for this in the CLI, rather than imposing a limit. I rarely think someone would ever hit a limit like the example I gave, but if it was needed for a user, I think it should be the UI should adjust and not the user.

intoeetive (changes requested):
@mithra62 This looks good (did not test though, just looked at the code) - but can you add extra spaces around operators to make the code PSR-12 compatible? (there's also `phpcs.ruleset.xml` file in the root of the project)

You may also want to merged `6.dev` back into your branch (we previously had an issue that prevented third-party PRs to pass the tests_

eric-lamb62:
> can you add extra spaces around operators to make the code PSR-12 compatible?

absolutely.

eric-lamb62:
@intoeetive there wasn't anything to merge from `6.dev` for me; am I missing something? I pulled but nothing there :shrug: 

Updated the code FWIW. Let me know if there's anything else though.

intoeetive:
You need to pull from `ExpressionEngine/ExpressionEngine`, not from your fork

eric-lamb62:
@intoeetive did not know that lol 

thanks! updated.

intoeetive (approved):

intoeetive (changes requested):
@mithra62 can you please add spaces around `.` and `+` to make the code PSR-12 compatible.

Otherwise - tested and it looks good to me.

eric-lamb62:
Found me a fancy schmancy psr12 format thing in PHPStorm; should have done the trick. Lemme know if there's anything else though.

intoeetive (approved):

matthewjohns0n (approved):

PR #1630, opened 2021-12-09
6.2.0

intoeetive:


ops-andy (approved):

.github/workflows/tests.yml
```diff
@@ -96,7 +87,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [5.6, 7.1, 7.2, 7.4, 8.0]
```
matthewjohns0n, line 99: Why are we removing testing around 5.6? We haven't dropped 5.6 support yet

PR #1637, opened 2021-12-09
Escape the key in search query

matthewjohns0n:
EECORE-1543

### DIFF
--- a/system/ee/EllisLab/Addons/member/mod.member_settings.php
+++ b/system/ee/EllisLab/Addons/member/mod.member_settings.php
@@ -1658,7 +1658,7 @@ class Member_settings extends Member {
 			{
 				if ($val != '')
 				{
-					$search_query[] = ee()->db->escape_like_str($key)." LIKE '%".ee()->db->escape_like_str($val)."%'";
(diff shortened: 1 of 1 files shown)


system/ee/EllisLab/Addons/member/mod.member_settings.php
```diff
@@ -1658,7 +1658,7 @@ function do_member_mini_search()
 			{
 				if ($val != '')
 				{
-					$search_query[] = $key." LIKE '%".ee()->db->escape_like_str($val)."%'";
+					$search_query[] = ee()->db->escape_like_str($key)." LIKE '%".ee()->db->escape_like_str($val)."%'";
```
intoeetive: Shouldn't this use `escape_str` instead?

matthewjohns0n: Yeah, good catch - I didn't realize there were two functions and let autocomplete fill in that function when I was typing it. I'm updating it and will resolve this conversation once it is pushed up.

intoeetive (approved):

obfuscode (approved):

PR #1640, opened 2021-12-09
Remove name from email templates

matthewjohns0n:
EECORE-1165


intoeetive (approved):
This does not make much sense to me particularly in context of EE5, since this is only affecting new installation. However, no objections.

obfuscode (approved):

PR #1658, opened 2021-12-14
Add support for field groups within a fluid field

bakin1999:
## Overview

php ini file can have short_open_tag=On or off.  Yulya and I were both getting same error in this and I think its because We have ours turned to off (I am not sure if she did as well but that would be my guess). I think we should use off incase people have it set to that when running.

Replace this paragraph with a description of your changes and reasoning. Provide links to external references/discussions if appropriate.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Addons/fluid_field/views/fieldgroup.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/views/fieldgroup.php
@@ -38,8 +38,8 @@
             </div>
         </div>
 
-        <? foreach ($field_group_fields as $field) : ?>
-            <? if($field->getType() !== 'fluid_field'): ?>
+        <?php foreach ($field_group_fields as $field) : ?>
+            <?php if($field->getType() !== 'fluid_field'): ?>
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/fluid_field/views/fieldgroup.php
```diff
@@ -57,10 +57,10 @@
 
                 echo $form;
                 ?>
-                <?= isset($errors) ? $errors->renderError($field_name) : '' ?>
+                <?php isset($errors) ? $errors->renderError($field_name) : '' ?>
```
intoeetive, line 60: This line does not output anything because it's missing `echo` or `=`

intoeetive (changes requested):
Some bit is missing; good otherwise

intoeetive:
As per @bryannielsen this is already done without PR

PR #1661, opened 2021-12-15
Bug/6.2.0/fix cli updater

matthewjohns0n:
EECORE-1536


intoeetive (approved):
I am approving this, but I also had to push 2 updates:

- one to disable loading session library for the updater - that does not seem to be needed and is causing SQL error
- and another related to fact that CLI was off by default because of wrong setting name

So probably need someone else to quickly test again

matthewjohns0n:
I tested @intoeetive's changes and they were good. I also came across another issue in the logic of the DB updater and fixed that.

PR #1668, opened 2021-12-17
Use original filename of uploaded files as fallback for title; #1660

### LINKED ISSUE #1660: Original file name as fallback to title field
When the **Title** field isn't filled, the filename is used as a default value, but the edited version, the one used to save the file on the system. It's weird to use an already edited value as fallback. There are users complaining about it and I think they have a point.

We could have a `use_orig_name_as_fallback_title` config override to change this behavior. I'm searching for where to implement it.

robsonsobral:
## Overview

~~Allow use of original filename as title fallback by the use of `use_orig_name_as_fallback_title` config override.~~

Resolves [#1660](https://github.com/ExpressionEngine/ExpressionEngine/issues/1660).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


robsonsobral:
Probably, I need to change the name of the config to something shorter, but I need suggestions.

system/ee/ExpressionEngine/Service/File/Upload.php
```diff
@@ -468,7 +468,12 @@ public function validateFile(FileModel $file)
         $action = ($file->isNew()) ? 'upload_filedata' : 'edit_file_metadata';
 
         $file->set($_POST);
-        $file->title = (ee()->input->post('title')) ?: $file->file_name;
+
+        if (ee()->config->item('use_orig_name_as_fallback_title') === 'y') {
+            $file->title = (ee()->input->post('title')) ?: $_FILES['file']['name'];
```
intoeetive: This code is assuming the file has been uploaded as field with name `file` - which is not always true. I think it needs to be more abstract to work in all cases.

intoeetive (commented):
I am not completely sure what this code does.

Does it revert the file title to original file name when edited and saved with empty title or something like that?

robsonsobral:
@intoeetive , my intention here is to fix something I found weird: the title fallback is the filename **already** converted.

Giving an example! Imagine you submit a file named "Yuri lives in a cold country.jxl", but don't fill the field _Title_. Both the filename and title will be set to "yuri_lives_in_a_cold_country", which clearly breaks the purpose of the fallback. The original name can even have some characters which have to be URL encoded.

intoeetive:
I see! This totally makes sense and needs to be fixed. I think we might even not need config override, as the current behaviour as you describe it does not seem to be correct

Just see the other note on code that I made

(and no, I don't live in a cold country - but it's certainly much colder than yours! =))

system/ee/ExpressionEngine/Service/File/Upload.php
robsonsobral: @intoeetive , I'm having a hard time to find my way around the code. Maybe the [`finishUpload()`](https://github.com/ExpressionEngine/ExpressionEngine/blob/96307e89a6702aecd70f88e6a258c7d5d7e78423/system/ee/ExpressionEngine/Controller/Files/Files.php#L240) or [`upload()`](https://github.com/ExpressionEngine/ExpressionEngine/blob/96307e89a6702aecd70f88e6a258c7d5d7e78423/system/ee/ExpressionEngine/Controller/Files/Files.php#L147) functions of the [**Files** controller](https://github.com/ExpressionEngine/ExpressionEngine/blob/6.dev/system/ee/ExpressionEngine/Controller/Files/Files.php)... I've no idea.

intoeetive: I had a look at the code and it seems this function is only being called when uploading the file via Files controller. Which does not seem right, but it is what it is. And it's already using some hard-coded POST variable names.

So it's probably fine to use `$_FILES['file']` here - just make sure it's defined before using

robsonsobral:
I also removed the file extension

robsonsobral:
I needed to update some tests because I removed the file extension from the fallback title.

robsonsobral:

- changes on tests;

intoeetive (approved):
Looks good to me

robsonsobral:
@intoeetive , is PR still needed? I guess it is, at least for version 6.

intoeetive:
I don't think it is relevant for EE7, as the title cannot be specified during upload there - but yes, we still intend to include it for EE 6.

PR #1669, opened 2021-12-17
Improve expired search link message

### LINKED ISSUE #1318: Expired search links could have a clearer message
When an expired search token is used on a URL, the feedback message just says it's an invalid action.

```
array(lang('invalid_action'))
```

We could have a better one, like `array(lang('expired_search_results'))` or something like that:

```
        // Fetch the cached search query
        $query = ee()->db->get_where('search', array('search_id' => $search_id));

        if ($query->num_rows() == 0 or $query->row('total_results') == 0) {
            // This should be impossible as we already know there are results
            return ee()->output->show_user_error(
                'general',
                array(lang('expired_search_results'))
            );
        }
```

robsonsobral:
## Overview

When an expired search token is used on a URL, the feedback message just says it's an invalid action. We can have a better message.

Resolves [#1318](https://github.com/ExpressionEngine/ExpressionEngine/issues/1318).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


system/ee/language/english/search_lang.php
```diff
@@ -152,6 +152,8 @@
 
     'today' => 'Today',
 
+    'expired_search_results' => 'Expired search. Try again.',
```
intoeetive: I think this needs better wording, but I'd prefer native speaker to come forward here. @matthewjohns0n or @ops-andy any ideas?

(we also need to merge latest `6.dev` to make the tests pass)

intoeetive (changes requested):
I think this needs better wording

system/ee/language/english/search_lang.php
robinsowell: I like 'expired search' better than 'invalid action' which is definitely not helpful.  Not sure about the 'try again'.

intoeetive: @robinsowell how about something similar to "The search results have expired. Please submit search form again." (although I'm not sure on part RE the form)

robsonsobral: I updated the message, but ignored the "submit form" part, because maybe there's no form involved.

intoeetive (dismissed):

system/ee/language/english/search_lang.php
```diff
@@ -152,6 +152,7 @@
 
     'today' => 'Today',
 
+    'expired_search_results' => 'The search have expired. Please, try again.',
```
intoeetive: @matthewjohns0n need your input here as native English speaker :). How can we word this better? Also maybe worth removing "please try again" part completely, as some people will just try to refresh the page, which will not help

bryannielsen: @intoeetive I made a few tweaks to the message

bryannielsen (approved):

PR #1671, opened 2021-12-17
More detailed instructions for running Cypress

litzinger:
Everything in this file "worked" for me the first time, except when I went to actually run the tests locally it kept failing on the db:seed task. It would be helpful if this file was more detailed so someone with no Cypress experience could setup and run the tests locally without error. 

Does it require a minimum version of Node/NPM? If so it should be mentioned in the README.

Does the user need to have local instance of MySQL server running on their machine? If so this should be specifically mentioned and not assumed.

Also fixed the path to the config.php file.


tests/cypress/readme.md
```diff
@@ -15,18 +15,19 @@ You can change the environment constants to match your local setup by copying th
 `cp tests/cypress/cypress.env.example.json tests/cypress/cypress.env.json`
 
 ## Settup Composer
-cd tests/cypress  and then run composer install
+
+`cd tests/cypress  and then run composer install`
```
robsonsobral: Shouldn't it be "`cd tests/cypress` and then run `composer install`"?

intoeetive (dismissed):

intoeetive (approved):

PR #1677, opened 2021-12-22
support PHP & EE version check

intoeetive:
EEPRO-280

Pro branch: https://github.com/EllisLab/ExpressionEngine-Pro/tree/feature/2FA

This adds support for 
```
    'requires'       => [
        'php'   => '7.1',
        'ee'    => '6.2.0'
    ],
```
in `addon.setup.php` so that add-on that does not meet the requirements cannot be installed

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Members/Roles/Roles.php
+++ b/system/ee/ExpressionEngine/Controller/Members/Roles/Roles.php
@@ -662,11 +662,24 @@ class Roles extends AbstractRolesController
                     'fields' => [
                         'require_mfa' => [
                             'type' => 'yes_no',
+                            'disabled' => version_compare(PHP_VERSION, 7.1, '<'),
                             'value' => $role->RoleSettings->filter('site_id', ee()->config->item('site_id'))->first()->require_mfa,
                         ]
                     ]
                 ],
             ]);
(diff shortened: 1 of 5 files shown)


system/ee/ExpressionEngine/Controller/Members/Roles/Roles.php
```diff
@@ -662,11 +662,24 @@ private function renderRoleTab(Role $role, $errors)
...
+                $section = array_merge($section, [
+                    ee('CP/Alert')->makeInline('mfa_not_available')
+                        ->asWarning()
+                        ->withTitle(lang('mfa_not_available'))
+                        ->addToBody(sprintf(lang('version_required'), 'PHP', 7.1))
+                        ->cannotClose()
+                        ->render()
+                        . form_hidden('require_mfa', 'n')
+                ]);
+            }
```
TomJaeger, line 682: Not loving the PHP version check and a specific functionality warning being generated in line here.

What if we abstract this a bit.  Could we make this part of the MFA service?  (Maybe both the addition of the setting as well as the check and inline warning) ?

Tagging @intoeetive and @bryannielsen for additional thoughts?

In part I'm thinking this is the first minor release of Pro... it's likely we'll run into this multiple times in the next few years.

TomJaeger (approved):

PR #1678, opened 2021-12-22
fixed block display for Value/Label Pairs option in checkbox inside Grid, EECORE-1583

Yulyaswan:
The problem was that when saving any other checkbox option besides `Value/Label Pairs`, there was no possibility to edit `Value/Label Pairs`
This only applies to `Checkbox`, `Multi checkbox`, and `Radio buttons` inside `Grid`


cp-styles/app/styles/components/_no-results.scss
```diff
@@ -44,3 +44,5 @@ table .field-no-results {
 	text-align: center;
 	padding: $sq-inset-s;
 }
+
+.field-no-results.hidden { display: none !important;}
```
intoeetive: can you rewrite here using SASS syntax? (i.e. nesting the selectors)

themes/ee/asset/javascript/src/cp/grid.js
```diff
@@ -259,6 +260,13 @@ Grid.Publish.prototype = Grid.MiniField.prototype = {
 		if (this.isFileGrid) {
 			this.root.toggleClass('hidden', rowCount == 0)
 		}
+
+		if(rowCount == 0) {
+			setTimeout(function(){
```
intoeetive: Is there a specific need for `setTimeout` here? (thinking because the code above does not use it)

intoeetive (changes requested):

themes/ee/asset/javascript/src/cp/grid.js
Yulyaswan: we need this `setTimeout`, since the element is being replaced and we need to catch and fix the new element

cp-styles/app/styles/components/_no-results.scss
Yulyaswan: rewrote code

themes/ee/asset/javascript/src/cp/grid.js
intoeetive: In such case, is `50` milliseconds sufficient? Or maybe we need to check for actual elements existance (in a loop with timeout)

Yulyaswan:
I think we need to leave this js code. `.field-no-results` is always present, but we need this 50ms to run the code after changing its status

themes/ee/asset/javascript/src/cp/grid.js
obfuscode: I agree with Yuri. 50 ms is 1/20th of a second. That seems to be such a small amount of time that I would think it would still fail that check if the page has a lot of other elements to load or if the user has a slower computer. I think this needs to be in a setInterval loop with a clearInterval when it's found.

Yulyaswan: ok, I changed setTimeout to setInterval with clearInterval

obfuscode (approved):

PR #1680, opened 2021-12-22
Fix bug with login page redirections

matthewjohns0n:
EECORE-1582

### DIFF
--- a/system/ee/legacy/controllers/cp/login.php
+++ b/system/ee/legacy/controllers/cp/login.php
@@ -26,6 +26,22 @@ class Login extends CP_Controller
         $this->lang->loadfile('login');
     }
 
+    /**
+     * Magic method for redirecting if someone attempts to reach a login method that does not exist
+     *
+     * @access  public
+     * @return  redirect
+     */
+    public function __call($name, $arguments)
+    {
+        // has their session Timed out and they are requesting a page?
+        // Grab the URL, base64_encode it and send them to the login screen.
+        $safe_refresh = ee()->cp->get_safe_refresh();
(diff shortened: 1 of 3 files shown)


intoeetive (approved):
@matthewjohns0n looks good to me, but I'm wondering about potential regression here. Should we add couple of Cypress tests here to make sure it's all good?

TomJaeger:
I agree, tests around this would be good. After a brief conversation with @intoeetive The following would be good to get tests written for (initially at least the first 4)

1. test the original issue
2. test login after logout
3. test login after removing a cookie
4. test "login as user" functionality
5. test idle login popup (that's tricky to trigger, but I have some ideas on that front)

Hat tip to Yuri on the list.

matthewjohns0n:
I added a few tests surrounding this. I wrote tests for 1, 2 and 4. 

I wasn't exactly sure how to trigger login after the idle popup, except by maybe injecting JS into the browser and forcing it to get triggered.

I also wasn't sure exactly what was meant by "test login after removing a cookie". What is the significance of this, and what cookie would we be removing? Maybe @TomJaeger  or @intoeetive can offer some more information about this.

Also, while working on this I did notice that the last login test "rejects when cookie domain is wrong" is setting the `cookie_domain` EE config setting, but then it's not getting set back properly. It becomes an issue if you test multiple times from the same device, but probably want to address this, as tests written after that would start failing.

TomJaeger:
Test login after cookie delete.. is basically setting admin session to cookie, deleting the cookie and making sure we're all good.

matthewjohns0n:
I added tests for #3 and #5:

> test login after removing a cookie
> test idle login popup

The idle login popup now is actually a collection of tests which are just different cases that someone might try if they are presented with that modal.

PR #1693, opened 2022-01-07
Fix duplicate CSS links in HTML header - load_package_css()

### LINKED ISSUE #916: CP Package Loading (js v.s. css)

The legacy library [cp](https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L14), located at `system/ee/legacy/libraries/Cp.php`,  exposes two methods for loading add-on package resources, one for JavaScript files and one for CSS files:

- [`load_package_js()`](https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L821)
- [`load_package_css()`](https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L837)

Ideally, both types of resources require the same procedure to render their contents to the output. The difference being that JavaScript files are located in a directory named `./addon/javascript/..` and have the file name extension `.js`, and CSS StyleSheet files are located in a directory `./addon/css/...` and have the file name extension `.css`. 

Both of these file types can be managed in such a way that loading a single file twice does not result in duplicate output of that file. For example, the `load_package_js()` accomplishes this by indexing the file names by the parent addon: 

https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L547-L551

This maintains a unique dataset of JavaScript files to be loaded, discouraging duplicates. The variable definition for this array can be see here:

https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L26-L32

Now looking at the `load_package_css()` method, you will notice that the package resource is automatically appended to the head (of the output). No work is done to prevent duplicate css link elements in the head of the html output. 

https://github.com/ExpressionEngine/ExpressionEngine/blob/63c0f8cc181e83aa7dbe58c3245d004c54c7bfd3/system/ee/legacy/libraries/Cp.php#L829-L849

**How To Reproduce**

```php
// even with two identical calls, duplicate output is avoided 
ee()->cp->load_package_js('main.min');
ee()->cp->load_package_js('main.min');

// two identical calls to the css loader results in duplicate css link elements in the head of the html output

ee()->cp->load_package_css('main.min');
ee()->cp->load_package_css('main.min');
```

The problem arose when loading css packages from within a custom field type. The field type accepts `grid` content, therefore, every row in the grid which had my custom field added an extra css link element to the head of the output html.  

**Environment Details:**
 - Version: Probably all, it’s in the cp legacy library

**Possible Solution**

Implement a similar technique for managing loaded css files like one that is done for JavaScript files. 

**Additional context**

Currently, JavaScript files utilize the combo loader. Allowing the combination of footer scripts to be loaded in a single call. I believe there is an identical combo loader for css files, maybe utilize that.

RK311y:
## Overview

The `ee()->cp->load_package_css()` does not prevent duplicate css links in the HTML header.

**See [CP Package Loading (js v.s. css)](https://github.com/ExpressionEngine/ExpressionEngine/issues/916) for more details.**

Resolves [#916](https://github.com/ExpressionEngine/ExpressionEngine/issues/916).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No
- [x] I think so...

### DIFF
--- a/system/ee/legacy/libraries/Cp.php
+++ b/system/ee/legacy/libraries/Cp.php
@@ -32,6 +32,8 @@ class Cp
         'fp_module' => array()
     );
 
+    public $css_files = array();
+
     /**
      * Constructor
      *
@@ -875,13 +877,17 @@ class Cp
         $current_top_path = ee()->load->first_package_path();
         $package = trim(str_replace(array(PATH_THIRD, 'views'), '', $current_top_path), '/');
 
-        if (REQ == 'CP') {
-            $url = BASE . AMP . 'C=css' . AMP . 'M=third_party' . AMP . 'package=' . $package . AMP . 'file=' . $file;
-        } else {
-            $url = ee()->functions->fetch_site_index() . QUERY_MARKER . 'ACT=' . ee()->functions->fetch_action_id('Channel', 'combo_loader') . AMP . 'type=css' . AMP . 'package=' . $package . AMP . 'file=' . $file;
+        if (! is_array($file)) {
+            $file = array($file);
         }
 
-        $this->add_to_head('<link type="text/css" rel="stylesheet" href="' . $url . '" />');
+        if (array_key_exists($package, $this->css_files)) {
+            $this->css_files[$package] = array_merge($this->css_files[$package], $file);
+        } else {
+            $this->css_files[$package] = $file;
+        }
+
+        $this->css_files = array_map('array_unique', $this->css_files);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


intoeetive:
@RK311y By looking at the code, I don't think this solves the issue you have originally posted.
It still stacks multiple `<link type="text/css"` calls and not combining those into single combo load.
And we can probably achieve the same result by just calling `array_unique` on `$this->its_all_in_your_head`
What would be your thoughts?

RK311y:
@intoeetive My thoughts are that; you are exactly right...

Moving forward, will the indexed array I have constructed in my proposed code be useful for the construction of the combo-loaded CSS `href` string (i.e. similar to that that is made in the JS combo load)? Or, should I revert this commit and start anew?

Also, would it be necessary for the CSS to have a `seal_css_combo_load()` - ([similar to this one here](https://github.com/ExpressionEngine/ExpressionEngine/blob/96ce93f5007523d4c7f7073b42f462b518b8c6d7/system/ee/legacy/libraries/Cp.php#L620)?)

intoeetive:
Honestly I'm not completely sure.

We will probably not accept this your PR in it's current form because it does not look complete. I also don't know whether `seal_css_combo_load` would be necessary. There's really a lot of stuff going on with loading CSS and JS and I never dived into that too deep.

There's also somewhat related PR opened (not fully approved yet) that refactors some of related code (I don't think it directly interfers with yours; just mentioning) https://github.com/ExpressionEngine/ExpressionEngine/pull/1259

intoeetive:
@RK311y I am going to close this draft PR, as it's not moving anywhere and we can't accept it in current form.

We still have the [discussion](https://github.com/ExpressionEngine/ExpressionEngine/discussions/3006) open that you originally created so we'll keep this request on our radars

PR #1722, opened 2022-01-14
fixed relationship field to show Add button if max value is empty, EECORE-1614

### LINKED ISSUE #1717: Relationships display, min, max issues in CP
Quoted From Slack: Found a bug on min/max and display relationships.  (love these additions, btw).   when the maximum entries to show is set to 10, minimum is set to 0 and maximum selection is left blank.. when 11 relationships are set in an entry, the “Relate Entry” button is gone.. Based on how I think this should work.. the maximum entries to show should be how many are shown in the relationship dropdown.   and maximum entries should be what limits how many can be related.

Yulyaswan:
fixes #1717 
EECORE-1614

1. Changed description text to` Maximum entries displayed in dropdown`, `Minimum selection` to  `Minimum relationships`, and `Maximum section` to `Maximum relationships`. 
2. Fixed behavior for Relate Entry button to show it if `Maximum relationships` is empty

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -80,17 +80,6 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
 
         $validator = ee('Validation')->make($rules);
 
-        $validator->defineRule('gtLimit', function ($key, $value, $parameters, $rule) use ($data) {
-            if (
-                ((isset($data['allow_multiple']) && $data['allow_multiple'] == true) || (isset($data['relationship_allow_multiple']) && $data['relationship_allow_multiple'] == 'y'))
-                && $data['limit'] < $data['rel_min']
-            ) {
-                    $rule->stop();
-                    return lang('rel_ft_min_settings_error');
-            }
-            return true;
-        });
-
         $this->errors = $validator->validate($data);
 
         return $this->errors;
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -74,7 +74,7 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
     public function validate_settings($data)
     {
         $rules = [
-            'rel_min' => 'isNatural|gtLimit',
+            'rel_min' => 'isNatural',
             'rel_max' => 'isNaturalNoZero'
         ];
 
--- a/system/ee/language/english/fieldtypes_lang.php
+++ b/system/ee/language/english/fieldtypes_lang.php
@@ -47,9 +47,9 @@ $lang = array(
 
     'rel_ft_include_future' => 'Future entries',
 
-    'rel_ft_limit' => 'Maximum entries',
+    'rel_ft_limit' => 'Maximum number of available entries',
 
-    'rel_ft_limit_desc' => 'Maximum entries displayed in dropdown. <br><i>Leave blank to allow all entries.</i>',
+    'rel_ft_limit_desc' => 'Sets the number of entries displayed in the field\'s dropdown.<br><i>Leave blank to allow all entries.</i>',
 
     'rel_ft_order' => 'Order by',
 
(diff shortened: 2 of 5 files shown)


system/ee/language/english/fieldtypes_lang.php
```diff
@@ -75,19 +75,19 @@
...
 
-    'rel_ft_min' => 'Minimum selection',
+    'rel_ft_min' => 'Minimum relationship',
 
     'rel_ft_min_desc' => 'Sets the minimum number of entries this field can be used to relate',
 
     'rel_ft_min_error' => 'You need to select at least %d entries.',
 
-    'rel_ft_min_settings_error' => 'The minimum selection cannot be set to number less than maximum entries to be shown',
+    'rel_ft_min_settings_error' => 'The minimum relationship cannot be set to number less than maximum entries to be shown',
```
intoeetive: Wondering if we'd better use plural here? E.g. `Maximum relationships`

system/ee/language/english/fieldtypes_lang.php
```diff
@@ -49,7 +49,7 @@
 
     'rel_ft_limit' => 'Maximum entries',
 
-    'rel_ft_limit_desc' => 'Maximum number of entries to show in relationship field.<br><i>Leave blank to allow all entries.</i>',
+    'rel_ft_limit_desc' => 'Maximum entries displayed in dropdown. <br><i>Leave blank to allow all entries.</i>',
```
ops-andy: Maybe revamp this to something like:  'Sets the number of entries initially displayed in the field's dropdown. A high number of entries can increase the load time of the field.'

system/ee/language/english/fieldtypes_lang.php
```diff
@@ -49,7 +49,7 @@
 
     'rel_ft_limit' => 'Maximum entries',
```
ops-andy: I think we need to update this label to reduce confusion. What about: "Number of available entries displayed"

system/ee/language/english/fieldtypes_lang.php
```diff
@@ -75,19 +75,19 @@
 
     'rel_ft_display_entry_id_desc' => 'When enabled, entry IDs will be displayed together with entry title inside the field.',
 
-    'rel_ft_max' => 'Maximum selection',
+    'rel_ft_max' => 'Maximum relationship',
```
ops-andy: possibly change to "Maximum number of related entries"

system/ee/language/english/fieldtypes_lang.php
```diff
@@ -75,19 +75,19 @@
...
 
     'rel_ft_max_desc' => 'Sets the maximum number of entries this field can be used to relate.',
 
     'rel_ft_max_error' => 'You can select no more than %d entries.',
 
-    'rel_ft_min' => 'Minimum selection',
+    'rel_ft_min' => 'Minimum relationship',
```
ops-andy: possibly change to "Minimum number of related entries"

system/ee/language/english/fieldtypes_lang.php
intoeetive: I'd not include 'initial' because AFAIK this also applies to AJAX search results

system/ee/language/english/fieldtypes_lang.php
intoeetive: I'd keep it short (however I don't quite like the existing option name either)

system/ee/language/english/fieldtypes_lang.php
ops-andy: I actually don't think we should throw this error. While I don't think it's ideal, I can find more entries than what's initially displayed by using the keyword search. I don't see a reason to require the minimum related entries to under what is initially disaplyed.

system/ee/language/english/fieldtypes_lang.php
robinsowell: I wish you could actually scroll and have more show up, so 'initially' would be correct in that case.  But that's more a feature request, and could revisit saying 'initial' then- because otherwise, I lean toward Yuri's take.

system/ee/language/english/fieldtypes_lang.php
TomJaeger: Gut reaction is that I like this suggestion.

system/ee/language/english/fieldtypes_lang.php
TomJaeger: Gut reaction is that I like this suggestion.

system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
```diff
@@ -80,17 +80,6 @@ public function validate_settings($data)
 
         $validator = ee('Validation')->make($rules);
 
-        $validator->defineRule('gtLimit', function ($key, $value, $parameters, $rule) use ($data) {
```
TomJaeger, line 83: @intoeetive Did we just drop this validation as a whole?  Is it still needed?

intoeetive: Yes. Please see @ops-andy comment: https://github.com/ExpressionEngine/ExpressionEngine/pull/1722#discussion_r786040361

TomJaeger (dismissed):

TomJaeger (approved):

PR #1736, opened 2022-01-18
Adding role_ids to session userdata

gilbert-hop:
## Overview

Adding an array of role_ids that the user is assigned to.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ x ] 🚀 Implements a new feature
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ x ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Session.php
+++ b/system/ee/legacy/libraries/Session.php
@@ -386,6 +386,7 @@ class EE_Session
 
         $this->userdata['member_id'] = (int) $member_id;
         $this->userdata['role_id'] = (int) $this->member_model->role_id;
+        $this->userdata['role_ids'] = $this->member_model->Roles->pluck('role_id');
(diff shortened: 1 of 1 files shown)


intoeetive:
Unfortunately `member_model->Roles` will not give full picture. There might be as well roles assigned via RoleGroups. Also primary role might or might not be part of Roles.

For this reason, we have `getAllRoles()` method on Member model, which can be called like `ee()->session->getMember()->getAllRoles()`

But we also prefer not to have it called when establishing `userdata` to avoid the load that might be not necessary.

If there is particular user case where this information is needed for `userdata`, please let us know by creating an issue. But we can't accept this particular PR in the form it is now

PR #1743, opened 2022-01-20
fix the bug where custom member fields were not saved 

intoeetive:
on member registration

EECORE-1626


robinsowell (approved):
ARGH, I wish I had figured that out.  Well, I wouldn't have figured out the actual line to insert, but I could have figured out exactly where it was failing.  So THAT is annoying.

But tested the fix, and it solves it, so YAY, that is NOT annoying!

TomJaeger (approved):

PR #1748, opened 2022-01-20
Fix scanned PDF mime type detection

### LINKED ISSUE #1747: Scanned PDF files aren't detected as valid
On file upload, scanned PDF files which just wraps a bitmap image aren't detected as valid for upload.

robsonsobral:
## Overview

Fix fallback detection of scanned PDF files

Resolves [#1747](https://github.com/ExpressionEngine/ExpressionEngine/issues/1747).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


TomJaeger (dismissed):

system/ee/ExpressionEngine/Library/Mime/MimeType.php
```diff
@@ -113,6 +113,10 @@ public function ofFile($path)
             $file_opening = file_get_contents($path, false, null, 0, 50);//get first 50 bytes off the file
             if (strpos($file_opening, 'RIFF') === 0 && strpos($file_opening, 'WEBPVP8') !== false) {
                 $mime = 'image/webp';
+                // PDF files start with "%PDF" (25 50 44 46) or " %PDF"
+                // @see https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Examples
+            } else if (strpos($file_opening, '%PDF') >= 0) {
```
matthewjohns0n: In this case, `strpos` is going to evaluate as `false` if not found, and `0` if found at the start of the string.  Unfortunately, `false >= 0`  evaluates to true and so does `0 >= 0` so this if statement doesn't work. 

Can you change this to `!==  false`?

matthewjohns0n (changes requested):

system/ee/ExpressionEngine/Library/Mime/MimeType.php
robsonsobral: I tested with different files. Sometime, the `%PDF` string appears at position 0 and sometimes at position 1.

matthewjohns0n: Yeah,  that's fine. The problem is that if no `%PDF` string appears it is still going to evaluate to true.

matthewjohns0n (dismissed):
Looks good now. Thanks, @robsonsobral!

TomJaeger (dismissed):

PR #1674, opened 2021-12-20
Allow to use the same template on Search Module to results and no res…

robsonsobral:
## Overview

Allow to use the same template on Search Module to results and no results

Resolves [#1319](https://github.com/ExpressionEngine/ExpressionEngine/issues/1319).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [x] 🚀 Implements a new feature
- [ ] 🛁 Refactors existing code
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔥 Removes unused files / code
- [ ] 🔒 Improves security 

## Is this backwards compatible?

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/345

PR #1750, opened 2022-01-21
Refactor `exp:search:total_results`

### LINKED ISSUE #1193: Parse {exp:search:total_results} as any other template tag
Currently. despite `{exp:search:total_results}` can be used as a variable pair, it has no difference from a single variable

I would like to be able to code something like:

```
{exp:search:total_results}
    <p>
        {total_results} found.
    </p>
    {if no_results}
        <p>No results found.</p>
    {/if}
{/exp:search:total_results}
```

robsonsobral:
## Overview

Refactor `exp:search:total_results` to be parsed as any template, allowing `no_results` and other conditionals.

Resolves [#1193](https://github.com/ExpressionEngine/ExpressionEngine/issues/1193).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


robsonsobral:
I'm not sure about the third parameter on `ee()->TMPL->parse_variables_row(ee()->TMPL->tagdata, (array) $query->row(), true);`.

system/ee/ExpressionEngine/Addons/search/mod.search.php
```diff
@@ -1013,10 +1013,18 @@ public function total_results()
...
-            return $query->row('total_results') ;
-        } else {
-            return 0;
+            if (empty(ee()->TMPL->tagdata)) {
+                return $query->row('total_results');
+            }
+
+            return $query->row('total_results') == 0
+            ? ee()->TMPL->no_results()
+                : ee()->TMPL->parse_variables_row(ee()->TMPL->tagdata, (array) $query->row(), true);
```
intoeetive: No need to do typecasting for `(array) $query->row()`, you can just use `$query->row_array()`. Also since 3rd parameter is default, it can just be skipped (which is what we usually do)

intoeetive (changes requested):
Requested small change.

I think we also need to have this added to Cypress test to make sure this does not break existing code where people has like `{if total_results = 0}`

robsonsobral:
> Requested small change.
> 
> I think we also need to have this added to Cypress test to make sure this does not break existing code where people has like `{if total_results = 0}`

@intoeetive , I don't think anybody ever used that, because it never worked. This is the reason I proposed this change.

robsonsobral:
Despite I've added the tests in the dark, I guess the fail is caused by outdated branch. I gonna check it.

intoeetive:
It looks like this PR is assuming that https://github.com/ExpressionEngine/ExpressionEngine/pull/1674 is already merged. However it is not. Maybe worth combining the two into a single PR?

robsonsobral:
I let some tests commented. They can be enabled after the merge of #1674

intoeetive:
I'm not sure commenting out part of the test is the best way to make it pass =)

Can we just make the "no results" test use separate "no results" page?

robsonsobral:
> I'm not sure commenting out part of the test is the best way to make it pass =)



Oh, I'm sorry. My intention was to not to "make them pass", but to save us from write them back after the merge. They only make sense after the merge. To write 2 forms and 2 pages to test only one feature which will work the same in both situations, looks a little too much.

But... In case you tink it's best, I can do it!

intoeetive:
If you think it's worth just combining these two PRs into one, just go ahead and do it. You have my approval on both ones, so it's just the matter of tests

robsonsobral:
DONE!

robsonsobral:

- instructions on how to fill the changelog files properly. I'm always making a mess on them;

bryannielsen (dismissed):

robsonsobral:
Do you want me to fix this conflicts, @intoeetive ?

intoeetive (changes requested):
@robsonsobral yes please, if you could resolve the conflicts that would be great (and sorry for being slow here)

Issue #1563, opened 2021-11-15
[EE6] Improvements to storage and access of Pages module data


PaulBaileyAtChan:
The Pages module currently stores data in an encoded form, within a single cell of the EE `exp_sites` table. This presents a number of significant issues:

- It is potentially a single point of failure in the case of database corruption.
- The data are not human readable/modifiable within the database itself. This is in contrast to essentially all of the per-entry (and per-template) data stored elsewhere within the database. In the general case, the ability to both view and edit information directly in the database is an important and useful fallback, for situations where the CP option is unavailable, or ineffective (for example, in situations involving DB corruption).
- It scales poorly. The size of the encoded data is proportional to the number of pages present, and all must be decoded/encoded each time, even if only a single entry is needed. This is inefficient, and works against the relational database model.

Without being too specific about how a solution might be implemented:

- Pages module data should be stored in a human readable/modifiable form.
- Data should be stored as discrete per-entry values, to allow efficient access.
- Data should be in their own table, or as a new column in a suitable, existing table which stores data per-entry.

robsonsobral:
There's a legend Pages Module was coded by a guy name Leslie and other developer in a hurry. I don't know for sure if it's true, but is totally believable.

I don't think we should have page modules anymore:

- it should be a per channel configuration. Something like **Allow URI override**;
- ~~the URI could be saved on `channel_data_titles` table~~ the URI should have its own table together with some nesting options;
- a hierarchical data field type would be fucking useful to create menus and keep pages in order. Even better, the Pages Module could be used just for this: put things in order without the need to edit entry by entry;
- a template field type (or to allow templates as source to the Select field type) would close the deal.

Too much?

PaulBaileyAtChan:
I have no idea about the origins of the Pages module, but I've always had it in the back of my mind that part of the reason Ellis Labs always pretty much actively discouraged people from using it, was that they were aware the implementation wasn't very robust.

I worry about it the way that I worried about Macromedia applications when they were taken over by Adobe, who had similar applications they were likely to prioritise (RIP Fireworks). It's understandable that Packet Tide would put more time and effort into Structure, which is more familiar to them, and is directly connected to revenue.

Pages obviously isn't the only tool for mapping URLs to content, and in most cases isn't the right one, but it is still an important part of EE, and, despite the greater attention that Structure gets these days (for obvious reasons), Pages is still the core EE method for assigning URLs to single entries. It's _long overdue_ some love.

[...]

PR #1752, opened 2022-01-21
Increase the allowed size of the page_uri

robinsowell:
The url_title limit is 200, but the current page_uri in the pages module is only 100 due to the max size setting on the form.  Bumping it up per a request and for consistency.  It's saved in an array, so no need for db change.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pages/tab.pages.php
+++ b/system/ee/ExpressionEngine/Addons/pages/tab.pages.php
@@ -90,7 +90,7 @@ class Pages_tab
                 'field_required' => 'n',
                 'field_data' => $pages_uri,
                 'field_text_direction' => 'ltr',
-                'field_maxl' => 100,
+                'field_maxl' => 200,
                 'field_instructions' => '',
                 'field_placeholder' => lang('example_uri')
             ),


PaulBaileyAtChan:
Can I gently suggest that any tweaks to Pages module storage really should be part of a wider rethink of how it's done. See https://github.com/ExpressionEngine/ExpressionEngine/issues/1563

intoeetive (approved):

TomJaeger:
@PaulBaileyAtChan I agree.  In this case, we're really making the pages URI line up with the length of a url_title, to bring it more in line with expected behavior.

We're really not planning to move pages forward, without larger changes. :D

PaulBaileyAtChan:
Okay, not going to keep hammering this, except to say that:

- Doubling the allowable length of the Pages URI potentially puts greater strain on the existing, all-in-one storage method. Because url_title data are stored separately per-entry, longer values aren't such a strain on how those are stored, so longer values are more justifiable.
- Because url_title data and Pages URI data represent different things (the url_title is just a unique string identifier for the entry, which *might* be used in the front-end as part of a URL, but doesn't have to be; and the Pages URI is specifically intended to be a complete path to the page), it's not obvious why their lengths should be the same, and it's not especially inconsistent for them not to be.

PR #1756, opened 2022-01-24
fix issue when models could not be cached

intoeetive:
due to error in `Serializable` interface implementation

EECORE-1630

### DIFF
--- a/system/ee/ExpressionEngine/Library/CP/URL.php
+++ b/system/ee/ExpressionEngine/Library/CP/URL.php
@@ -186,13 +186,17 @@ class URL implements \Serializable
     }
 
     /**
-     * Serializes the object by JSON encoding
-     *
-     * @return string A JSON encoded string.
+     * Serializes the object
      */
     public function __serialize()
     {
-        return $this->serialize();
+        return array(
+            'path' => $this->path,
+            'session_id' => $this->session_id,
+            'qs' => $this->qs,
+            'base' => $this->base,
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Library/CP/URL.php
```diff
@@ -205,17 +209,29 @@ public function serialize()
             'requested_uri' => $this->requested_uri
         );
 
-        return json_encode($serialize);
+        return serialize($serialize);
```
bryannielsen: @intoeetive does it make sense to dry this up and only build the array in `__serialize()`.  I think this could just be `return serialize($this->__serialize());`

system/ee/ExpressionEngine/Library/CP/URL.php
```diff
@@ -225,7 +241,7 @@ public function __unserialize($serialized)
      */
     public function unserialize($serialized)
     {
-        $data = json_decode($serialized, true);
+        $data = unserialize($serialized);
 
         if (! is_array($data)) {
```
bryannielsen, line 234: @intoeetive similarly I think we could dry this up by just doing `return $this->__unserialize($data);` after line 244

bryannielsen (commented):
There are just a few places where it looks like we could reduce some code duplication.  Let me know if you think it's not a good idea to consolidate them though.

bryannielsen (approved):

PR #1765, opened 2022-01-26
make the var_dump output for models cleaner

intoeetive:
EECORE-1646


bryannielsen (commented):
@intoeetive this looks like a nice approach to cleanup the var_dump output and lower impact than what we discussed before.  The only thing I'm noticing is the variable name `$footptint` is that supposed to be `$footprint`?

intoeetive:
@bryannielsen - good catch, fixed! Apparently I started copypasting without reading :)

bryannielsen (approved):

PR #1781, opened 2022-01-31
output optimization

intoeetive:
always send ajax requests as application/json 

deprecate user_agent library as it's too old and not used
 
do not refresh CSRF tokens if those are disabled

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -1642,7 +1642,7 @@ GRID_FALLBACK;
         }
 
         if ($this->json) {
-            return $this->send_ajax_response(
+            return ee()->output->send_ajax_response(
                 array(
                     'success' => (empty($this->errors) && empty($this->field_errors)) ? 1 : 0,
                     'errors' => (empty($this->errors)) ? array() : $this->errors,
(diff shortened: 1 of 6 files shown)


system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
```diff
@@ -1642,7 +1642,7 @@ function ($v) {
         }
 
         if ($this->json) {
-            return $this->send_ajax_response(
+            return ee()->output->send_ajax_response(
```
bryannielsen, line 1645: Since we're changing the call here should we deprecate or remove the `Channel_form_lib@send_ajax_response()` method? https://github.com/ExpressionEngine/ExpressionEngine/blob/d9157e68ccf7ff8c11820f2b94e1481bb050c1b1/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php#L2920-L2938 It looks like a duplicate of what the Output class is doing, but it is public and possibly used by third parties.  Should we at least bring it inline with the Output class and remove the user agent detection and conditional?

bryannielsen (commented):
I searched through the code for other mentions of the User Agent library and couldn't find any.  I think we should be fine to deprecate and give third party developers time to remove references if they're still using it.

system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
intoeetive: It might be good to deprecate it, but `ee()->output->send_ajax_response` is also using legacy `Output` class which will might be replaced with something else in EE7. I'm not sure how good it to deprecate in favor of something that we expect to eventually be gone as well. If anyone uses this function, they would still see deprecation message related to `user_agent`

matthewjohns0n (approved):

PR #1783, opened 2022-02-01
Fixed bad path to native plugins

robinsowell:
After updates, frontend throws
```
require_once(ee/ExpressionEngine/Addons/pi.xml_encode.php): failed to open stream: No such file or directory

```
Why this wasn't happening to everyone, not entirely sure.  I've had exactly one report of it.  But that's an old legacy path from when plugins weren't in folders.

### DIFF
--- a/changelogs/patch.md
+++ b/changelogs/patch.md
@@ -55,6 +55,7 @@ Bullet list below, e.g.
    - Fix #1195 improves `template_post_parse` hook when fired after sub or layout (i.e. partial) template parse completes
    - Fix WEBP Mime type detection
    - Fix channels being ignored on search module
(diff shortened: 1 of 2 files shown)


intoeetive (dismissed):

TomJaeger:
So I took a peek at this... I'm guessing that the  parse_plugin function is really only present to support add-ons that might be looking to parse a plugin's contents before returning results etc.

@intoeetive did you size this up the same way?

intoeetive:
I *think* (not tested) that this has something to do with plugins providing custom text formatting functionality https://docs.expressionengine.com/latest/development/legacy/libraries/typography.html#using-a-plugin-for-text-formatting

I also think that the fix is correct

PR #1796, opened 2022-02-08
changed Entries Listing Notifications position, EECORE-1658

Yulyaswan:


intoeetive (commented):
It looks like the container is taking the space whether the banner is present or not, is that correct?

Yulyaswan:
no, it doesn't have extra space if the notification is empty - https://share.getcloudapp.com/E0uoJLz1
they only appear when a notification appears - https://share.getcloudapp.com/o0uZOJky

intoeetive (approved):

PR #1821, opened 2022-02-15
Resolved #1808 where `base_` variables from **Query module** aren't parsed

### LINKED ISSUE #1808: Parse `base_url` on Query module
When using the query module, some data using `{base_url}` returns.

We could have a parameter `parse_urls` on `exp:query`.

Jesus... The only thing I can think of is to **not** use `{base_url}` on settings.

The row being parsed should have a column named `site_id` for that to work. And the parser need to run row by row.

robsonsobral:
## Overview

Parse `{base_` variables on query module and add `parse_variables_query_results_by_default` to config overrides

Resolves [#1808](https://github.com/ExpressionEngine/ExpressionEngine/issues/1808).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/398

### DIFF
--- a/system/ee/ExpressionEngine/Addons/query/mod.query.php
+++ b/system/ee/ExpressionEngine/Addons/query/mod.query.php
@@ -63,7 +63,7 @@ class Query
 
                 $chunk = ee()->TMPL->parse_switch($chunk, $count - 1);
 
-                if (strpos($chunk, '{base_') !== false) {
+                if (strpos($chunk, LD . 'base_') !== false) {
                     $chunk = isset($row['site_id'])
                         ? parse_config_variables(
                             $chunk,
@@ -77,7 +77,7 @@ class Query
             $parsed = ee()->TMPL->parse_variables(ee()->TMPL->tagdata, $results);
         }
 
-        if (get_bool_from_string(ee()->TMPL->fetch_param('parse_files', config_item('parse_variables_query_results_by_default'))) && strpos($parsed, '{filedir_') !== false) {
+        if (get_bool_from_string(ee()->TMPL->fetch_param('parse_files', config_item('parse_variables_query_results_by_default'))) && strpos($parsed, LD . 'filedir_') !== false) {
             ee()->load->library('file_field');
             $parsed = ee()->file_field->parse_string($parsed);
         }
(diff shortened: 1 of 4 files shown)


robsonsobral:

- the `parse_variables_query_results_by_default` maybe could have a better name;
- the variable config override should have its default value changed to `y` for release of version 7;

robsonsobral:
I guess we can go ahead with this PR, @TomJaeger .

system/ee/ExpressionEngine/Addons/query/mod.query.php
```diff
@@ -50,15 +50,40 @@ public function __construct()
...
+                        )
+                        : parse_config_variables($chunk);
+                }
+                $parsed .= $chunk;
+            }
+        } else {
+            $parsed = ee()->TMPL->parse_variables(ee()->TMPL->tagdata, $results);
+        }
+
+        if (get_bool_from_string(ee()->TMPL->fetch_param('parse_files', config_item('parse_variables_query_results_by_default'))) && strpos($parsed, LD . 'filedir_') !== false) {
```
intoeetive, line 80: Shall we not get same (or better) result if we just apply `ee()->TMPL->parse_globals($str)` at the end?

intoeetive (changes requested):
@robsonsobral please see my comment on code

It also needs a conflict to be resolved (we have removed changelog md files as those don't really help but only cause conflicts. Here's what we'll try to use instead: https://github.com/ExpressionEngine/ExpressionEngine/blob/6.dev/CONTRIBUTING.md#documentation )

Sorry it all is taking so long! I'll set the PR as draft for now

system/ee/ExpressionEngine/Addons/query/mod.query.php
robsonsobral: @intoeetive , I can't see the point. [`parse_globals()`](https://github.com/ExpressionEngine/ExpressionEngine/blob/1b8071562a910323032f6a006a7973ddd63f6a9a/system/ee/legacy/libraries/Template.php#L3012) looks a total overkill for this case. It looks for a lot of things we can't get from database, except in case we query for templates themselves, which could create a total mess.

Even more, how would we get data from different sites? I'm using a column `site_id` to check for the right value.

PR #1857, opened 2022-02-28
Add check for location and override

### LINKED ISSUE #1856: `URL` query string allows for unwanted bot traffic redirects from front end
## Description of the problem

ExpressionEngine currently has a query string (`?URL=`) that is checked with all network calls. This currently exists with both Control Panel calls and front end calls, and is used by various parts of the backend.

When in use by the backend of an EE site, it is assumed this would be fine behavior. However, the front end query string use can currently be used by any user, and can be utilized by bot traffic.

Example: [https://nerdcore.codes/?URL=example.com](https://nerdcore.codes/?URL=example.com)

EE 6.2.x and 5.4.x introduced a robots NOFOLLOW NO INDEX for these redirects. However, it does continue to provide behavior that is not ideal and can be manipulated by bots and bad behavior users. It also allows for a simple screen to redirect unsuspecting users.

## How To Reproduce

Go to [https://nerdcore.codes/?URL=example.com](https://nerdcore.codes/?URL=example.com)

## Recommended Solution

- Disallow by default these types of redirect from the frontend
- Add config override to enable these types of redirects

dougblackjr:
## Overview

Disallows use of the `?URL=` query string from the front end for redirecting.

Adds `allow_url_redirects_from_site` config override that allows user to override new default behavior.

## Example

With this PR in place, this will automatically redirect anyone attempting to access the `URL` query string from the front end to a 404.

Screenshot from 2022-02-28 12-21-10

Adding `$config['allow_url_redirects_from_site'] = 'y';` to the config will allow for the URL to still be accessible from the frontend:

Screenshot from 2022-02-28 12-22-05

Resolves [#1856](https://github.com/ExpressionEngine/ExpressionEngine/issues/1856).

## Nature of This Change

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [X] 🔒 Improves security

- [X] Yes
- [ ] No

## Documentation
User Guide Pull Request: incoming once approved


aasun:
This is great.. Working for me.

dougblackjr:
Per conversation, adds secondary check for whether user has an active CP setting, with appropriate config override:

```
$config['allow_url_redirects_from_site'] = 'y';
$config['allow_url_redirects_from_site_without_cp_session'] = 'y';
```

intoeetive (changes requested):
I don't think `access_cp` is part of session userdata at all. Instead, `ee()->session->access_cp` should be used (but also there's some chance `ee()->session` is not set at this point at all, so that needs to be checked too

I also don't recall discussion 2 separate config overrides, but that might be just my memory leaking.

dougblackjr:
@intoeetive Pushed up a fix for the session data.

RE: The second override, added that in case someone wanted to allow URL redirects even if the user does not have a CP session. I can pull that out, whatever you think is best.

intoeetive:
ok, I've modified code somewhat and removed the second override - IMO it's redundant

@bryannielsen please review

bryannielsen (approved):

matthewjohns0n (approved):

intoeetive (approved):

PR #1866, opened 2022-03-02
fix the issue where PHP notices were shown in CP during MFA auth

intoeetive:
EECORE-1697


robinsowell (approved):
Checks out on install that was throwing the error, no errors, no issues.

PR #1902, opened 2022-03-18
Added possibility to defer initialization of Relationship field

### LINKED ISSUE #1882: React components result in long first time to interaction
We’re seeing some extremely slow CP entry publish page page load times, mostly regarding browsers time to interaction in the CP. We have a page with 10 Relationship fields on it (each inside of Bloqs), and when a page in the CP is done rendering the content from the server, it hangs for a few seconds, sometimes upwards of 10 seconds or more before a user can actually interact with the fields in the publish form. As soon as I commented out these lines in the relationship.js file, it got _noticeably_ better. I have a feeling if you did the same thing with Fluid you would see similar results. Wygwam has a deferred initialization option (basic html display by default, then initialize the React field when the user clicks on it). I feel like the Relationship field, maybe even every React based fields, could benefit from a similar option. It could be worth exploring showing the selected relationship options, then when the user clicks on the "Relate Entry" button, it initializes the React component and builds it on demand.

```
// $(document).ready(function () {
//   Relationship.renderFields();
// });
// Grid.bind("relationship", "display", function (cell) {
//   Relationship.renderFields(cell);
// });
// FluidField.on("relationship", "add", function (field) {
//   Relationship.renderFields(field);
// });
```

We had a similar issue with a custom fieldtype we made using Select2. At first it was a few hundred options in the list, so it was fast. As the site grew, we’re into a couple thousand options in each Select2 field, and adding a few of these fields to the page (via Bloqs in this case), and it was noticeably slower. We got the same delay in time to interaction. I since refactored it to load all the options via ajax so there wasn’t a few thousand extra DOM elements in the page that the Select2 or other JS/jQuery scripts had to parse through. It was a big performance boost, but now we’re seeing similar things happen with the Relationship field. The React components just take forever to fully initialize when there are multiples on the page.

Add 10-15 Relationship fields to an entry and I'm guessing you'll see similar long times to interaction. E.g. the response is back from the server, but you won't be able to actually focus your mouse into any form fields, click any tabs to toggle display of content etc. This happens regardless of the browser used.

You can see the renderFields() function taking exceptionally long time to execute:

timeline

(removed video b/c it contained sensitive info. I'll send it via Slack) And the timeline in action. Note around the 17 second mark (yes, the server side response takes that long for various reasons), you'll see it hang for several seconds. Even after it responds again and "completes" the request, It's several more seconds until I can focus my mouse into the Title field.

litzinger:
## Overview

Creates new “Edit Relationships” button, which when clicked, will instantiate the React component on the field, thus deferring the initial React bootup on page load. When many Relationship fields are in an entry, it can cause the page to “hang” and create a delayed time to interaction for the user. This greatly decreases the time to interaction because it removes nearly all the JS events and interactions for every Relationship field until the user chooses to interact with it.

See screenshots and video in #1882 for the result of this PR.

Resolves [#1882](https://github.com/ExpressionEngine/ExpressionEngine/issues/1882).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -787,6 +787,7 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
         // Boolstring conversion
         $save['allow_multiple'] = get_bool_from_string($save['allow_multiple']);
         $save['display_entry_id'] = get_bool_from_string($save['display_entry_id']);
+        $save['deferred_loading'] = get_bool_from_string($save['deferred_loading']);
 
         foreach ($save as $field => $value) {
             if (is_array($value) && count($value)) {
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -509,6 +509,7 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
         }
 
         return ee('View')->make('relationship:publish')->render([
+            'deferred' =>  isset($this->settings['deferred_loading']) ? $this->settings['deferred_loading'] : false,
             'field_name' => $field_name,
             'choices' => $choices,
             'selected' => $selected,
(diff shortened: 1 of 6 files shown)


Yulyaswan:
Сhanges in the php file fixed the ability to turn off the deferred toggle button
Previously: after saving changes to the Relationship field, the button was always `on`.
Now user can save his changes 

Changes in the js file check and render previously saved items and hide `“Edit Relationships”` button (switch it to `"Relate Entry"` button or hide all buttons if no more items can be added)

Yulyaswan:
EECORE-1710

Yulyaswan (dismissed):

litzinger:
Thank you for addressing the Cypress tests... I wasn't sure how to do that :) I've had mixed results getting those tests running locally :(

intoeetive (approved):

intoeetive (approved):

PR #1917, opened 2022-03-24
added short name to the first column for conditional fields, EECORE-1740, EECORE-1739

Yulyaswan:


system/ee/ExpressionEngine/View/_shared/form/condition/condition-rule-group.php
```diff
@@ -13,7 +13,7 @@
 $conditionFieldArr = [];
 
 foreach ($fieldsList as $fieldLabel) {
-    $conditionFieldArr[$fieldLabel['field_id']] = $fieldLabel['field_label'];
+    $conditionFieldArr[$fieldLabel['field_id']] = $fieldLabel['field_label'].'<span class="short-name">{'.$fieldLabel['field_name'].'}</span>';
```
intoeetive: This needs spaces around `.` for PSR-12 compatibility

intoeetive (changes requested):

intoeetive (approved):

PR #1919, opened 2022-03-24
Wraped Revisions table in panel, EECORE-1744

Yulyaswan:


ajohnson (dismissed):

ajohnson (approved):
@Yulyaswan Just noticed the Auto-Saves table was in the same situation, so I added a panel to it as well.

PR #1940, opened 2022-03-29
Allow loading application core outside of request run

bryannielsen:
This change allows calling `loadApplicationCore()` outside of the `run()` method.  It also caches the loaded `$application` so that it does not break any future calls to `run()`.

### DIFF
--- a/system/ee/ExpressionEngine/Core/Core.php
+++ b/system/ee/ExpressionEngine/Core/Core.php
@@ -26,7 +26,7 @@ abstract class Core
     protected $booted = false;
 
     /**
-     * @var bool Application instance
+     * @var \ExpressionEngine\Core\Application Application instance
      */
     protected $application = null;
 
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Core/Core.php
```diff
@@ -306,8 +311,12 @@ public function setTimeLimit($t)
     /**
      * Setup the application with the default provider
      */
-    protected function loadApplicationCore()
+    public function loadApplicationCore()
     {
+        if ($this->application) {
```
intoeetive: Should this be `!is_null($this->application)` instead? Maybe we should not rely on object to evaluate to true. But - we're doing that in lots of other places, and it's probably safe to do.

intoeetive (dismissed):

system/ee/ExpressionEngine/Core/Core.php
bryannielsen: @intoeetive I like that, I've pushed a change for this.

intoeetive (approved):

PR #1955, opened 2022-03-30
Fix toggle off working

matthewjohns0n:
EECORE-1771


robsonsobral:
Shouldn't we always use the `get_bool_from_string()` helper? ~~Maybe to add `0` and `1` to it.~~

intoeetive:
@robsonsobral I think you're actually right. We have separate check for `null` so it's safe to just reuse that function.

@matthewjohns0n I'm closing your PR in favor of other one that I submitted - please review https://github.com/ExpressionEngine/ExpressionEngine/pull/1957

matthewjohns0n:
Good call @robsonsobral. And sounds good @intoeetive - I will review that PR now.

PR #1971, opened 2022-04-05
URL title and filename for SEO guidelines

ferditarakci:
Hi,

Filenames and URL title formatted with Latin characters in accordance with SEO guidelines


robsonsobral:
Shouldn't this use the [Foreign Characters config](https://github.com/ExpressionEngine/ExpressionEngine/blob/6.dev/system/ee/ExpressionEngine/Config/foreign_chars.php)? Or to update them?

intoeetive (changes requested):
@ferditarakci do I understand correctly that the purpose of this change is to automatically rename the files that contain non-Latin characters to contain Latin only?

I agree with Robson that this should use `foreign_chars` config (which might need an update? not sure). And also this behavior should probably be configurable - I imagine not everyone will want automatic file renaming

I am turning this PR into draft for now

PR #1977, opened 2022-04-06
Feature/regex fixes

### LINKED ISSUE #1083: Open javascript in template triggers php strpos error
Was getting errors in EE5 and EE6 sites citing some version of 

> Warning
> strpos(): Empty needle
> ee/legacy/libraries/channel_entries_parser/Parser.php, line 297
> Severity: E_WARNING

After much digging around on one site at least I have isolated the cause to be the inclusion in the template of this block of included javascript

```
    var zf_DateRegex = new RegExp(
        "^(([0][1-9])|([1-2][0-9])|([3][0-1]))[-]([0][1-9]|1[012])[-](?:(?:19|20)[0-9]{2})$");
    var zf_MandArray = ["Name_First", "Name_Last", "SingleLine", "PhoneNumber_countrycode", "Email",
        "Address_ZipCode", "Dropdown", "Radio", "SingleLine2"
    ];
    var zf_FieldArray = ["Name_First", "Name_Last", "SingleLine", "PhoneNumber_countrycode", "Email",
        "Address_AddressLine1", "Address_AddressLine2", "Address_City", "Address_Region", "Address_ZipCode",
        "Dropdown", "Radio", "SingleLine1", "ImageUpload", "SingleLine2"
    ];
    var isSalesIQIntegrationEnabled = false;
    var salesIQFieldsArray = [];

```
Turns out the error is being caused by presence of the line 

```
    var zf_DateRegex = new RegExp(
        "^(([0][1-9])|([1-2][0-9])|([3][0-1]))[-]([0][1-9]|1[012])[-](?:(?:19|20)[0-9]{2})$");
```

The javascript is OK - if you put the same block into an externally loaded script the error goes away and template is processed without issue.  So supposition is that for some reason EE's template parser is trying to evaluate the line.

Issue happens repeatably on an EE6.0.3 setup, on Apache server running php 7.4.16

### LINKED ISSUE #1522: Validation Service - regex rule cannot contain character ranges
The [Validation Service](https://docs.expressionengine.com/latest/development/services/validation.html#validation-service) includes the ability to define a rule using regex.

The syntax for defining a regex rule includes the use of square brackets.
The syntax for regex uses square brackets to define character ranges.

If you need to combine these two bits of syntax, the Validation Service fails to interpret the resulting regex rule correctly.

This is a valid regex rule (e.g. it works fine as an HTML 5 validation rule): `[a-zA-Z0-9-'\s.]`
When you try to use this rule with the Validation service you end up with a rule definition that looks like this:
```
'First_Name' => "regex[/[a-zA-Z0-9-\'\s.]/]",
```
When this is parsed by the Validation Service the rule read terminates at the first `]` so the rule seen by the validation service becomes `[a-zA-Z0-9-'\s.` which produces very different results to the correctly parsed rule.

The Validation Service regex parser will need to be updated to correctly handle `]` within the regex.

### LINKED ISSUE #1852: Form validation string triggers deprecation warning from EE Channel Entries Parser
I have a site with a form that uses HTML5 validation.
One of these rows tests for a UK Post Code value (e.g. of the form AB1 2CD) - the regex to test for this is 
```
^([a-zA-Z]){1}([0-9][0-9]|[0-9]|[a-zA-Z][0-9][a-zA-Z]|[a-zA-Z][0-9][0-9]|[a-zA-Z][0-9]|[0-9][a-zA-Z]){1}([ ])?([0-9][a-zA-z][a-zA-z]){1}$
```
So the form input looks like this... 
```
<input 
    class='form-control w-100' 
    type='text' 
    id='Post_Code' 
    name='Post Code' 
    maxlength='15' 
    required pattern="^([a-zA-Z]){1}([0-9][0-9]|[0-9]|[a-zA-Z][0-9][a-zA-Z]|[a-zA-Z][0-9][0-9]|[a-zA-Z][0-9]|[0-9][a-zA-Z]){1}([ ])?([0-9][a-zA-z][a-zA-z]){1}$"
>
```
When this element is included in the form, I get a deprecation warning from the EE Channel Entries parser thus:
```
Deprecated
strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior

ee/legacy/libraries/channel_entries_parser/Parser.php, line 297

    Severity: E_DEPRECATED
```
As you will note, the form input value concerned does not contain any EE template variables (but does sit within a Channel Entries loop).

Removing the three `{1}` elements from the validation regex fixes the issue (but breaks the regex).

This is a bit of overreach by the Channel Entries parser I think...

EE 6.2.6 running on php 7.4.

intoeetive:
fix the error when `[]` could not be used in `regex` validation rule
fixes #1522
EECORE-1790

 
fix the issue when HTML form validation could cause PHP notices from channel entries parser
fixes #1852
fixes #1083
EECORE-1791

### DIFF
--- a/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
+++ b/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
@@ -264,7 +264,7 @@ class LegacyParser
             // date variables
             elseif (strpos($val, 'format') !== false && preg_match("/.+?\s+?format/", $val)) {
                 $var_single[$val] = $this->extractDateFormat($val);
-            } else {  // single variables
+            } elseif (!is_numeric($val)) {  // single variables
                 $var_single[$val] = $val;
             }
         }
--- a/system/ee/ExpressionEngine/Service/Validation/Validator.php
+++ b/system/ee/ExpressionEngine/Service/Validation/Validator.php
@@ -295,7 +295,7 @@ class Validator
      */
     protected function parseRuleString($string)
     {
-        if (preg_match("/(.*?)\[(.*?)\]/", $string, $match)) {
+        if (preg_match("/(.*?)\[(.*?)\]$/", $string, $match)) {
             $rule_name = $match[1];
             $parameters = $match[2];
 


system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
```diff
@@ -264,7 +264,7 @@ public function extractVariables($tagdata, $target = null)
             // date variables
             elseif (strpos($val, 'format') !== false && preg_match("/.+?\s+?format/", $val)) {
                 $var_single[$val] = $this->extractDateFormat($val);
-            } else {  // single variables
+            } elseif (!is_numeric($val)) {  // single variables
                 $var_single[$val] = $val;
```
robsonsobral, line 268: Isn't an `else` missing here?

intoeetive: No - we want to skip parsing if the variable is not matching any of rules

bryannielsen (approved):

PR #1979, opened 2022-04-06
Skip validation for custom fields hidden by conditional logic

bryannielsen:
## Overview

Before this fix it was possible to get validation errors due to validation rules on a custom field that was not visible due to conditional logic.  This fix just moves the check for whether or not the field is hidden by conditional logic to before any validation rules are added for the field.

The example where this bug was found involves setting up a Grid field with conditional logic and `Minimum Rows: 1` and at least one column that is **Required**.  Even if the Grid field is not visible due to the conditional logic it will throw a validation error because the required column fails validation.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/ExpressionEngine/Model/Content/ContentModel.php
+++ b/system/ee/ExpressionEngine/Model/Content/ContentModel.php
@@ -258,7 +258,9 @@ abstract class ContentModel extends VariableColumnModel
         $facades = $this->getCustomFields();
 
         foreach ($facades as $name => $facade) {
-            if (! $this->isNew() && ! $this->isDirty($name)) {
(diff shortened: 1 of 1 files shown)


intoeetive (approved):
I wasn't able to replicate the error.

However, to me this change makes sense - if the field is hidden, all validation checks should be skipped, so approving

TomJaeger:
Fix worked for me.

PR #1980, opened 2022-04-07
display the conditional field when ANY of sets is matched

intoeetive:
This change is (almost) reversing the logic on conditional sets.

Previously, the field was hiding if any of the conditional sets checks was returning false before the positive checks (the sets order did matter). This made it not possible to use different sets for different channels (field combinations)

With this change, the field is only hiding if ALL sets checks are negative (the order does not matter). If at least one set returns true, the field is displayed


intoeetive:
As per Slack conversation, this is working as expected already

https://packettide.slack.com/archives/C030Y8HG9L3/p1649332591536259

PR #1984, opened 2022-04-07
hide global field settings for notes, EECORE-1795

Yulyaswan:


intoeetive (changes requested):
I think "include in search" and "hide field" toggles should also be hidden here.

Also there's high chance that once implemented we'll need similar changes for other fields (e.g. remove "include in search" from relationship fields). So probable makes sense to have this script in a file that's always included on Fields controller, not specific `notes.js`

matthewjohns0n:
This is definitely moving in the right direction. I do think we should allow add-ons to define what fields to hide though, that way any add-on could use this feature by defining some sort of "hide settings" array.

intoeetive (approved):
I think it's good for now. We can improve it later on

matthewjohns0n (approved):

PR #1991, opened 2022-04-11
fix the issue where searching entries by `entry_id` in CP did not wor…

intoeetive:
…k when entering less than 3 digits

EECORE-1593

### DIFF
--- a/system/ee/ExpressionEngine/Service/EntryListing/EntryListing.php
+++ b/system/ee/ExpressionEngine/Service/EntryListing/EntryListing.php
@@ -345,7 +345,11 @@ class EntryListing
                     break;
             }
 
-            $entries->search($search_fields, $this->search_value);
+            if (is_numeric($this->search_value) && strlen($this->search_value) < 3) {
+                $entries->filter('entry_id', $this->search_value);
+            } else {
+                $entries->search($search_fields, $this->search_value);
+            }
         }
 
         $filter_values = $this->filters->values();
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Service/EntryListing/EntryListing.php
```diff
@@ -345,7 +345,11 @@ protected function setupEntries()
                     break;
             }
 
-            $entries->search($search_fields, $this->search_value);
+            if (is_numeric($this->search_value) && strlen($this->search_value) < 3) {
```
matthewjohns0n, line 348: So will this fix only work for entry IDs less than 100?

intoeetive: If it's greater than 999 the regular search will work (in all fields), as it was previously

bryannielsen: This is strlen < 3 so less than 100 unless this is an error and we want `strlen <= 3`

intoeetive: No, with 3 character we already use "normal search"

intoeetive: Previously this was based on rule set up in javascript `>= 3`
Now we also want to expand search to numbers `<3`

So if the search keyword is integer, we accept any number of characters, and search by entry ID only. If it's `>=3` (so `! < 3`) we search in all fields

matthewjohns0n (approved):

PR #1997, opened 2022-04-12
Resolved #1996 where slider as CP component did not have updated styling

### LINKED ISSUE #1996: Slider fieldtype rendering incorrectly as CP component
I just noticed in 6.3 that the slider field, when used inside a settings array on a fieldtype settings page does not render correctly. This is the code I've been using for quite awhile, and it renders the field as seen in the screenshot below, but it does not indicate to the user what the value is. Screenshots are from Safari, but I tested in Chrome too, same issue.

```
'fields' => [
                    'blockdefinition_nesting[exact_children]' => [
                        'required' => false,
                        'type' => 'slider',
                        'unit' => '',
                        'min' => 0,
                        'max' => 25,
                        'value' => $exactChildren,
                    ],
                ]
```
 
Screen Shot 2022-04-12 at 4 47 28 AM

For comparison, this is what renders with the same code in 6.2.7

Screen Shot 2022-04-12 at 4 51 35 AM

intoeetive:
fixes #1996

EECORE-1807

### DIFF
--- a/system/ee/ExpressionEngine/View/_shared/form/field.php
+++ b/system/ee/ExpressionEngine/View/_shared/form/field.php
@@ -75,17 +75,6 @@ case 'short-text': ?>
         <?php endif;?>
     </label>
     <?php break;
-case 'slider':
-    if ($class): ?>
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):
This looks good, tested on the Spam addon settings page

PR #2016, opened 2022-04-14
Improved UI for editing system message templates

Yulyaswan:
Added panel wrapper to system message templates, EECORE-1792

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Design/System.php
+++ b/system/ee/ExpressionEngine/Controller/Design/System.php
@@ -151,7 +151,7 @@ class System extends AbstractDesignController
 
         $this->loadCodeMirrorAssets();
 
-        ee()->view->cp_page_title = sprintf(lang($template->template_name));
+        ee()->view->cp_page_title = lang($template->template_name);
         ee()->view->save_btn_text = lang('save');
         ee()->view->save_btn_text_working = 'btn_saving';
 
--- a/system/ee/ExpressionEngine/Controller/Design/System.php
+++ b/system/ee/ExpressionEngine/Controller/Design/System.php
@@ -132,9 +132,28 @@ class System extends AbstractDesignController
             'author' => (empty($author)) ? '-' : $author->getMemberName(),
         );
 
+        $vars['buttons'] = [
+            [
+                'name' => 'submit',
+                'type' => 'submit',
+                'value' => 'save',
+                'text' => 'save',
+                'working' => 'btn_saving'
+            ],
+            [
+                'name' => 'submit',
+                'type' => 'submit',
+                'value' => 'save_and_close',
+                'text' => 'save_and_close',
+                'working' => 'btn_saving'
+            ]
+        ];
+
         $this->loadCodeMirrorAssets();
 
-        ee()->view->cp_page_title = sprintf(lang('edit_template'), lang($template->template_name));
+        ee()->view->cp_page_title = sprintf(lang($template->template_name));
+        ee()->view->save_btn_text = lang('save');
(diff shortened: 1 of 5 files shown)


system/ee/ExpressionEngine/Controller/Design/System.php
```diff
@@ -132,9 +132,28 @@ public function edit($template_id)
...
+                'value' => 'save_and_close',
+                'text' => 'save_and_close',
+                'working' => 'btn_saving'
+            ]
+        ];
+
         $this->loadCodeMirrorAssets();
 
-        ee()->view->cp_page_title = sprintf(lang('edit_template'), lang($template->template_name));
+        ee()->view->cp_page_title = sprintf(lang($template->template_name));
```
intoeetive: `sprintf` is missing parameter here. I think we could just remove it and keep only `lang($template->template_name)`

system/ee/ExpressionEngine/View/design/system/index.php
```diff
@@ -2,18 +2,19 @@
...
-  <div class="panel-heading">
+	<div class="panel-heading">
 		<div class="title-bar">
 			<h3 class="title-bar__title"><?=$cp_heading?></h3>
 		</div>
-  </div>
-
-    <div class="app-notice-wrap"><?=ee('CP/Alert')->getAllInlines()?></div>
+	</div>
+	<div class="design-system-template-body">
```
intoeetive: Why do we need this new element? Shouldn't we place the table and alert into `panel-body` instead? Also, the indents for alert don't look right currently in Chrome
image

intoeetive (changes requested):

system/ee/ExpressionEngine/Controller/Design/System.php
Yulyaswan: changed

system/ee/ExpressionEngine/View/design/system/index.php
Yulyaswan: `.design-system-template-body` was added as an additional class for the main `system message templates`
`.single-templates-notice-wrapper` was added as an additional class for the single `system message templates` to fix spaces around the notice message.
if place the table and alert into `panel-body` it will break the current design:
Screenshot 2022-05-04 at 10 38 11

intoeetive: Then maybe question for @ajohnson 

Shouldn't we use same UI here which is there for the template, template variables and partials? I.e. with the left sidebar.

Also it looks like Email templates and Member templates (if enabled) need similar updates

ajohnson: @intoeetive @Yulyaswan Correct, the `panel-body` should not be used when displaying a table list like the System Message Templates.

So the only inconsistency I'm noticing is the placement of the `app-notice` confirmation. When saving a template we place the confirmation within the `panel-heading` above the title like this:

Screen Shot 2022-05-18 at 9 50 33 AM

But with the System Message Templates we're displaying the `app-notice` confirmation below the `panel-heading` like this:

Screen Shot 2022-05-18 at 9 50 48 AM

So it'd be nice to change the location of the System Message Templates `app-notice` to be displayed in the `panel-heading` to match Templates.

But let me know if I'm misunderstanding your question @intoeetive . Thanks!

Yulyaswan: @ajohnson I moved notification block above the title
https://packettide.atlassian.net/browse/EECORE-1792?focusedCommentId=16204

ajohnson (approved):

intoeetive (approved):

PR #2030, opened 2022-04-19
templates for matrix getting warnings from here

bakin1999:
## Overview

Matrix templates in php 8.1 complain about these if there is not information in them such as just a hanging empty matrix opener. Not big deal but others may run into same type of issue from not checking nulls first here.

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation


intoeetive (changes requested):
I think we should treat `null` as empty string (probably cast to string) and still perform replacement here.

Also the code is not PSR-12

PR #2053, opened 2022-04-27
Cast ft replace to a string since it should always be one

matthewjohns0n:
EECORE-1834

### DIFF
--- a/system/ee/legacy/libraries/channel_entries_parser/components/Custom_field.php
+++ b/system/ee/legacy/libraries/channel_entries_parser/components/Custom_field.php
@@ -105,7 +105,7 @@ class EE_Channel_custom_field_parser implements EE_Channel_parser_component
                     ));
 
                     if (method_exists($obj, $parse_fnc)) {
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/channel_entries_parser/components/Custom_field.php
```diff
@@ -104,13 +105,13 @@ public function replace($tagdata, EE_Channel_data_parser $obj, $ft_api)
                     ));
 
                     if (method_exists($obj, $parse_fnc)) {
-                        $entry = $ft_api->apply($parse_fnc, array(
+                        $entry = (string) ft_api->apply($parse_fnc, array(
```
bryannielsen: @matthewjohns0n looks like you lost the `$` on `$ft_api` here

matthewjohns0n: Thanks, fixed!

intoeetive (approved):

PR #2060, opened 2022-05-03
fix the issue where PHP notices were shown when using slider as CP component

intoeetive:
also fix PHP 8.1 notices

EECORE-1837


bryannielsen (approved):
This looks good and fixes the issue in Low Search

matthewjohns0n (approved):

PR #2075, opened 2022-05-10
Resolved #2063, where outdated resources could be served

### LINKED ISSUE #2063: CSS templates are synchronized only when visiting CP -> Templates
Summary: If I edit the style.css in the template folder ("/templates/default_site/mydirectory/de/style.css") remotely via FTP/SFTP, the changes are no longer transmitted/displayed after a (hard) reload .

When I load the "style.css" file from the template folder with my CSS editor "Nova" and make changes, the browser does not accept these changes. It doesn't matter whether I work on my provider's server or with my local web server (MAMP).

But if I open the CSS file in the EE backend with the template editor (only open it, don't even save it), then I see the changes and suddenly the browser also shows them (this applies to Chrome, Firefox, Safari and the caches I had deactivated, template caching is also deactivated).

Another note: I have upgraded to EE 6.3.2 for two customers and the problem occurs there too.
I changed the style.css via editor for a customer with EE 6.2.7. The changes were immediately visible online in the "page source code" of the browser. After that I updated this website to EE 6.3.2. After that I had the same problem: ExpressionEngine doesn't propagate the CSS changes to the browser. Even if I use a different editor, a different Mac computer or a different IP.

Now I have no choice but to reference the CSS file directly from another directory in the root (so not from "user/templates/etc"). Then everything works as usual again.

PHP 7.4
MySQL 5.7.37
EE 6.3.2 and 6.3.3
Editor: Nova from Panic

robsonsobral:
## Overview

- requests of resources with version suffixes, like `/index.php?css=resources/style.v.1652145060`, wasn't clearing the cache;
- resources on MSM sites where messing the checking of each other's modification / edition times;
- serve of fresh resources on cached pages wasn't being tested.

Resolves [#2063](https://github.com/ExpressionEngine/ExpressionEngine/issues/2063).

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [x] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/ExpressionEngine/Library/Resource/Request.php
+++ b/system/ee/ExpressionEngine/Library/Resource/Request.php
@@ -41,11 +41,13 @@ class Request
         $name = '';
         $site_name = '';
         $template_version = 0;
+        $modified_since = ee('Request')->header('IF_MODIFIED_SINCE');
 
+        // requests by trigger segments, the ones without version suffixes
         if (in_array(ee()->uri->segment(1), ee()->uri->reserved) && false !== ee()->uri->segment(2)) {
             $resource = ee()->uri->segment(2) . '/' . ee()->uri->segment(3);
             $this->type = ee()->uri->segment(1);
-        } else {
+        } else { // requests query strings, the ones with version suffixes
             foreach (self::TYPES as $type) {
                 if (ee('Request')->get($type)) {
                     $resource = ee('Request')->get($type);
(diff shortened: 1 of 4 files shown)


robsonsobral:
May we go ahead and merge this, @bytebaby?

bytebaby:
Yes. In my test environment, everything is now running correctly again, @robsonsobral

intoeetive (approved):

PR #2086, opened 2022-05-18
Improved modals UI

Yulyaswan:
Fix rounded corners on modals

EECORE-1611


ajohnson:
@Yulyaswan Looks good! I pushed a couple additional styling tweaks.

Using `overflow: hidden` definitely fixes the rounded corner issue. But it looks like we had previously removed that property because of some new features that use dropdown menus within modals.

So to avoid the dropdowns from getting clipped off I just added rounded corners to the `dialog__header` and `dialog__actions` instead.

Thanks!

ajohnson (approved):

PR #2089, opened 2022-05-19
fixed close button position for Template settings modal

Yulyaswan:
EECORE-1880


intoeetive (approved):

intoeetive (dismissed):
@Yulyaswan I don't think this was done right. The 'x' position is now incorrect on "select file" modal in entries.

Also since this is really minor UX change which is not influencing functionality, we could just skip it for the time being, if the solution would be too complex

bryannielsen (approved):

PR #2091, opened 2022-05-20
Updated return from grid model to get the correct array element to pr…

TomJaeger:
Updated the grid parser to greatly reduce queries and execution time when files are in a grid.

EECORE-1886


intoeetive:
the changes have been pushed directly to performance branch (@bryannielsen did review)

PR #2092, opened 2022-05-20
Feature/grid file performance

TomJaeger:
Big performance improvement for files in grid.


system/ee/legacy/libraries/Grid_parser.php
```diff
@@ -247,6 +247,10 @@ public function parse($channel_row, $field_id, $params, $tagdata, $content_type
 
         $columns = ee()->grid_model->get_columns_for_field($field_id, $content_type);
 
+        // since the get_columns_for_field was passed a single filed ID and NOT
+        // an array we need to make it an array as pre_loop expects an array
+        $this->_pre_loop(array($columns));
```
intoeetive, line 252: If I am reading the code right, `get_columns_for_fields` returns an array - no matter what you passed it. So here you looks like you are making it array of arrays?

Also, what's the reason we need `_pre_loop` to be added here?

intoeetive (commented):
Can you say a bit more on what this change does overall? 

Also see the other comment on code here.

And some of the changes are not PSR-12

system/ee/legacy/libraries/Grid_parser.php
TomJaeger: Pre loop, calls fieldtypes pre_loop function to allow them to fetch data once for the template parser loop... In this case, when a grid is in a fluid, it runs 1 query for the whole grid field.. prior to this it ran a DB query per grid row... (Specifically impacted performance with a file field in a grid).

The get_columns_for_field in this case returns a fairy flat array... however the _pre_loop function works on an array of arrays... which is why it's wrapped.

intoeetive:
the changes have been pushed directly to performance branch (@bryannielsen did review)

PR #2097, opened 2022-05-24
Added support for table alias in SQL joins when using Active Record, #2096

### LINKED ISSUE #2096: Add the ability to define join aliases in Active Record
Sometimes it is necessary to use an alias in a JOIN - when you want to make two separate table joins to the same column for example. Active Record does not support this, nor does it support the addition of aribtrary SQL to a query it is building, so currently you cannot (for example) use a lookup table multiple times in the same request. For example (from the [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/sql-statements.html#join) you might want to do this
```
SELECT t1.name, t2.salary
  FROM employee AS t1 INNER JOIN info AS t2 ON t1.name = t2.name;
```
Currently this is not possible to specify using Active Record' [join()](https://docs.expressionengine.com/latest/development/legacy/database/active-record.html#jointable-cond-type--) method.

It would be good to add the ability to specify an alias in the join() method. 
It would be easy enough to add to the function by defining a further optional parameter to those already in place, but it would be the necessary to disable the application of the `_protect_identifiers()` method to the join condition, since otherwise AR would attempt to (at the very least) prefix the alias prefix (similar to the effect of setting `$escape = false` in a select() statement).

Usually the need for aliased joins can be avoided by data table design (e.g. [avoiding the need for look-ups by duplicating data](https://dev.mysql.com/doc/refman/8.0/en/optimization.html#data-size-joins)), but if you are working with legacy data this might not be possible.

A small addition would be needed to the AR documentation to note the additional parameter, and that table-name protection is removed when an alias specified.

- [X] I am capable and would like to work on implementation of this feature if it is considered.

jcogs-design:
## Small modification to AR join() method to support use of join aliases

Adds optional fourth parameter to specify an alias for a join (no previous fourth parameter so backward compatible etc).
If an alias is specified it disables `_protect_identifiers` processing on condition in case alias is referenced in condition (identifier protection applied in that case would break the JOIN statement by prefixing the alias with the EE table prefix).

Resolves [#2096](https://github.com/ExpressionEngine/ExpressionEngine/issues/2096).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [X] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/458

### DIFF
--- a/system/ee/legacy/database/DB_active_rec.php
+++ b/system/ee/legacy/database/DB_active_rec.php
@@ -303,7 +303,7 @@ class CI_DB_active_record extends CI_DB_driver
         $this->_track_aliases($table);
 
         // Strip apart the condition and protect the identifiers
-        if (preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match) && !$alias) {
+        if (is_empty($alias) && preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match)) {
             $match[1] = $this->_protect_identifiers($match[1]);
             $match[3] = $this->_protect_identifiers($match[3]);
 
@@ -312,8 +312,8 @@ class CI_DB_active_record extends CI_DB_driver
 
         // If a join alias specified, extract it now
         $join_alias = '';
-        if ($alias) {
-            $join_alias = ' '.$alias.' ';
+        if (! is_empty($alias)) {
+            $join_alias = ' ' . $alias . ' ';
         }
 
         // Assemble the JOIN statement
--- a/system/ee/legacy/database/DB_active_rec.php
+++ b/system/ee/legacy/database/DB_active_rec.php
@@ -283,9 +283,10 @@ class CI_DB_active_record extends CI_DB_driver
      * @param	string $table The table to join
      * @param	string $cond The condition to join on
      * @param	string $type the type of join (left, right, outer, inner, left outer, right outer)
+     * @param	string $alias give the join an alias
      * @return	CI_DB_active_record The active record object
      */
-    public function join($table, $cond, $type = '')
+    public function join($table, $cond, $type = '', $alias = '')
     {
         if ($type != '') {
             $type = strtoupper(trim($type));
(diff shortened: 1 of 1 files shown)


system/ee/legacy/database/DB_active_rec.php
```diff
@@ -302,15 +303,21 @@ public function join($table, $cond, $type = '')
...
             $match[3] = $this->_protect_identifiers($match[3]);
 
             $cond = $match[1] . $match[2] . $match[3];
         }
 
+        // If a join alias specified, extract it now
+        $join_alias = '';
+        if ($alias) {
+            $join_alias = ' '.$alias.' ';
+        }
```
intoeetive: Can you add spaces around `.` to make the code PSR-12? Also, since `$alias` is never a boolean, it is probably more foolproof to check it with `is_empty` (thinking of future PHP versions here)

system/ee/legacy/database/DB_active_rec.php
```diff
@@ -302,15 +303,21 @@ public function join($table, $cond, $type = '')
         $this->_track_aliases($table);
 
         // Strip apart the condition and protect the identifiers
-        if (preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match)) {
+        if (preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match) && !$alias) {
```
intoeetive: It might be better to check `$alias` first (and maybe check with is_empty - see the comment below) to save a little bit by avoiding preg_match processing if it's not needed

intoeetive (changes requested):
I think this is really good addition. Thank you!

Just have couple of small comments - check those out

jcogs-design:
Have updated as requested.

intoeetive (approved):

intoeetive (changes requested):
@jcogs-design I am so sorry.

As @bryannielsen pointed out, `is_empty` is not valid PHP tag, `empty` should be used instead - can you make that update?

jcogs-design:
Have updated as requested.

intoeetive (approved):

intoeetive:
Thank you!

PR #2110, opened 2022-06-12
Feature/6.x/object addons

eric-lamb62:
EE6 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/2367

Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/503

(x.X.x) Small additive changes should target the next minor branch (release/next-minor if a numbered branch does not yet exist)

## Overview

This PR allows for Add-on developers to create their Add-on methods using an abstracted object layer while keeping backwards compatibility intact. It covers Extensions, Modules (Action and Tags), and the Control Panel layers. It's very simple to implement (just inherit an object), while maintaining the legacy implementation in play. 

To implement, you simply set your Add-on `mod.`, `ext.`, and/or `mcp` files to inherit from a base object (depending on the implementation).

## Add-on File Examples

Here's a breakdown for how the Add-on files would look for an Add-on named `Custom_addon`.

> Note that this works by using set or determined namespaces for a given Add-on. By default, this Controller implementation will use the Add-on name to get the configured namespace within the Add-on. For this example of `Custom_addon`, it's `Mithra62\PlayThing`; all route namespaces are determined by that. 

### Module

The `mod` file would consist exclusively of the below. 

```php
use ExpressionEngine\Service\Addon\Module;

class Custom_addon extends Module
{
    protected $addon_name = 'custom_addon';
}
```

Once that's done, your Add-on Module is now set to look for objects for all calls to it. Those objects depend on what "type" (Tag or Action). Using the configured namespace for the Add-on, the Module routes would be looked for in `Mithra62\PlayThing\Module\Tags|Action`. 

#### Template Tag 

For a template tag called as `{exp:custom_addon:test_template_tag}`, you'd setup your Tag object like the below:

```php
namespace Mithra62\PlayThing\Module\Tags;

use ExpressionEngine\Service\Addon\Controllers\Tag\AbstractRoute;

class TestTemplateTag extends AbstractRoute
{
    public function process()
    {

    }
}
```

#### Actions
Just like with Template Tags, Actions get stored in the `Actions` namespace. The below is for an Action called `my-test-action`. 

```php
<?php
namespace Mithra62\PlayThing\Module\Actions;

use ExpressionEngine\Service\Addon\Controllers\Action\AbstractRoute;

class MyTestAction extends AbstractRoute
{
    public function process()
    {

    }
}
```

### Extension

The `ext` is structured just like the others:

```php
use ExpressionEngine\Service\Addon\Extension;

class Custom_addon_ext extends Extension
{
    protected $addon_name = 'custom_addon';
}
```

Again, just like the others, Extensions are structured similarly.

```php
<?php
namespace Mithra62\PlayThing\Extensions;

use ExpressionEngine\Service\Addon\Controllers\Extension\AbstractRoute;

class TestExtension extends AbstractRoute
{
    public function process()
    {

    }
}
```

Note that the above `process` method allows you to accept variable parameters (just like regular Extensions). For example, to build an Extension around `template_post_parse`, you'd write your Extension like the below:

```php
namespace Mithra62\PlayThing\Extensions;

use ExpressionEngine\Service\Addon\Controllers\Extension\AbstractRoute;

class TestExtension extends AbstractRoute
{
    public function process(string $final_template, bool $is_partial, string $site_id, array $currentTemplateInfo): string
    {
        return $final_template;
    }
}
```

### Module Control Panel (MCP)

The `mcp` layer acts more as a Controller/Router than the others in that the CP is supposed to be rendered as Views. To start, just like all the others, you extend from the `Mcp` object. Once that's done, all requests will be routed to the object layer. 

```php
use ExpressionEngine\Service\Addon\Mcp;

class Custom_addon_mcp extends Mcp
{
    protected $addon_name = 'custom_addon';
}
```

#### Route Examples

A basic 'index' route (normally done through a method attached to the `mcp` object) would look like the below:

```php
namespace Mithra62\PlayThing\Mcp;

use ExpressionEngine\Service\Addon\Controllers\Mcp\AbstractRoute;

class Index extends AbstractRoute
{
    /**
     * @var string
     */
    protected $route_path = 'index';

    /**
     * @var string
     */
    protected $cp_page_title = 'home';

    /**
     * @param false $id
     * @return AbstractRoute
     */
    public function process($id = false): AbstractRoute
    {
        return $this;
    }
}
```

The generation of the CP array format is done automatically by the Controller so all the devs really do is assign variables and ensure their `process` method returns an instance of their Route:

```php
namespace Mithra62\PlayThing\Mcp;

use ExpressionEngine\Service\Addon\Controllers\Mcp\AbstractRoute;

class Index extends AbstractRoute
{
    /**
     * @var string
     */
    protected $route_path = 'index';

    /**
     * @var string
     */
    protected $cp_page_title = 'home';

    /**
     * @param false $id
     * @return AbstractRoute
     */
    public function process($id = false): AbstractRoute
    {
        $this->addBreadcrumb('test-link', 'test-link')
             ->addBreadcrumb('test-link2', 'test-link2');

        $variables = [];
        $this->setBody('my-view', $variables);
        return $this;
    }
}
```

## Nature of This Change

- [x] 🐛 Fixes a bug
- [x] 💅 Fixes coding style
- [x] ✅ Adds tests
- [x] 💥 Improves Maintainability
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine/pull/2110


eric-lamb62:
I do feel there's an argument to be made to switch the namespace(s) up to `\ExpressionEngine\Controllers\Addon\*` but with the `Installer` object being where it is, 'twas a little confusing on which was ideal.

system/ee/ExpressionEngine/Library/String/Str.php
```diff
@@ -0,0 +1,43 @@
...
+     * @param string $value
+     * @return string
+     */
+    public static function dash2ns(string $value): string
+    {
+        return self::path2ns(
+            str_replace(' ', '/',
+                str_replace('cartthrob', 'CartThrob', // ;)
+                    ucwords(str_replace('_', ' ', $value))
+                )
```
intoeetive: I think hardcoded `CartThob` probably went in here by mistake?

intoeetive (changes requested):
@mithra62 This looks really solid!

I see you are using type hinting for function return values, - unfortunately we can't take that into EE 6, because it needs to support PHP 5.6, and return type hinting is PHP7+ thing only.

Is that something you can update?

system/ee/ExpressionEngine/Library/String/Str.php
eric-lamb62: Yeah, can't provide special consideration to Add-ons; just won't scale lol

Will clean up.

eric-lamb62:
Thanks @intoeetive! I built this using the rules we had for the "Cp/Form" Libraries, so focused on 7.1 as the baseline. I'm sure it wouldn't be too difficult to remove the hinting though. Just to double check; you want me to downgrade this to work with PHP 5.6, right?

intoeetive:
@mithra62 yes, that is correct.

Or maybe you'll want to change the base for this be 7.dev and make a new one that's compatible with PHP 5.6 to go into 6.dev - that's up to you (unless you think this needs to go into EE7 only, we want to have this included both for 6 and 7 versions of CMS)

eric-lamb62:
Gotcha and agree; it'd be more helpful to have in both 6 and 7. Will clean it up this week(ish).

eric-lamb62:
@intoeetive I believe I hit them all but let me know if anything was missed.

intoeetive:
Looks like `system/ee/ExpressionEngine/Library/String/Str.php` wasn't updated

eric-lamb62:
my bad @intoeetive all pushed

intoeetive (dismissed):
Looks good to me! Approving

If you could also add some docs for this (even just based on the description of this PR), that would be awesome

intoeetive:
@mithra62 could you please also have this code PR'ed against `7.dev` in addition to this PR? I think you should be able to just do that from the same branch

eric-lamb62:
Done @intoeetive 

https://github.com/ExpressionEngine/ExpressionEngine/pull/2367

eric-lamb62:
Hey @intoeetive, 

Sorry for delay but I was able to create a PR for the docs for this new feature. 

https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/503 

Please lemme know if there's anything else required.

intoeetive:
Closing in favor of https://github.com/ExpressionEngine/ExpressionEngine/pull/2459 because we messed up here (

PR #2111, opened 2022-06-15
Resolved issue where PHP notice could be shown when calling Template library from an add-on

jcogs-design:
Line 1779 of `ee/legacy/libraries/Template.php` is passing null to `str_replace` under some conditions and assuming null will be replaced by ''.  php 8.1 doesn't like this and issues a deprecation warning. This uses the null coalescing operator to replace null with '' when the function is called if necessary.

## Overview

Replace this paragraph with a description of your changes and reasoning. Provide links to external references/discussions if appropriate.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [ ] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -1750,7 +1750,7 @@ class EE_Template
                 if ((strtolower($class_name) == strtolower($meth_name)) or ($meth_name == '__construct')) {
                     $return_data = (isset($EE->return_data)) ? $EE->return_data : '';
                 } else {
-                    $return_data = $EE->$meth_name();
+                    $return_data = (string) $EE->$meth_name();
                 }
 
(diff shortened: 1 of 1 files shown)


intoeetive (changes requested):
Unfortunately `??` operator is not available in PHP 5.6 (and EE6 still supports that)

How about typecasting on line 1793, `$return_data = (string) $EE->$meth_name();` - would that solve the issue? (I'm assuming `null` in this case is coming from an add-on?)

jcogs-design:
Have updated patch to use suggested alternative php5.6 compatible solution.

intoeetive (approved):

bryannielsen (approved):

PR #2137, opened 2022-07-12
Resolved issue where some logins could trigger an erroneous cookie domain error

robinsowell:
Fixed a bug where some logins could trigger an erroneous 'The configured cookie domain does not match the site URL' error.


intoeetive:
(this is already in EE7, but needed for EE6)

intoeetive (approved):

Issue #2145, opened 2022-07-21
Unexplained errors with fluid grids & file fields, can't save entries


murtaugh:
I've been getting an unspecified error when submitting entries, and I think I've nailed the problem down to an issue with fluid image grids.

1. Create a new entry.
2. In a fluid fieldtype, insert either a file grid field or a grid field with a file field in it. Make sure there's at least one row of content.
3. Add another grid field. As far as I can tell the content of this second grid field doesn't matter. Make sure there's at least one row of content.
4. The entry will save normally.
5. Return to the entry and add a row to the first grid. Trying to save the entry will now result in an error.

<img width="578" alt="image" src="https://user-images.githubusercontent.com/308752/180301827-886f1e8d-498c-42b3-911b-d339f6ab2feb.png">

There are no indications of what the error is within the fluid field. At the bottom of the fluid field is this:

<img width="502" alt="image" src="https://user-images.githubusercontent.com/308752/180301952-53ddc0af-d139-4a32-b2c1-812589d8fe5d.png">

I've noticed that if the second grid field has radio buttons, the radio buttons have a red border around them:

<img width="591" alt="image" src="https://user-images.githubusercontent.com/308752/180302580-96563418-8381-467b-94a8-ac596ab023b7.png">

Interestingly, as I'm testing now, this field *appeared* with the red border around it. I never submitted the entry with this field in the fluid fieldtype, so I don't know why it has the error applied to it.

If you re-order the fields, so that the grid field with the file in it is no longer is position 1, the entry will save without an error, but the grid field that is *now* in position 1 *will* start throwing errors when you add a row, if it has a file field in it.

EE 6.2.4
PHP 8.0.8

jejuna:
Yeah, me too. I reported this June 1, here:
[https://github.com/ExpressionEngine/ExpressionEngine/issues/2105](url)

It seems to be some odd validation bug. At least I'm not the only one who's got this issue, as it's driving my client insane.

matthewjohns0n:
Just wanted to give a little update on this one. We're making progress on this. I've figured out part of this issue, but not exactly why it's happening yet. We'll get this fixed soon though!

matthewjohns0n:
@murtaugh @jejuna I believe this PR fixes the grid validation issue you're running into. We should be getting this fix into an upcoming release. If you're able to, it would be great to verify this fixes the issues you're running into.

jejuna:
Thanks, I will report back!

[...]

PR #2149, opened 2022-07-25
Fixes bug with grid validation error

### LINKED ISSUE #2105: Updating grid field with radio field. Nope.
We have several grid fields used in a fluid field. This site has been working perfectly for nearly 2 years.
Recently upgraded to EE 6.3.4. PHP is 8.02. 

Now, anytime the client adds a row to an existing entry that has a radio fieldtype in a grid field, it won't save. 

I thought it might be a corrupted entry, so I made a new one, and still, same error. 

I made a new test grid field with a radio button in it, and still, same error.

It will save the first time with no issues. But adding a row to an existing grid field makes it bork.

I thought it might be an issue with the new conditional field, but I've checked, and none are marked conditional. 

Here's the screenshot.
screenshot

It's not just this field. It's every field that has a radio fieldtype.

screenshot2

This seems like a bug.

matthewjohns0n:
Fixes #2105 #2145

### DIFF
--- a/system/ee/ExpressionEngine/Addons/grid/ft.grid.php
+++ b/system/ee/ExpressionEngine/Addons/grid/ft.grid.php
@@ -820,6 +820,11 @@ class Grid_ft extends EE_Fieldtype
         $entry_id = (isset($this->settings['entry_id']))
             ? $this->settings['entry_id'] : ee()->input->get_post('entry_id');
 
+        // If entry id isnt set and this is an edit entry, we'll get it from the URL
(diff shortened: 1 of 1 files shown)


intoeetive (commented):
The potential problem with this solution is that it will not work with channel forms, or wherever the URL structure is different.

But maybe it's not an issue there

intoeetive (changes requested):
This is causing issues with Fluid. 
We'll need to work on better solution.
Closing for now

PR #2192, opened 2022-08-03
Revert "fixed the sortable behaviour for relationship field"

matthewjohns0n:
Reverts ExpressionEngine/ExpressionEngine#2179


matthewjohns0n:
This one was passing tests, so we are not reverting it.

PR #2204, opened 2022-08-05
fix the File Usage utility to correctly update Fluid fields

intoeetive:
fix to keep the total_usage correct on subsequent runs

EECORE-2072


robinsowell:
OK, I had not confirmed the bug before testing the fix, so I won't swear I took the proper steps to see it.

On 7.0.1, took a screen cap of current files page, including the usage stats.  Ran the file usage utility.  Numbers the same.  Ran it again, numbers still the same.  Re-running it doesn't seem to matter.  I also looked at an individual entry and the usage count for it stayed the same.

Screen Shot 2022-08-05 at 10 33 41 AM

matthewjohns0n (approved):

PR #2219, opened 2022-08-08
Resolved #2157 where Color Picker field was cut off when used inside Grid field

### LINKED ISSUE #2157: Color picker field panel, hidden inside of grid field

Yulyaswan:
the same code as here https://github.com/ExpressionEngine/ExpressionEngine/pull/1507
but for EE version 7
fixes #2157 
EECORE-2085

### DIFF
--- a/cp-styles/app/styles/components/_fieldsets.scss
+++ b/cp-styles/app/styles/components/_fieldsets.scss
@@ -53,6 +53,7 @@ fieldset {
 
     // margin-bottom: 25px;
 	border: 0;
+	min-width: 0;
 
 	&.last {
 		margin-bottom: 0;;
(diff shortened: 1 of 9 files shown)


intoeetive:
@Yulyaswan It looks like this PR is changing more than is described in referenced GitHub issue - I also see changes in validation. Can you double check and make sure these are the changes that we need here?

Yulyaswan:
@intoeetive Changes from #2157 won't work without changes in #1507. That's why I created this branch for v7 with changes regarding grid and then fixed color picker problem

intoeetive (approved):

PR #2240, opened 2022-08-09
changed sortable to draggable for grid view for file manager

Yulyaswan:
EECORE-2062

### DIFF
--- a/system/ee/ExpressionEngine/View/_shared/thumb.php
+++ b/system/ee/ExpressionEngine/View/_shared/thumb.php
@@ -58,7 +58,7 @@ if (count($data)) {
 
                 <?php if ($i == 1) : ?>
                     <div class="file-thumbnail__wrapper">
-                        <a<?php foreach ($row['attrs'] as $key => $value) :?> <?=$key?>="<?=$value?>"<?php endforeach; ?> data-file-id="<?=$row['attrs']['file_id']?>">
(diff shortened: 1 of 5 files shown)


system/ee/ExpressionEngine/View/_shared/thumb.php
```diff
@@ -58,7 +58,7 @@
 
                 <?php if ($i == 1) : ?>
                     <div class="file-thumbnail__wrapper">
-                        <a<?php foreach ($row['attrs'] as $key => $value) :?> <?=$key?>="<?=$value?>"<?php endforeach; ?> data-file-id="<?=$row['attrs']['file_id']?>">
```
intoeetive, line 61: @Yulyaswan is there specific reason all extra attributes are being removed?

Yulyaswan: @intoeetive  before these changes, the main wrapper for each element was `<a>` with all attributes, then I changed the wrapper to a `<div>` to fix the issue of `"Checkbox in thumbnail view opens an image when it shouldn't" `and put the `<a>` inside a block.
In this case, all attributes are duplicated, which can be a problem in the future. That's why I left all attributes for the main wrap and only the `href` for the link.

intoeetive (approved):

PR #2242, opened 2022-08-10
Resolved #2105 where adding new row for File Grid inside Fluid could cause validation error in some cases

### LINKED ISSUE #2105: Updating grid field with radio field. Nope.

### LINKED ISSUE #2145: Unexplained errors with fluid grids & file fields, can't save entries

intoeetive:
…cause validation error in some cases

Fixes #2145

EECORE-1899

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -79,7 +79,7 @@ class File_ft extends EE_Fieldtype implements ColumnInterface
                 // Is this an edit?
                 if ($this->content_id) {
                     // Are we validating on grid data?
-                    if (isset($this->settings['grid_row_id'])) {
+                    if (isset($this->settings['grid_row_id']) || isset($this->settings['grid_row_name'])) {
                         $fluid_field_data_id = (isset($this->settings['fluid_field_data_id'])) ? $this->settings['fluid_field_data_id'] : 0;
 
                         ee()->load->model('grid_model');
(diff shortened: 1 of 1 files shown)


robinsowell (approved):

robinsowell:
Tests out for me in a clean install of 6.3.5.

PR #2247, opened 2022-08-10
Added support for table alias in SQL joins when using Active Record, #2096

intoeetive:
EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/2097

### DIFF
--- a/system/ee/legacy/database/DB_active_rec.php
+++ b/system/ee/legacy/database/DB_active_rec.php
@@ -303,7 +303,7 @@ class CI_DB_active_record extends CI_DB_driver
         $this->_track_aliases($table);
 
         // Strip apart the condition and protect the identifiers
-        if (preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match) && !$alias) {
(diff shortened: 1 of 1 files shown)


system/ee/legacy/database/DB_active_rec.php
```diff
@@ -302,15 +303,21 @@ public function join($table, $cond, $type = '')
         $this->_track_aliases($table);
 
         // Strip apart the condition and protect the identifiers
-        if (preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match)) {
+        if (is_empty($alias) && preg_match('/([\w\.]+)([\W\s]+)(.+)/', $cond, $match)) {
```
bryannielsen: I think this line is causing the tests to fail.  This probably just needs to change from `is_empty($alias)` to `empty($alias)`

intoeetive: @bryannielsen updated

bryannielsen (approved):

PR #2267, opened 2022-08-15
Add jquery-migrate-3.0.0 library, and include it in the control panel footer - EECORE-2092

matthewjohns0n:
Fixes EECORE-2092


matthewjohns0n:
@Yulyaswan Let me know what you think of including this. I'm not sure if there is a better way to fix this, but this seems to solve my specific issue.

Yulyaswan:
I checked and didn't find a reason to don't use this file.

Yulyaswan (approved):

PR #2275, opened 2022-08-16
Resolved #2104 where opening link in new tab could falsely alert about leaving the page

### LINKED ISSUE #2104: "Data will be lost warning" triggered on command-click
The "data will be lost" warning alert triggers when you command click on a link, meaning the page you're on isn't going to be navigated away from, and no data is going to be lost.

image

Mac OS, Safari 15.5

Yulyaswan:
EECORE-2083
fixes #2104

### DIFF
--- a/themes/ee/asset/javascript/src/cp/publish/publish.js
+++ b/themes/ee/asset/javascript/src/cp/publish/publish.js
@@ -62,6 +62,17 @@ $(document).ready(function () {
 		});
 	}
 
+	// check if command is clicked
+	var cmdPressed = false;
+	$(window).keydown(function(evt) {
+		if (evt.which == 91) { // command/Windows
+			cmdPressed = true;
+		}
+	}).keyup(function(evt) {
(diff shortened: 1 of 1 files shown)


intoeetive (changes requested):
I think we need a piece of code which would reset `cmdPressed` back to `false` after click / button press is made

Yulyaswan:
it's a reset variable using the code below. As soon as the user releases the Cmd (Window) button, the `cmdPressed = false;`
```
.on('keyup', function(evt) {
		if (evt.which == 91 || evt.which == 224) { // command/Windows
			cmdPressed = false;
		}
	});
```

intoeetive (changes requested):
@Yulyaswan this does not work in Windows.

I think this is because `ctrl` key code is `17` and not `91`

https://www.toptal.com/developers/keycode

intoeetive:
@Yulyaswan windows+click is not a thing (works same as click). It should not be in code

intoeetive (approved):

PR #2276, opened 2022-08-16
Resolved #2104 where opening link in new tab could falsely alert about leaving the page

### LINKED ISSUE #2104: "Data will be lost warning" triggered on command-click

Yulyaswan:
EECORE-2083
fixes #2104


intoeetive (changes requested):
see https://github.com/ExpressionEngine/ExpressionEngine/pull/2275

intoeetive (changes requested):
@Yulyaswan looks like it needs same changes that are in https://github.com/ExpressionEngine/ExpressionEngine/pull/2275

intoeetive (approved):

PR #2290, opened 2022-08-18
Resolved #2131, #913 where moving fields between tabs in Channel Layouts did not work well

### LINKED ISSUE #913: Field landing target fails to appear in empty tab when dragging field in channel publish layout screens (CP)
If you create a new tab in a publish layouts screen in CP, when dragging an existing field from another tab to that newly created tab, when view switches to the new tab (drag field, then hover over tab label), no 'landing target' is shown for the new field within the tab.
The absence of a landing target makes it tricky to add the first field to an otherwise empty tab: if you don't hit the right part of the view (somewhere near the tab itself) the drag partially fails - the dragged field ends up at the top of the tab you were dragging it from.  
Once one field is in place, the landing target appears as normal.

### LINKED ISSUE #2131: Saving New Channel Layout rearranges fields on Publish tab and newly created tab
_EE 6.3.4, PHP Version 7.3.23, Google Chrome Version 103.0.5060.114_

When logged into EE and creating a new Channel Layout, I add a new tab to the layout setup. Drag and drop fields (textareas and text inputs) from the Publish tab over to the newly created tab. When I click Save for the New Publish Layout, I receive the confirmation/success that it's been created, but the ordering of the fields on the Publish tab and the newly created tab are completely different from where they were actually placed; some fields that were on the new tab end up on the Publish tab, and vice versa. Doesn't seem to be any correlation of which fields get moved to the new tab, either.

Screen Shot 2022-07-07 at 11 12 26 AM
Screen Shot 2022-07-07 at 11 12 43 AM

Yulyaswan:
EE6 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/2317

EECORE-2094
fixes #2131 
fixes #913

### DIFF
--- a/cp-styles/app/styles/components/_layout-item.scss
+++ b/cp-styles/app/styles/components/_layout-item.scss
@@ -1,5 +1,6 @@
 
 .layout-item-wrapper {
+	min-height: 20px;
 	.layout-item {
 		margin-bottom: $s-s;
 	}
--- a/cp-styles/app/styles/legacy/_tabs.scss
+++ b/cp-styles/app/styles/legacy/_tabs.scss
@@ -136,6 +136,17 @@
 				};
 		}
 
+		.js-layout-item { margin-bottom: 10px; }
+		.custom-drag-placeholder {
+			.none {
+				background-color: color(bg-0);
+				border: 1px dashed color(border-dark);
+				padding: 10px;
+				border-radius: $border-radius-base;
+				margin-bottom: 10px;
+			}
+		}
+
 		@include clear-float;
 
 		// @include m-tablet-up {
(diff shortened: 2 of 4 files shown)


robinsowell (dismissed):
This looks like it fixes the ability to move a cat group into an empty tab and it keeps the order.  We do still have an issue that after saving, the category group has a weird 'title':
Screen Shot 2022-08-18 at 10 04 04 AM

I think that's because category groups don't HAVE a shortname to show, there should just be the name of the category group or maybe Group Name (category group) so it's clear what it is if you move it around.

But I think that's a seperate issue and quite possibly a php one.  And this fixes the reported problem.

intoeetive:
@Yulyaswan can we have same PR for 7.x please

Yulyaswan:
@intoeetive sure, will do. Still working on this one

intoeetive (dismissed):

intoeetive (approved):
All these changes are already in EE7. Approving

PR #2291, opened 2022-08-18
Update default filesystem to use root of system path

bryannielsen:
## Overview

Update the path used by default filesystem instance to be the root of SYSPATH.  This is more in line with what previous versions of ExpressionEngine allowed and should resolve issues where an installation is spread out across a volume.

Resolves EECORE-2093

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

### DIFF
--- a/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
+++ b/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
@@ -983,6 +983,10 @@ class Filesystem
     public function createTempFile()
     {
         $file = tmpfile();
+        if($file === false) {
+            throw new \Exception('Cannot create temp file at "'.sys_get_temp_dir().'"');
+        }
+        
         $path = stream_get_meta_data($file)['uri'];
 
         return compact('file', 'path');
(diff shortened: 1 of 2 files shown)


intoeetive (dismissed):
I have tested this with file manager, templates and 1-click update using various system folder locations, and it seems to work fine on Windows.

Approving, but I think we need more testing and approving here.

intoeetive:
Looks like this does not solve some of possibly related issues: https://github.com/ExpressionEngine/ExpressionEngine/issues/2303#issuecomment-1222199606

Yulyaswan:
I've tested file manager: create a new directory, add folders, move files between folders and directories, edit, delete, copy link, drag and drop files, and add a new file using the `upload new` button. Everything seems to be working.

Yulyaswan (dismissed):

intoeetive (approved):

matthewjohns0n (approved):
I tested this on my site that is set up with a separate private path and this is now working.

Issue #1993, opened 2022-04-11
Member Login Ajax Validation


eric-lamb62:
**Is your feature request related to a problem? Please describe.**
When users log in via the `member:login_form` template tag and `member_login` Action, it's not possible to do inline error messages to the user. This isn't ideal. 

Digging into the code makes it obvious this isn't something that can "just be done"; there are lots of peccadillos that have to be taken into account due to the logic chain spread over multiple objects and methods (and globals). This, too, is not ideal, but that's a different thing. 

**Describe the solution you'd like**
A nice happy medium could involve a simple Ajax/Json endpoint Action that would allow front end developers to validate the login credentials before submitting the form. There's already the handy dandy `ee()->auth->verify()` method that could be wrapped in an Action method to return JSON. 

I've already implemented something like this for a client:

```php
public function member_login_ajax_validate()
{
    ee()->load->library('Auth');
    $verify = ee()->auth->verify();
    $reasons = [];
    if(!$verify) {
        $reasons = ee()->auth->errors;
    }

    $email = '';
    echo json_encode([
        'result' => ($verify ? 'success' : 'fail'),
        'reason' => $reasons,
        'username' => ee()->input->get('username'),
    ]);
}
```

**Describe alternatives you've considered**
Telling my client they can't have inline error messages on their login form but then I remembered it's 2022 and I'd look silly. 

**Teachability, Documentation, Adoption, Migration Strategy**
Yeah yeah yeah, I'll write the docs for this (if it's accepted)

- [x] I am capable and would like to work on implementation of this feature if it is considered.

intoeetive:
Yeah this looks like really nice to have.

I wonder if the best approach is to submit extra field which would trigger "validation only" mode (like this is done with `ee_fv_field` in CP).

But - if you already have the code that is working, maybe you could just submit the PR, and that would be it.

eric-lamb62:
Oh nice; tbh, I don't follow you on the `ee_fv_field` but I can put a PR together for this, no problem.

intoeetive:
If you blur the entry field (or almost any field) in CP, you'll notice the ajax request going out; it submits all the same data and goes to same endpoint, but also adds that `ee_fv_field` variable with the name of the field that's currently validated. That tells the backend that this is validation request, and entry should not be saved.

Here's random example of this in action
```
        if (AJAX_REQUEST) {
            ee()->form_validation->run_ajax();
            exit;
[...]

PR #2324, opened 2022-08-23
Adds Ajax Login Validation

eric-lamb62:
## Overview

Adds an Action to validate a username and password pair, ideally within an Ajax request prior to login form submission. 

Resolves [#1993](https://github.com/ExpressionEngine/ExpressionEngine/issues/1993).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member.php
@@ -1614,9 +1614,9 @@ class Member
             $reasons = ee()->auth->errors;
         }
 
-        echo json_encode([
-            'result' => ($verify ? 'success' : 'fail'),
-            'reason' => $reasons,
+        ee()->output->send_ajax_response([
+            'messageType' => ($verify ? 'success' : 'error'),
+            'messages' => $reasons,
             'username' => ee()->input->get('username'),
         ]);
     }
--- a/system/ee/ExpressionEngine/Addons/member/mod.member.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member.php
@@ -1602,6 +1602,26 @@ class Member
     }
 
     /**
+     * Allows for validation of username and password using Ajax
+     * @return void
+     */
+    public function login_validate()
+    {
+        ee()->load->library('Auth');
+        $verify = ee()->auth->verify();
+        $reasons = [];
+        if(!$verify) {
+            $reasons = ee()->auth->errors;
+        }
+
+        echo json_encode([
+            'result' => ($verify ? 'success' : 'fail'),
+            'reason' => $reasons,
+            'username' => ee()->input->get('username'),
+        ]);
+    }
+
+    /**
      * Username/password update
      */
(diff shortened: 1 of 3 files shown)


eric-lamb62:
@intoeetive we'll still need to update the Installer to create the `exp_actions` row but I'm not sure how best to handle that tbh. Never played with the Installer lol

system/ee/ExpressionEngine/Addons/member/mod.member.php
```diff
@@ -1601,6 +1601,26 @@ public function login_form()
...
+        $reasons = [];
+        if(!$verify) {
+            $reasons = ee()->auth->errors;
+        }
+
+        echo json_encode([
+            'result' => ($verify ? 'success' : 'fail'),
+            'reason' => $reasons,
+            'username' => ee()->input->get('username'),
+        ]);
```
intoeetive: It is better to use `ee()->output->send_ajax_response` as it sends the correct header and assures there's no profiler present in AJAX response, if it's enabled

Also not real issue, but `result => fail` is not being used in EE, more common are things `'messageType' => 'error',` or `'status' => 'error',`. As well as `reason` - could it be `error`? Again, this will not create any issues, just trying to keep it more consistent with what we have in other places - let me know if you have specific reason to keep them this way.

I will also make the PR draft until the action record part and the docs are ready

eric-lamb62: Nope, no reason whatsoever for it; was just a carry over from previous implementations. What you think of something like this:

```php
        ee()->output->send_ajax_response([
            'messageType' => ($verify ? 'success' : 'error'),
            'errors' => $reasons,
            'username' => ee()->input->get('username'),
        ]);
```

eric-lamb62: ... maybe `messages` in lieu of `errors`...

eric-lamb62:
I think I found how to do the updater stuff; shoulda realized it was in the Add-on proper. @intoeetive can you poke at the upd.member change? I think I did it right but confidence isn't as high as I'd like.

system/ee/ExpressionEngine/Addons/member/upd.member.php
```diff
@@ -80,6 +83,11 @@ public function update($current = '')
             ee()->db->update('actions', ['method' => 'do_member_search']);
         }
 
+        if (version_compare($current, '6.3.4', '<')) {
```
intoeetive: In case of add-ons updater, `$current` represents the add-on version, not EE version. So you would need to update version in `addon.setup.php` and use corresponding version in this check

eric-lamb62: oh, good to know; updated and pushed anew :)

PR #2332, opened 2022-08-25
Resolved #2331 where saving template revision was causing SQL error when using long table prefix

### LINKED ISSUE #2331: item_table field in revision_tracker table is not long enough for table names
We are using ExpressionEngine 6.3.5. PHP 7.4.30, MySQL 10.4.25-MariaDB

When turning on template revision tracking in template preferences and trying to save a template, we got the attached error. I looked at the database, and the item_table field in the database looks like it's set to 20 characters, but our table name is 31 characters long.

Let me know if you need any more info. 
Screen Shot 2022-08-23 at 9 42 43 AM

intoeetive:
fixes #2331

### DIFF
--- a/system/ee/installer/updates/ud_7_01_03.php
+++ b/system/ee/installer/updates/ud_7_01_03.php
@@ -64,7 +64,6 @@ class Updater
             ]
         );
     }
-    }
 }
 
 // EOF
--- a/system/ee/installer/updates/ud_7_01_03.php
+++ b/system/ee/installer/updates/ud_7_01_03.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * This source file is part of the open source project
+ * ExpressionEngine (https://expressionengine.com)
+ *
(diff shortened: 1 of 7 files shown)


bryannielsen (approved):

system/ee/installer/updates/ud_7_01_03.php
```diff
@@ -0,0 +1,70 @@
...
+                'item_field' => [
+                    'name' => 'item_field',
+                    'type' => 'varchar',
+                    'constraint' => 32,
+                    'null' => false
+                ]
+            ]
+        );
+    }
+    }
```
bryannielsen: It looks like this extra closing brace is responsible for the failing tests - https://github.com/ExpressionEngine/ExpressionEngine/runs/8009758506?check_suite_focus=true

bryannielsen (dismissed):

matthewjohns0n (approved):

PR #2334, opened 2022-08-25
Resolved issue where Color Picker swatches were not properly working inside grid

Yulyaswan:
EECORE-2113


intoeetive (approved):
approving as the change exists and works in EE7

PR #2344, opened 2022-09-01
Resolved #1651 where dropdowns in File field were showing the directories that member was not allowed to access

### LINKED ISSUE #1651: Users can see upload destinations they don't have access to and get an inaccurate error when uploading
I have an upload destination only admins can upload to. When a non-admin uses a field that allows uploads to any destination, they can see the forbidden destination in the dropdown within the dropzone. (The dropdown associated with the "upload new" button correctly hides it.)

When they try to upload a file to the forbidden destination, they get the following message:

`Unexpected error. Check your post_max_size setting in your PHP configuration.`

intoeetive:
Resolved #1651 where dropdowns in File field were showing the directories that member was not allowed to access

Testing instructions:
1.
- have a non-superadmin role ("editors") that can access entries and files
- have a file fields that can use all upload directories
- create several upload destinations, allow your role to access few of those, but not all
- login as editor
- the dropdowns in file field should only show the directories editor has access to
- uploading and browsing works

2.
- edit the role so thy can access only one directory
- login as editor
- the file field should not have dropdowns; the only allowed directory is preselected
- uploading and browsing works

3. Test # 2 with the subfolder - there should be dropdown with the only allowed directory and it's subfolderd

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -142,7 +142,7 @@ class File_ft extends EE_Fieldtype implements ColumnInterface
     {
         $allowed_file_dirs = (isset($this->settings['allowed_directories']) && $this->settings['allowed_directories'] != 'all')
             ? $this->settings['allowed_directories']
-            : '';
+            : 'all';
         $content_type = (isset($this->settings['field_content_type'])) ? $this->settings['field_content_type'] : 'all';
         $existing_limit = (isset($this->settings['num_existing'])) ? $this->settings['num_existing'] : 0;
         $show_existing = (isset($this->settings['show_existing'])) ? $this->settings['show_existing'] : 'n';
--- a/system/ee/ExpressionEngine/Addons/file/views/publish.php
+++ b/system/ee/ExpressionEngine/Addons/file/views/publish.php
@@ -38,6 +38,7 @@
 <?php
 $component = [
     'allowedDirectory' => $allowed_directory,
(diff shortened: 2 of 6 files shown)


robinsowell:
I'm having some issues here with subdirectories.  I THINK but won't swear they are related, because as superadmin, everything worked fine.

As non-superadmin with access to 2 file upload directories, I get errors when a directory has a subfolder.
Screen Shot 2022-09-08 at 3 07 26 PM

 So basically, click 'choose existing' it properly gives me the 2 directories (but not the subdirectory), I choose the directory with a sub in it and I got the above error.  If it's hard to see in image:

```
Warning
Undefined array key "href"
ee/ExpressionEngine/View/_shared/thumb.php, line 61
```

And as a user who only has access to a single directory that has a subfolder in it:

File field
- click 'choose existing' and I get a blank modal opening.
- click 'upload new' opens ok, I pick an image, it appears to upload it fine, but when it shows it, I get a 'you are not authorized': 
Screen Shot 2022-09-08 at 3 14 19 PM
- click 'choose directory' I get my 1 directory and the sub, pick the sub, upload image, appears to work but then I get that 'you are not authorized' after upload.

intoeetive:
@robinsowell should be fixed now - can you check again?

As of the permission error, this must be the edge case whether the member can access directory A only, but the field is set to directory B only. The UI is currently does not respect that - if it needs to, we'd need a separate task, but this seems like real edge case. But the backend is fully respecting the restrictions, not allowing you to upload or select a file.

robinsowell (approved):
approving, dropdowns show correctly, but I do still have that weird error where they can upload to a directory they should be able to upload to- but get a permission error and can't view the file.

PR #2345, opened 2022-09-01
Enabled entering HTML line break tags into Entry Title fiels; #1633

### LINKED ISSUE #1633: Entry titles (in the control panel) should support the <wbr> tag
## Description
When using a longer word in a title I would like to inform the browser of a possible breaking point use a `<wbr>` tag.
However, when I use the `<wbr>` tag I get an error message:

`This field can only contain the following HTML tags: b, cite, code, del, em, i, ins, markspan, strong, sub, sup. If you want to use angle brackets < in your text, but not HTML please try &lt; to replace < and &gt; to replace >.`

## Solution
I would like the title field to accept the  `<wbr>` tag.

## Example
`Supercalifragilistic<wbr>expialidocious`

## Alternatives
Don't use the title field - use a custom field instead.
Don't use the control panel - use a custom build entry form instead.

## What about...
Using the title of an entry is quite excellent in the `<title>Supercalifragilistic<wbr>expialidocious</title>`
But we wouldn't want those <wbr> tags visible in the page title. This means the site creator needs to add extra code to remove it. Either using an addon ([striptags](https://github.com/nathanvanbakel/expression-engine-striptags)) or by using the build in php functionality.

- [ ] I am capable and would like to work on implementation of this feature if it is considered.

intoeetive:
Closes #1633

(pre-approved in GH issue by Andy J)

### DIFF
--- a/system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
+++ b/system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
@@ -147,7 +147,7 @@ class ChannelEntry extends ContentModel
         'author_id' => 'required|isNatural|validateAuthorId',
         'channel_id' => 'required|validateMaxEntries',
         'ip_address' => 'ip_address',
-        'title' => 'required|maxLength[200]|limitHtml[b,cite,code,del,em,i,ins,markspan,strong,sub,sup]',
+        'title' => 'required|maxLength[200]|limitHtml[b,cite,code,del,em,i,ins,markspan,strong,sub,sup,br,wbr]',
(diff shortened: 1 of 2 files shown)


Yulyaswan:
It doesn't work for me.
Screenshot 2022-09-23 at 15 35 55
Screenshot 2022-09-23 at 15 36 05

Tag <br> make space but doesn't inserts a single line break

intoeetive:
@Yulyaswan this should work now, can you please test

because this is code in Template parser, I'd like to get code review from @bryannielsen or @matthewjohns0n as well

Yulyaswan (approved):

PR #2373, opened 2022-09-06
Resolved #2366 where is was not possible to use title in `search` parameter of `exp:channel:entries` tag

### LINKED ISSUE #2366: search:title="=xx" does not seem to work in EE7
When using the search parameter in an {exp:Channel:entries} tag, I expected it to be able to search the "title" field, what I got instead were inaccurate results.

**How To Reproduce**
1. Create a channel with a few fields.
2. in the title, use simple text like "AAS01, HE01, BP01, AS01" and then varying data in the other fields
3. Create a channel Entries tag like so: {exp:channel:entries channel="[channelname]" search:title="=HE01"}
4. See if the returned results are what you expect.

No Error Messages, just incorrect results

**Environment Details:**
 - Version: [e.g. 7.1.2]
 - PHP Version [e.g. 8.1.1]
 - MySQL Version [e.g. 5.7]
 - OS: [macOS]
 - Web Server: [e.g. Apache on MAMP]

**Possible Solution**
I really don't know... it jus tseems to not be behaving correctly.
**Additional context**
Only other thing I can think now, is that I had changed the Title Label to “Customer Number” in the Channel setup,  but I changed it back when I started having issues.

intoeetive:
fixes #2366

EECORE-2127

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
+++ b/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
@@ -646,7 +646,10 @@ class Channel
                 // We're goign to repeat the search on each site
                 // so store the terms in a temp.  FIXME Necessary?
                 $terms = $search_terms;
-                if (! isset($this->cfields[$site_id][$field_name])) {
+                if (in_array($field_name, ['title', 'url_title'])) {
+                    $table = 't';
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
Tested out fine for me doing both an exact and contains search.  Yay!

PR #2376, opened 2022-09-07
Resolved #2375 where Profiler section in CP had some English words hardcoded

### LINKED ISSUE #2375: Hardcoded English artifacts in the CP
In Performance.php, line 56, the word "of" is hardcoded instead of using `lang('of')`. This only becomes apparent when using a separate language pack.

Also, the words "Total Execution Time" seem to be hardcoded. I have not been able to find a corresponding language string to translate. I am running EE 7.1.12 on a local staging server.

igreil:
## Overview

In system/ee/ExpressionEngine/Service/Profiler/Section/Performance.php, line 56, the word "of" is hardcoded. You'll only notice when using a language pack, where it should be translated to, say, "de" in French or "von" in German, but is not.

fixes #2375

## Nature of This Change

Simply use `lang('of')` which uses the correct string from the language files.

- [X] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security

- [X] Yes
- [ ] No

### DIFF
--- a/system/ee/ExpressionEngine/Service/Profiler/Section/Performance.php
+++ b/system/ee/ExpressionEngine/Service/Profiler/Section/Performance.php
@@ -53,7 +53,7 @@ class Performance extends ProfilerSection
 
         $data = array();
         if (function_exists('memory_get_usage') && ($usage = memory_get_usage()) != '') {
-            $data[lang('profiler_memory')] = $this->fmt_factory->make('Number', $usage)->bytes() . ' of ' . ini_get('memory_limit');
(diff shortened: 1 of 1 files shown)


intoeetive (changes requested):
@igreil just realized you missed an extra space that's needed before 'of' - can you add that?

igreil:
I'm not sure it's actually needed (it worked fine for me without), but sure: I added a space before "of".

intoeetive (approved):

PR #2380, opened 2022-09-08
Added ability to edit related entries content from relationship field; #656

### LINKED ISSUE #656: Edit relationship field entries directly from a channel entry
To help improve workflow, for a relationship field, It would be useful to see a option to edit a relationship entry directly from the channel entry you are linking from.

For there to be a button that will allow editing of a entry item within the relationship field next to the current delete button.

Yulyaswan:
EECORE-2116
fixes #656


intoeetive (changes requested):
@Yulyaswan when saving related entry the original entry page is reloaded, so if you have any changes there, they will be lost. I think this needs to be handled in a better way

themes/ee/cp/js/build/components/relationship.js
```diff
@@ -258,10 +279,7 @@ function (_React$Component) {
         })),
         full: true,
         iframe: true,
-        dataType: 'json',
```
intoeetive: I think this line might be needed? It's not in the source, but that could be mistake

intoeetive:
@Yulyaswan when I open related entry, and it has relationship field, I have the pencil icon there as well. Clicking it does nothing (which is kinda expected, because I'm in modal already), but I was wondering if we could not show the icons when entry is opened in modal?

intoeetive (approved):

PR #2383, opened 2022-09-08
Resolved issue where File Manager page was not updated after uploading SVG file

intoeetive:


robinsowell (approved):
worked for me- went to file manager, used upload button top right, picked a directory, selected an .svg file, uploaded, and it showed the thumb in the file manager once upload finished.  That was as superadmin.

PR #2398, opened 2022-09-21
Resolved issue where entries could be not displayed on front-end if channel name also existed on other MSM site

intoeetive:
EECORE-2133


robinsowell (approved):
Tests out on both msm sites, with and without using site param.

PR #2405, opened 2022-09-23
Fixed a bug where channel entry search parameter could only be used once

robinsowell:
You should be able to do:

{exp:channel:entries channel="searching" search:seo_title="seriouslies|sblarg" search:seo_title="second|booter" search:contact_email="Third" search:contact_phone="first"}

But in the query, it was only querying one of the custom field tables instead of all 3.

Screen Shot 2022-09-23 at 4 43 20 PM

See how it keeps hitting exp_channel_data_field_10 instead of the tables for the other 2 searches?

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
+++ b/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
@@ -641,6 +641,7 @@ class Channel
                 continue;
             }
 
+            $title_field = FALSE;
             $sites = ($site_ids ? $site_ids : array(ee()->config->item('site_id')));
             foreach ($sites as $site_name => $site_id) {
                 // We're goign to repeat the search on each site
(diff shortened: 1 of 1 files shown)


robinsowell:
@intoeetive Pinged you for review, I know you've been in this code recently.  This was broken before the title search fix- it all came down to:

```if (!isset($search_column_name))```

So it was skipping the other search params once one was set.  Taking that out fixed it- and then I had to add in a check for a title search to make sure that still worked.

What I can't figure out is why that if (!isset($search_column_name)) was in there in the first place.  There must have been a reason, but... I couldn't figure it out.  And everything tests out fine now, as far as I can see.  But I'd be happier if I knew why it was in there at all.

intoeetive:
@robinsowell the `!isset` part was there to check if the search wasn't set up for title or url_title... but apparently that did not work with multiple parameters (or even multiple sites!)

Having that said, I think I have a better fix which does not involve extra variable

Yulyaswan (approved):

PR #2413, opened 2022-09-28
Resolved #2411, #2412 where filepicker was not always working outside of File field

### LINKED ISSUE #2411: 7.1.6 ee('CP/FilePicker') produced dialog shows "loading…" forever if file field is not one of the fields
In my field type, Ansel, I use a native file picker dialog. And we recently [fixed a related bug](https://github.com/ExpressionEngine/ExpressionEngine/issues/2390), but now that it's partially working, I've discovered more bugs.

In this bug, if there are no files/images present in the source directory, and there is no file field, the dialog never shows "no results found" It only continues to show "loading" with the loader behind it. See screenshot.

image

Since I have uploads enabled, it should show a "drop file(s) here to upload" screen. When I include a native file field in the channel's fields, it works properly. See screenshot where I've added a file field to the fields in the channel:

image

### LINKED ISSUE #2412: 7.1.6 ee('CP/FilePicker') produced dialog produces a console error when clicking "upload" if no file field is present
In my field type, Ansel, I use a native file picker dialog. And we recently [fixed a related bug](https://github.com/ExpressionEngine/ExpressionEngine/issues/2390), and I've [opened a new related bug](https://github.com/ExpressionEngine/ExpressionEngine/issues/2411).

In this bug, if a native file field is not present in the field set for the channel, clicking the upload button on the dialog will result in a console error, and nothing happens.

image

If a native file field is present in the channel's fields, the upload button works as expected.

Yulyaswan:
fixes #2411 
fixes #2412 
EECORE-2138

### DIFF
--- a/themes/ee/asset/javascript/src/cp/files/picker.js
+++ b/themes/ee/asset/javascript/src/cp/files/picker.js
@@ -47,7 +47,7 @@
 				}
 			}
 
-			if ($('div[data-file-field-react]').length && typeof(FileField) !== 'undefined') {
+			if ($('div[data-file-field-react]').length) {
 				FileField.renderFields();
 			}
 		});
--- a/system/ee/ExpressionEngine/Addons/rte/Service/CkeditorService.php
+++ b/system/ee/ExpressionEngine/Addons/rte/Service/CkeditorService.php
@@ -25,10 +25,19 @@ class CkeditorService implements RteService
     protected function includeFieldResources()
     {
         if (! static::$_includedFieldResources) {
+            ee()->load->library('file_field');
+            ee()->lang->loadfile('fieldtypes');
+            ee()->file_field->loadDragAndDropAssets();
+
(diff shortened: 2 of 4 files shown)


themes/ee/asset/javascript/src/cp/files/picker.js
```diff
@@ -47,7 +47,7 @@
 				}
 			}
 
-			if ($('div[data-file-field-react]').length && typeof(FileField) !== 'undefined') {
```
intoeetive, line 50: @Yulyaswan I think we had a specific reason why this `!undefined` check was added?

Yulyaswan: It was the wrong decision to fix the previous issue (2390). Because it fixed one issue but created another.

bryannielsen (approved):

PR #2428, opened 2022-10-13
Moved the selected checkboxes to the top of the list if the list is too long; #653

### LINKED ISSUE #653: Show selected categories at the top when editing entry
Please make Categories UI in CP more like new Relationships field - it's great!

Current UI not so great

Yulyaswan:
EECORE-2142
EE7 version of #2427
fixes #653


intoeetive:
I think we need approval from @TomJaeger here as it's affecting every big React Checkboxes field, not just categories

Note: in several other places we implemented this already on the backend

bryannielsen (approved):

PR #2461, opened 2022-10-24
ExpressionEngine 7.2.0

intoeetive:


robinsowell:
I get a javascript error on the publish page if I have a textarea field with the default html buttons on it.  Only field in my test channel, I start typing in the text area, hit enter, get
Screen Shot 2022-10-26 at 1 29 58 PM

Andy confirmed it.  I tested in 7.1 and same exact test didn't trigger the error.

robinsowell:
I'm getting an error trying to use the RTE on the frontend.

Initially had the field set to Redactor.  If I have a custom JS template set, the combo load returns an error "No such property: 'JS_TMPL' on ExpressionEngine\Legacy\Facade"

If I switch the field to use a different Redactor toolbar, I don't get that error, but I get a js error.
Screen Shot 2022-10-28 at 5 20 26 PM

matthewjohns0n (approved):

bryannielsen (approved):

ops-andy (approved):

PR #2494, opened 2022-11-01
Update Generate add-on to use EEObjects and maintain a compatibility mode

matthewjohns0n:
EECORE-2153


system/ee/ExpressionEngine/Service/Generator/ExtensionHookGenerator.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function build()
+    {
+        $hookData = Hooks::getByKey(trim(strtoupper($this->name)));
+
+        // If we didnt get a real hook, set up a default
+        if ($hookData === false) {
+            $hookData = [
+                'name' => $hook,
```
bryannielsen: This is undefined, but not sure where this should be passed from

bryannielsen (changes requested):

system/ee/ExpressionEngine/Service/Generator/ExtensionHookGenerator.php
matthewjohns0n: Good catch 👍

bryannielsen (approved):

PR #2504, opened 2022-11-11
Feature/inline member registration errors

TomJaeger:
Updating member registration tags to support inline errors.

Inline error handeling lines up with channel forms.

Requires you to add
error_handling="inline"

to your {exp:member:registration_form opening tag.
To output errors, each registration_form field name is prefixed with error.
if error:username}{error:username}{/if}

docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/525

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member_register.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member_register.php
@@ -671,8 +671,6 @@ class Member_register extends Member
             ->get('forum_boards');
     }
 
-
-
     /**
      * Member Self-Activation
      */
--- a/system/ee/ExpressionEngine/Addons/member/mod.member_register.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member_register.php
@@ -157,13 +157,6 @@ class Member_register extends Member
             $reg_form = ee()->TMPL->parse_variables($reg_form, array($field_values));
         }
 
-        // parse the errors
-        $error_tags = ee()->session->flashdata('error_tags');
-
-        if (!empty($error_tags)) {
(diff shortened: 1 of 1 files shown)


intoeetive (commented):
@TomJaeger I think we need some example in the docs on how to use it. I see it's setting `error_tags` flash variable but not 100% on how to access it in templates

Also you need to sign our CLA :)

matthewjohns0n:
Due to the failing tests, the CLA assistant blocking merge, and the docs that Yuri mentioned, I'm moving this to 7.2.x.

matthewjohns0n:
@intoeetive I noticed you changed the milestone to 7.2.x. Do you think this would be better in a later release?

TomJaeger:
Docs here https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/compare/feature/inline-member-registration-errors

intoeetive:
@matthewjohns0n you said in previous comment "I'm moving this to 7.2.x" but did not actually move, so I did that =)
Now that we have docs we can probably test and if it's ready then include in this release

intoeetive:
This worked on my tests, but for custom fields the syntax is `{error:m_field_id_1}` instead of `{error:field_name}`
Which might be correct, because `m_field_id_1` is what you need to use as name of field in reg form, but I just want to double confirm with @TomJaeger that this is how it should behave

TomJaeger:
@intoeetive Good question.  In this case, that is correct, we want it to match the field name in the registration form.  Eventually we should consider moving both toward field short name.. But we should keep it consistent with 20 years of implementation for now.

intoeetive (approved):

matthewjohns0n (approved):
Looks good to me 👍

PR #2512, opened 2022-11-14
Added options for displaying the grid layout

Yulyaswan:
EECORE-2155
Added 3 options for Grid layout view:

1. Auto (the grid will be placed in a horizontal row and stacked vertically if there is not enough space)
2. Always use a vertical grid layout
3. Always use a horizontal grid layout

### DIFF
--- a/cp-styles/app/styles/components/_fluid.scss
+++ b/cp-styles/app/styles/components/_fluid.scss
@@ -14,7 +14,7 @@
 	border: 1px solid color(fluid-item-border);
   position: static;
   width: 100%;
-  overflow-x: auto;
+  // overflow-x: auto;
 
   .fluid__item-fieldset {
     padding: 10px 15px 5px 15px;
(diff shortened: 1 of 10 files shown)


intoeetive:
Marking as draft as it needs same changes as https://github.com/ExpressionEngine/ExpressionEngine/pull/2513

intoeetive (changes requested):
@Yulyaswan for some reason the template routes grid became vertical and is now failing the tests. (this did not happen on 7.x version for this). Can you have a look?

ops-andy (approved):
looks good in Firefox and Chrome to me

ops-andy:
@robinsowell if you have a really crazy grid test it might be good to test this on it.

intoeetive (commented):
worked well on my tests on Grid field and Grid within Fluid.

Cypress tests seem to pass now.

However I also see changes to dropdown JS which I don't know if are intended to be here?

intoeetive (approved):

PR #2513, opened 2022-11-14
Added options for displaying the grid layout

Yulyaswan:
EECORE-2155
EE6 version - #2512
Added 3 options for Grid layout view:

1. Auto (the grid will be placed in a horizontal row and stacked vertically if there is not enough space)
2. Always use a vertical grid layout
3. Always use a horizontal grid layout

### DIFF
--- a/cp-styles/app/styles/components/_fluid.scss
+++ b/cp-styles/app/styles/components/_fluid.scss
@@ -98,6 +98,7 @@
 
 	border-radius: $corners-big;
 	transition: background 100ms ease, box-shadow 100ms ease;
+	width: 100%;
 }
 
 .fluid__item-tools {
--- a/cp-styles/app/styles/components/_fluid.scss
+++ b/cp-styles/app/styles/components/_fluid.scss
@@ -14,6 +14,7 @@
 	border: 1px solid color(fluid-item-border);
   position: static;
   width: 100%;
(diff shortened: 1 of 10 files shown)


intoeetive (changes requested):
It does not seems to respect previous settings.

I had a grid (used within a Fluid, but I don't know if that makes a difference) set to vertical layout. After I switched to this branch, the grid was displayed horizontally.

Also I think the text for settings needs some changes.
Proposing this:
Grid layout
Choose how the field is displayed on publish pages (in small font)
Auto
Vertical layout
Horizontal layout

If we need extra explanation for "auto" option, it should be in small font (but I don't think we need it)

intoeetive (changes requested):
Vertical now works fine, but setting to "horizontal" behaves same as "auto" - it sets the grid layout to vertical if I narrow the window.

Also another thing I noticed - on vertical layout, the column instructions are italic and black instead of grey.

matthewjohns0n:
I'm pushing this to 7.2.x since there still seem to be changes that are needed on this one.

intoeetive (dismissed):
Looks good for me - do we need extra reviews here @ops-andy ?

ops-andy (approved):
looks good in Firefox and Chrome to me

PR #2517, opened 2022-11-15
Better redactor settings

jamesseavers:
https://eecms.slack.com/archives/C04CUNNR9/p1668533312170649?thread_ts=1668521860.541449&cid=C04CUNNR9

Changes the Redactor RTE settings to add in image resizing and alignment.

Settings amended are:

imageResizable: true,
imagePosition: true

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x ] Yes
- [ ] No


Yulyaswan (approved):

intoeetive (commented):
I am not sure we should have `imageResizable` on by default. I feel like this is dangerous setting, as this might encourage some content editors to upload heavy images and resize them in RTE field, thinking this will make them smaller (thus causing the pages become very heavy).

With the new Advanced Configuration feature, people can still set this option with configuration JSON, but I believe this should be off by default.

Leaving this to @TomJaeger discretion and setting to draft for now

jamesseavers:
Thanks @intoeetive. Whilst I don't necessarily agree with not giving people access to valuable tools in case they don't know how to use them properly, I can confirm that the new Advanced Configuration in RTE resolves this issue.

PR #2519, opened 2022-11-15
Resolved #2490 where visual indication was missing when Directory Sync was started

### LINKED ISSUE #2490: Add wait cursor when File / Image Sync is running
When you Directory Sync on a File location the process is handled via some kind of async process. The only way of telling that this is running is the moving line in the progress indicator - but this starts small and may take a few seconds to appear. When the sync process is triggered the mouse pointer remains in its default state, and the Sync Directory button remains active. On slow systems this may give the user the impression that the button click has not been registered and may result in multiple clicks on the button etc. 

When a Directory Sync starts, the cursor shifts to a wait state, and the Directory Sync button is disabled (and possible has a wait animation replacing the Directory Sync term).

Getting a faster computer... 

Should be self explanatory.

Yulyaswan:
Resolved #2490 
After clicking the Directory Sync button, the cursor shifts to a wait state and the Directory Sync button is disabled

### DIFF
--- a/themes/ee/asset/javascript/src/cp/fields/synchronize.js
+++ b/themes/ee/asset/javascript/src/cp/fields/synchronize.js
@@ -24,7 +24,9 @@ EE.fieldManager.sync_listen = function() {
 		event.preventDefault();
 
 		// Disable sync button
-		$('.button', this).prop('disabled', true);
+		$('.button', $('.form-standard form')).each(function() {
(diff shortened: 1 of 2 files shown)


intoeetive:
@Yulyaswan can you confirm we don't need same change for EE7

Yulyaswan:
@intoeetive we already have this in EE7

intoeetive (approved):
Approving as this is the code we have working in EE7

PR #2527, opened 2022-11-17
Resolved #2525 where using `channel="not ..."` parameter was returning no results on `{exp:channel:entries}`

### LINKED ISSUE #2525: 7.2 channel entry tag wrong results with 'not' in channel param.
In both 7.2 and 6.4 there appears to be a bug in the channel entry tag when the channel parameter has the 'not' specifier in it.

ex: I have an entry in the 'about' channel with the url title 'about'.  This tag?  Does not pull it back:

```{exp:channel:entries channel="not blog" url_title="about|about-overview" dynamic="no"}```

This tag does

```{exp:channel:entries channel="about" url_title="about|about-overview" dynamic="no"}```

Definitely worked in 6.2.4

intoeetive:
Resolved #2525 where using channel="not ..." parameter was returning no results on {exp:channel:entries}


robinsowell (approved):
Worked for me, trying various combos of 'not' and without 'not'.

matthewjohns0n (approved):

PR #2537, opened 2022-11-18
Allow raw data to be set on template library

bryannielsen:
## Overview

This change should have no effect on existing template parsing behavior but it allows add-ons the opportunity to make the template library aware of the raw data used in processing a tag.

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No


intoeetive (changes requested):
Can you PSR-12 the code please? (spaces)

Also I think we need a line in the docs on what this does

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,34 @@
...
+    ->in(__DIR__);
+
+return (new PhpCsFixer\Config)
+    ->setRules([
+        '@PSR12' => true,
+        'array_indentation' => true,
+        'binary_operator_spaces' => [
+            'operators' => [
+                '=' => 'single_space',
+                '=>' => 'single_space',
```
matthewjohns0n: Lets change this rule to `null`

matthewjohns0n (changes requested):

matthewjohns0n (approved):

intoeetive (approved):

PR #2552, opened 2022-11-22
few fixed to legacy member model

TomJaeger:
I don't think the core uses this any more.. But some older add-ons do.

Still needs some updates from member_groups to role_groups


TomJaeger:
I'm realizing this should probably be PR'd against 6.4 as well.

intoeetive (approved):

matthewjohns0n (approved):

PR #2554, opened 2022-11-23
Resolved #2432 where some Structure queries were slow on large websites

### LINKED ISSUE #2432: Slow query with Structure when using channel:title parameter
There is an issue with Structure when using the channel:title="channel:field" parameter. The issue is only seen when there are a lot of Structure entries (in my case 600+), and also lots of fields (100+ across the entire site). These entries are spread across a few channels. The issue is that there is 1 query that is taking almost **2 seconds** to run every time the page is loaded (it doesn’t seem to get cached). 

The query is in sql.structure.php on line 992  in the **_get_page_titles()** function:

`// Get the Channel Entries that have are in a Structure Channel`
`        $channelEntries = ee('Model')->get('ChannelEntry')`
`            ->filter('channel_id', 'IN', $structure_channel_ids)`
`            ->filter('site_id', '==', $this->site_id)`
`            ->all();`

**How To Reproduce**
1. You need to add in quite a lot of fields and entries into Structure (as pages)
2. Use the `channel:title="channel:field"` parameter in a {exp:structure:nav_basic} tag
3. Run the template ... and you will see it's very slow. I've isolated down the query issue to the one above.

**Error Messages**
There are no error messages. It works ... it's just VERY slow.

N/A

**Environment Details:**
 - Version: EE7.1.6
 - PHP Version: 8.0.0
 - MySQL Version: 5.7.34
 - Web Server: Apache

**Possible Solution**
I've added some caching to the _get_page_titles() function, and this works really well after it's initially loaded (and the cache set). I think the Model could be re-written to be more performant with larger numbers of entries ... but I don't know how to do that.

intoeetive:
Resolved #2432 where some Structure queries were slow on large websites
 
mark all channels as assigned when performing CLI requests

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/sql.structure.php
+++ b/system/ee/ExpressionEngine/Addons/structure/sql.structure.php
@@ -990,8 +990,14 @@ class Sql_structure
         $structure_channel_ids = $this->_get_structure_channel_ids($include_listings);
 
         // Get the Channel Entries that have are in a Structure Channel
-        $channelEntries = ee('Model')->get('ChannelEntry')
-            ->filter('channel_id', 'IN', $structure_channel_ids)
+        $selectFields = array_map(function($field_data) {
+            return 'field_id_' . $field_data['field_id'];
+        }, $sql_fields);
+        $channelEntries = ee('Model')->get('ChannelEntry')->fields('entry_id', 'channel_id', 'site_id', 'title');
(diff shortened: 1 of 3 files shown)


intoeetive:
This has been tested and confirmed to work by the reporter: https://github.com/ExpressionEngine/ExpressionEngine/issues/2432#issuecomment-1325727106

Yulyaswan (approved):

PR #2572, opened 2022-11-30
Allow cloning of channel fields, layouts, member roles and categories; #2032, #561

### LINKED ISSUE #561: Ability to Duplicate and/or Export/Import Fieldtypes from the Field Manager
I'm always frustrated when I have to recreate common fieldtypes from scratch in the Field Manager. Often, I may have already created the fieldtype I want, I just need to change its Name + Short Name. 

When building a new website, I like to start out with 4 core fieldtypes. (3 are grid fields containing text or textarea fields, and 1 is just a textarea field). Then I create a fluid field to contain those 4 core fields. 

If the website is part of a Multiple Site Manager system or a multilingual website, I'll add the _**site's**_ Short Name to the _**field's**_ Short Name, but the fieldtypes remain the same. 

I would love a way to duplicate existing fieldtypes in the Field Manager and/or be able to export them and then import them into the same or other websites. 

I can see a challenge being if the field contains file directory settings, or is a relationship field dependent on channel entries, but for some of the most basic fields (like textarea), this would still speed up things for me. 

Ideally, there'd be a prompt (when duplicating or importing a field) to change the field's Name + Short Name. If it's also possible to also duplicate/import more sophisticated fields, then the prompt could also include reconfiguring those field settings.

### LINKED ISSUE #2032: Allow for the duplication of publish layouts
Often I'd like to only slightly modify a publish layout for a user type, but the work involved in re-creating a layout is just too much.

intoeetive:
Allow cloning channel layout for another member role; closes #2032

Allow Channel Fields to be cloned; closes #561

cloning member roles

Allow categories to be cloned


matthewjohns0n (dismissed):
This looks really good, but I did find one issue. When I go to clone a category, it works fine. But if I go back and clone the same original category, I get this error:

Screenshot 2023-01-19 at 5 02 20 PM

If I then fix the error shown and press save, it edits the _original category_:

Screenshot 2023-01-19 at 5 04 11 PM

Notice how my first category is now changed.

I think if we just make sure there cant be collisions when cloning, it would solve both problems.

matthewjohns0n (approved):
Looks good 👍🏼

PR #2597, opened 2022-12-02
Resolved issue where Pro Search filters were using Low Search naming

bakin1999:
## Overview

When we moved from low search to pro search.  We have pro search trying to pull in filters from other addons that are still extending off of low search.  My idea was that we separate out based on EE version to avoid errors in pro search whenever someone has another addon that has a filter.

The one thing that is a painpoint doing it this way is that all the addons that have a filter would have to just copy that file give it the name psf.addon_name.php and then change the class line.  (Super quick but seems redundent) and also all the addon managers would have to do that.  But I dont see other way to get around this while keeping addons compatible for ee 7 and ee6 and below.

If someone has better idea please share :)

## Nature of This Change

- [x] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_filters.php
+++ b/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_filters.php
@@ -102,15 +102,20 @@ class Pro_search_filters
                 }
 
                 // Compose file name
-                $file = $dir . "/lsf.{$item}.php";
+                // Compose class name
+                if (version_compare(ee()->config->item('app_version'), '7.0.0', '>=')) {
+                    $file = $dir . "/psf.{$item}.php";
+                    $class = 'Pro_search_filter_' . $item;
+                }
+                else {
+                    $file = $dir . "/lsf.{$item}.php";
+                    $class = 'Low_search_filter_' . $item;
+                }
 
                 // Skip if not a file
                 if (! file_exists($file)) {
                     continue;
                 }
-
-                // Compose class name
-                $class = 'Pro_search_filter_' . $item;
 
                 if (! class_exists($class)) {
                     require_once($file);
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_filters.php
```diff
@@ -102,16 +102,21 @@ public function __construct()
                 }
 
                 // Compose file name
-                $file = $dir . "/lsf.{$item}.php";
+                // Compose class name
+                if (version_compare(ee()->config->item('app_version'), '7.0.0', '>=')) {
```
intoeetive: This line would always return `true` because we're including this only for EE7 :)

system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_filters.php
```diff
@@ -102,16 +102,21 @@ public function __construct()
...
+                    $file = $dir . "/psf.{$item}.php";
+                    $class = 'Pro_search_filter_' . $item;
+                }
+                else {
```
intoeetive: Needs to be `} else {` to be PSR-12 compatible

intoeetive (changes requested):
I think this just will not work - see the comment to line 106

system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_filters.php
```diff
@@ -102,16 +102,15 @@ public function __construct()
                 }
 
                 // Compose file name
-                $file = $dir . "/lsf.{$item}.php";
```
matthewjohns0n, line 105: Unfortunately, I'm not sure if we can remove support for lsf files right now. I think it will most likely break any existing site that uses them. We probably need to check for the existence of `lsf.{$item}.php` and `psf.{$item}.php`. This is why I just left it as lsf, since it would allow everything existing to just work.

matthewjohns0n (changes requested):

matthewjohns0n:
I wasnt able to commit to @bakin1999's repo, so opened a new PR: https://github.com/ExpressionEngine/ExpressionEngine/pull/2706

PR #2598, opened 2022-12-04
Resolved issue when PHP deprecation messages could be shown in Pro Search when using PHP 8.1

projectnoontide:
Deprecated: Automatic conversion of false to array is deprecated

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_fields.php
+++ b/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_fields.php
@@ -281,11 +281,12 @@ class Pro_search_fields
         $cols = pro_get_cache(__CLASS__, $table);
 
         if (! isset($cols[$field_id])) {
-            // Init
+            // If $cols is not an array, we need to make it an array first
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_fields.php
```diff
@@ -282,7 +282,10 @@ private function _col($field_id, $col_name, $table = 'grid_columns')
 
         if (! isset($cols[$field_id])) {
             // Init
-            $cols[$field_id] = array();
+            if (!is_array($cols)) {
+                $cols = [];
+                $cols[$field_id] = [];
```
matthewjohns0n: Looks like if `$cols` is an array arleady, but `$cols[$field_id]` is not set, then `$cols[$field_id] = []` will not get set. I think if you move the `$cols[$field_id] = [];` assignment after the `if` block it should be good to include.

matthewjohns0n (changes requested):

matthewjohns0n (dismissed):

intoeetive (approved):

PR #2603, opened 2022-12-05
ExpressionEngine 7.2.3

matthewjohns0n:


system/ee/ExpressionEngine/View/_shared/add-on-card.php
```diff
@@ -10,6 +10,18 @@
 		<?php if (!empty($addon['description'])): ?>
 		<p class="add-on-card__desc" title="<?= $addon['description'] ?>"><?= $addon['description'] ?></p>
 		<?php endif; ?>
+
+		<?php if ($addon['installed'] && $addon['developer'] == 'EEHarbor') : ?>
```
intoeetive, line 14: Is this something we intended to include? We tried to avoid previously direct mentions of `EEHarbor` in the code, because even though it's almost same team of developers, technically this is different vendor, and we try to not include vendor-specific stuff

intoeetive:
What I noticed is that when I click "update available" for EEHarbor add-on, the URLs goes straight to external page, while clicking the documentation link shows the intermediate confirmation "you are about to leave this site" page first.

Also the "update" tooltip for this link does not seem to be using language file.

And when I click on license state ("unlicensed" or "trial") it does nothing

I think this feature is not ready to be rolled in and the merge needs to be reverted

ops-andy (approved):
tested. everything seems good to me

intoeetive (approved):

bryannielsen (approved):

PR #2613, opened 2022-12-07
modifier updates

### LINKED ISSUE #136: JSON variable modifier ignores file image manipulation

JSON Variable Modifiers does not parse / ignore file field with image manipulation; instead it will only output the original file url

**How To Reproduce**

1. Create a file field
2. Create any setting for Constrain or Crop in the File Manager page
3. Upload file
4. Parse template using:
`{your_file_field:json}`
`{your_file_field:your_crop:json}`

**Error Messages**

No error message. Instead it will output both tags value as the original file name url.

**Environment Details:**
 - Version: [e.g. 5.1.3]
 - PHP Version [e.g. 7.2]
 - MySQL Version [e.g. 5.7]
 - OS: [e.g. Linux]
 - Web Server: [Apache]

**Possible Solution**

**Additional context**

### LINKED ISSUE #1203: Provide for additional variable modifier traits to be registered by add-ons
n/a

The [variable modifier](https://docs.expressionengine.com/latest/templates/variable-modifiers.html#variable-modifiers) system is a useful addition / improvement over tag based variable modification in some use cases.  

Currently these modifications are defined as traits in the file `/system/ee/ExpressionEngine/Service/Template/Variables/ModifiableTrait.php` and there is no (obvious at least) way for add-ons to extend this list.  

For some add-ons the ability to extend the list would provide user benefits similar to those that are obtained from the existance of the current traits.

It may be that there is an (undocumented) method that could be used to extend this list - if so if it can be described then I'd happily add this explanation to the current docs.  

If there is not, one solution would be replace the current hard-wired list with code that reads defined traits from a table or config file or similar, and include in this the option to read additional traits from a separate table containing additional traits defined by add-ons.  Then a mechanism to support the registration of new traits could be added to the add-on setup process, and add-ons could add their own variable modifiers to EE... 

Add-ons can carry on using single / double tags as currently.

User teachability / documentation would be covered by documentation provided by add-on developers who would like to use the system.  However some additional developer documentation would be required to document the mechanism of defining / registering additional traits.

The change in approach would not necessarily have any impact on existing function of EE, so there should be no compatability issues or need for a migration strategy.

- [X] I am capable and would like to help with implementation of this feature if it is considered (but I would need some guidance from better minds about preferred method of implementing if idea is accepted etc.)

intoeetive:
Allow variables to use multiple modifiers; closes #136
Enabled additional variable modifiers to be registered by add-ons; closes #1203

Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/546

- allow specifying parameters for modifier by prefixing parameter name with modifier name. E.g. `limit:characters="1"` would work the same as `characters="1"` when using `limit` modifier (but also have higher priority)
- allow modifiers to be chained, e.g. `var_name:limit:spellout`. 
- allow add-ons to provide their modifiers. They would need to be registered in `addon.setup.php` file and then be called with `__call` in ModifiableTrait. The `method_exists` checks would then need to be updated to also check whether modifier is registered by add-on


intoeetive:
@bryannielsen I made the necessary updates so that File field modifiers could be chained.

As of moving the `__call` method contents into separate function, I figured out that is not necessary - implementing `__call` in fieldtype level does not make any effect, because all methods require certain syntax to get parsed / processed - there's no chance someone would implement those

bryannielsen (approved):

PR #2634, opened 2022-12-12
Allow swatches in Colorpicker fieldtype to have names

litzinger:
## Overview

Color picker is great, but the hex color value doesn't jive well with site css. Usually colors or themes are given a name and assigned through a class instead of relying on `<div style="color: {colorpicker_field}">` this change provides an optional Name field so we use `div class="my-thing__{colorpicker_field:name}">` instead.

- [ ] 🐛 Fixes a bug
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔒 Improves security 

- [x] Yes
- [ ] No

## Documentation
If this gets approved in this form or some form I'll update the docs

### DIFF
--- a/cp-styles/app/styles/legacy/_key-value-item.scss
+++ b/cp-styles/app/styles/legacy/_key-value-item.scss
@@ -36,7 +36,7 @@
 		}
 
 		.field-control{
-			flex: 1 1 auto;
+			flex: 1 1 0;
 			padding: 0 5px;
 		}
 
@@ -49,7 +49,7 @@
 	display: flex;
 
 		.field-instruct{
-			flex: 1 1 auto;
+			flex: 1 1 0;
 			margin-bottom: 5px;
 
 				label{
@@ -65,7 +65,7 @@
 				}
 
 				&:last-child{
-					padding-left: 0;
+					padding-left: 5px;
 				}
 		}
 }
--- a/system/ee/ExpressionEngine/Addons/colorpicker/ft.colorpicker.php
+++ b/system/ee/ExpressionEngine/Addons/colorpicker/ft.colorpicker.php
@@ -95,7 +95,7 @@ class Colorpicker_ft extends EE_Fieldtype
     {
         ee()->cp->add_js_script('file', array('library/simplecolor', 'components/colorpicker'));
 
-        return $this->create_colorpicker([
+        return $this->createColorPicker([
             'allowedColors' => $this->get_setting('allowed_colors'),
             'inputId' => $this->field_id,
             'inputName' => $this->field_name,
@@ -110,7 +110,7 @@ class Colorpicker_ft extends EE_Fieldtype
      * Creates a color picker input with the specified values
      * The values are passed to the react color picker component
     */
-    private function create_colorpicker($info, $disabled = false)
+    private function createColorPicker($info, $disabled = false)
     {
         $data = base64_encode(json_encode($info));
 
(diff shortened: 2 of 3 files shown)


litzinger:
Screenshot 2022-12-12 at 10 26 55 AM
Screenshot 2022-12-12 at 10 12 35 AM

intoeetive (changes requested):
@litzinger thank you for submission!

Found an issue with it though - if you have several swatches with color and name, and then you add a swatch and don't provide name for it - it would duplicate the name of the previous swatch. 

Also it would be good to have headers for the columns (like the dropdown field options have) because it's not 100% clear what should go into second column once the placeholder is not there (for existing rows)

litzinger:
Good points. The headers though do not align vertically with the columns when using the MiniGrid field... they haven't for quite awhile, so I opted to use the placeholder text.

I'll take a look at the duplication issue when I find some time.

litzinger:
I made a small update based on the changes you suggested.
Screenshot 2023-01-04 at 2 22 08 PM

This one intentionally should have no color in the 3rd row and 2 empty name values after I tested adding additional rows and also sorting them and then saving the field settings.
Screenshot 2023-01-04 at 2 43 46 PM

intoeetive:
@litzinger this does not seem to have issues I noticed before, so I think we're good about including it. Thank you!

@Yulyaswan can you just test this once again and if possible provide better alignment for 'Name' heading on Swatches table?

Yulyaswan (approved):

litzinger:
I took a look at the alignment thing and I think it'll involve moving the div that contains the headers inside of the `field-control` div, and only printing the header on the first row. I almost had this completed but decided to undo the changes b/c it felt like a separate unit of work, thus a separate PR. I don't know if there is any CSS Grid/Flexbox magic that could do it better though.

This wasn't 100% working but it was the path I was headed down:

```<div class="fields-keyvalue<?=(isset($classes)) ? $classes : ''?>">
    <div class="field-no-results hidden">
        <p><?=lang($no_results['text'])?>
            <?php if (! empty($no_results['action_text'])): ?>
                <a<?=$no_results['external'] ? ' rel="external"' : '' ?> rel="add_row" href="<?=$no_results['action_link']?>"><?=lang($no_results['action_text'])?></a>
            <?php endif ?>
        </p>
    </div>
    <div class="keyvalue-item-container">
        <?php
        $index = 0;
        foreach ($data as $heading => $rows):
            $rows = array($rows); ?>
            <?php foreach ($rows as $rowIndex => $row): ?>
                <?php
                if (! isset($row['attrs']['class'])) {
                    $row['attrs']['class'] = '';
                }
                $row['attrs']['class'] .= ' fields-keyvalue-item';
                ?>
                <div<?php foreach ($row['attrs'] as $key => $value):?> <?=$key?>="<?=$value?>"<?php endforeach ?>>
                    <?php if ($reorder): ?>
                        <ul class="toolbar">
                            <li class="reorder"><a href="#" title="reorder row"></a></li>
                        </ul>
                    <?php endif ?>
                    <?php foreach ($row['columns'] as $colIndex => $column): ?>
                        <div class="field-control">
                            <?php if ($index === 0 && isset($columns[$colIndex])): ?>
                                <div class="field-instruct"><?= $columns[$colIndex]['label'] ?></div>
                            <?php endif; ?>
                            <?=$column['html']?>
                        </div>
                    <?php endforeach ?>
                    <ul class="toolbar">
                        <li class="remove"><a href="#" rel="remove_row" title="remove row"></a></li>
                    </ul>
                </div>
            <?php endforeach ?>
        <?php
        $index++;
        endforeach ?>
    </div>
    <a class="button button--default button--small" rel="add_row" href=""><?=lang('add_a_row')?></a>
</div>
```

intoeetive (approved):

PR #2647, opened 2022-12-15
Enabled renaming custom tabs in publish layouts; #792

### LINKED ISSUE #792: Edit publish layout: allow tab renaming
Because I can't rename existing tabs, I have to delete existing tabs, then create a new one. This means I have to move fields out of the existing tab, create a new tab, then move fields into the new one (or alternatively create the new tab first, then move fields to it, then delete the old one).

Allow users to rename tabs.

Yulyaswan:
EECORE-2161
fixes #792


intoeetive (changes requested):
@Yulyaswan found some issues when doing deleting together with renaming 

I had an empty custom tab (tab1), then built-in Options tab, then another custom tab (tab2).
I deleted tab1, then tried to rename tab2, but instead it renamed my Options tab

intoeetive (approved):

PR #2665, opened 2022-12-22
Resolved #2500 where reordering Grid with images had some issues

### LINKED ISSUE #2500: Grid reordering with images is v. buggy
When adding images to file grid, or grid fields and then trying to reorder the rows with drag behaviour the UI feedback is jittery. It's a bit of lottery whether or not the row dragged up will appear above the field (usually not, sometimes moving a few rows down even). It's not that great without images, but with, it feels like a safecracking level of sensitivity and the UI indicators for row placement are not clear at all. 

This is especially apparent on grids with more rows than visible in the viewport (e.g. more than 5/6).

Other drag and drop behaviours in the control panel work well — Relationships, fluid fields or reordering the actual grid field type columns (admin view) for example.

Tested in Firefox (106.0.2) / Chrome (107.0.5304.87) — both Mac OS

Yulyaswan:
fixes #2500

### DIFF
--- a/themes/ee/asset/javascript/src/cp/grid.js
+++ b/themes/ee/asset/javascript/src/cp/grid.js
@@ -7,6 +7,7 @@
  * @license   https://expressionengine.com/license Licensed under Apache License, Version 2.0
  */
 
+
 (function($) {
 
 /**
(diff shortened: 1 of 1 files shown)


intoeetive (approved):
approved by reporter in the issue

PR #2694, opened 2023-01-04
Resolved issue with View Type preference not properly encoded

matthewjohns0n:


system/ee/ExpressionEngine/Service/File/ViewType.php
```diff
@@ -27,9 +27,8 @@ public function determineViewType($destination = 'all', $viewtype = 'list')
     {
         $views = ['list', 'thumb'];
         $viewtype_prefs = [];
-
         if (ee()->input->cookie('viewtype')) {
-            $viewtype_prefs = unserialize(ee()->input->cookie('viewtype'));
+            $viewtype_prefs = json_decode(ee()->input->cookie('viewtype'), true);
```
bryannielsen, line 31: If we were previously serializing these cookies do we need some error handling around this?  I'm not sure how long the cookies live for but I imagine it's possible we could be trying to json_decode a serialized value during the transition?

intoeetive: I agree with what Bryan is saying. We had similar issue with Entry Manager views (switching from serializing to JSON encoding or vice versa) and it was bad experience for people who had that saved already. We need some fallback, at least, if we can't avoid the change

matthewjohns0n: I was able to regex the settings out of the serialized string, and then json encode & set them in the cookie, so there is no loss of data at all.

intoeetive (approved):
worked well for me after fixing the typo

bryannielsen (approved):

PR #2706, opened 2023-01-06
Resolved issue where Pro Search filters were using Low Search naming

matthewjohns0n:
When we moved from low search to pro search. We have pro search trying to pull in filters from other addons that are still extending off of low search. My idea was that we separate out based on EE version to avoid errors in pro search whenever someone has another addon that has a filter.

The one thing that is a pain point doing it this way is that all the addons that have a filter would have to just copy that file give it the name psf.addon_name.php and then change the class line. (Super quick but seems redundant) and also all the addon managers would have to do that. But I don't see another way to get around this while keeping add-ons compatible for ee 7 and ee6 and below.

Original PR: #2597


intoeetive:
The code looks good to me, though I'm not sure how to actually test it

intoeetive (approved):

PR #2716, opened 2023-01-09
Fixed bug when user leaves low search in install

matthewjohns0n:


robinsowell (approved):
worked for me with low events and low search still in tehre

PR #2721, opened 2023-01-09
Set header_sorts to null only if it is not set

matthewjohns0n:


intoeetive (dismissed):
Approving; we still need PSR-12 fixes, but that might be larger undertake

Yulyaswan:
We still have a problem with Channel Form. For an empty grid is incorrectly calculated `colspan` and because of this `Add new row` the column does not occupy the entire row 

Screenshot 2023-01-10 at 11 11 17

TomJaeger:
Changed this to 7.2.x since we're reverting the the PR that caused this issue.

matthewjohns0n:
Fixed here: https://github.com/ExpressionEngine/ExpressionEngine/pull/2779 so I'm closing this one.

PR #2726, opened 2023-01-10
Resolved issue where Channel Form grid has extra unneeded column

Yulyaswan:
EE7 version of #2727 
EECORE-2175


intoeetive:
@Yulyaswan can you tell how to replicate the original issue? I wasn't able to replicate when using `{field:grid}` tag in 7.2.6

Yulyaswan:
@intoeetive choose `Vertical layout` for the grid and it should add this column

intoeetive (approved):

matthewjohns0n:
Tests were failing, and I think the root cause was fixed in `7.dev`, so I merged it back into this branch and am allowing the tests to re-run.

PR #2733, opened 2023-01-10
Resolved #2661 where is was not possible to search template group and template at the same time in Pages tab

### LINKED ISSUE #2661: Entry Template selection improvement
When editing an Entry and choosing which Template it should output to... the Search option in the dropdown isn't helpful when trying to select an Index template of a Template Group.

For example, if I have a Template group called "contact" and I want to assign an Entry to the "contact/index" template:

If I search for "contact", it only shows actual Templates named 'contact'... not the Template Group.
If I search for "index", it lists all the 'index' templates, but I have no idea when Group each one is within.

Screenshot 2022-12-21 at 12 33 06

Yulyaswan:
Added possibility to search through Template group
fixed #2661 
EECORE-2172


intoeetive (approved):
Tested, works well. Code looks good as well

PR #2751, opened 2023-01-13
Resolved #2437 where building Pro Search Collection via URL did not work properly

### LINKED ISSUE #2437: Horizontal grid fields can 'blow out'
It's happening to file location manipulations plus custom fields so it seems to be persistent no matter how those tables are being displayed.  If you resize the page, they break out of container:

grid-fields-trim

### LINKED ISSUE #2743: Pro search collection index url needs a key
Pro Search Collections can be built via cron etc using a url- see https://docs.expressionengine.com/latest/add-ons/pro-search/collections.html#automate-building-of-search-indexes

It requires a key, that is typically the low search license.  Which there isn't one.  And it doesn't appear to work without a key- I get a 'permission for this action' error when I visit without a key.

matthewjohns0n:
Fixes #2743

Adds setting to pro search:
Screenshot 2023-01-13 at 4 02 35 PM

It will automatically generate a key on install and on upgrade. If there is still a license key saved in the settings from Low Search, it will use that so existing ACT URLs continue to work.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro_search/addon.json
+++ b/system/ee/ExpressionEngine/Addons/pro_search/addon.json
@@ -1,7 +1,7 @@
 {
     "name": "Pro Search",
     "shortname": "pro_search",
-    "version": "8.0.1",
+    "version": "8.0.2",
     "description": "Powerful site search and Find &amp; Replace utility",
     "trial": true,
     "namespace": "Pro\\Search",
--- a/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_settings.php
+++ b/system/ee/ExpressionEngine/Addons/pro_search/libraries/Pro_search_settings.php
@@ -45,6 +45,7 @@ class Pro_search_settings
         'search_log_size'     => '500',
         'ignore_words'        => 'a an and the or of s',
         'disabled_filters'    => array(),
+        'build_index_act_key' => '',
         'stop_words'          =>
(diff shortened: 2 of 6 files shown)


robinsowell:
So I tested, and it did solve the permission issue, but I do get an invalid action.  Problem being, I don't know how it SHOULD work.  So- did this url:

https://7dev.local:8890/?ACT=52&key=171355B05AE5C3C562F9BC0417ABC3FAAF7F4E2D

Get a 'Invalid action' and from debugging, I know that's from line 1473ish in mod.pro_search.php
```
        } else {
            // Invalid action
            show_error('Invalid actionz');
        }
```
I stuck the 'z' on it so I knew what was being triggered.

This MAY be perfectly correct behavior.  I stuck a collectionid=1 on it, and now I get no error, so maybe that's just required?

robinsowell (dismissed):
Does throw an error in php 8.1:

```
preg_split(): Passing null to parameter #3 ($limit) of type int is deprecated

ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php, line 1499
```
But since that's a seperate error not introduced by this code, and works fine in php 7, approving.

matthewjohns0n:
@robinsowell I figured we might as well hit that PHP 8.1 issue now that we found it - just pushed up a fix so this needs to be re-reviewed.

robinsowell:
Yay, no php 8.1 error!  Plus I checked debug and as long as a collection is specified, it is doing the index update.

robinsowell (approved):

PR #2756, opened 2023-01-16
Resolved issue when files on non-local filesystem could not be resized or cropped from CP

bryannielsen:
This update fixes a bug where performing a crop or resize on an image through the 'edit' view would fail when using a non-local filesystem.  Changing the file manager methods to accept an optional filesystem object allow us to ensure the transformations happen locally in temp files and then are moved to the appropriate destination.

### DIFF
--- a/system/ee/legacy/libraries/Filemanager.php
+++ b/system/ee/legacy/libraries/Filemanager.php
@@ -30,7 +30,7 @@ class Filemanager
     /**
      * Constructor
      *
-     * @access	public
+     * @access public
      */
     public function __construct()
     {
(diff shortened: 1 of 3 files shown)


matthewjohns0n (commented):
I made some PSR-12 changes that github was complaining about. The code looks good, but I haven't had a chance to review it since I don't have an external filesystem setup.

@bryannielsen since I didn't pull this down and test it myself, I haven't "approved" yet. Might be good to pull someone in who has S3 or similar setup, and if no one else does I can spend a bit of time setting this up myself and testing it.

system/ee/legacy/libraries/Filemanager.php
```diff
@@ -2721,8 +2721,12 @@ public function _do_image_processing($redirect = true)
     /**
      * Image crop
      */
-    public function _do_crop($file_path)
+    public function _do_crop($file_path, $filesystem = null)
     {
+        $filesystem = ($filesystem) ?: ee('Filesystem');
+        $source = $filesystem->copyToTempFile($file_path);
+        $new = $filesystem->createTempFile();
```
intoeetive, line 2728: This seems obsolete when working with local filesystem

intoeetive (commented):
@bryannielsen please see the comment in code.

Also, does this work for pre-defined and on-the-fly (template) manipulations?

system/ee/legacy/libraries/Filemanager.php
bryannielsen: It is unnecessary on the local filesystem but it doesn't have a big performance hit and standardizes the code.  We're doing the same thing in other areas when we switched to Flysys.  It would be nice to refactor and avoid this down the road.

intoeetive:
@bryannielsen does this also work for pre-defined and on-the-fly (template) manipulations?

bryannielsen:
@intoeetive this was just fixing the behavior of the edit image controller to be in line with the on-the-fly and pre-defined manipulation behavior.  Those are already using temp files and working properly with remote filesystems.  This controller just used different methods for crop/resize/rotate that weren't updated with the 7.x filesystem changes.

intoeetive (approved):
tested, worked fine

Issue #767, opened 2021-01-08
Setting the 'search excerpt' field in a Fluid field


robinsowell:
In the channel settings, you can select a Fluid field to use as the 'search excerpt'.  When you do, it will combine all of the fluid field's field's data to make the excerpt, which is likely not desirable.

For example, I had an entry with a fluid field that had to textareas and a grid with several text fields.   My excerpt looked like:

`I have booty in my content Some more fluid content. tradd ct stuff||seriouslies|mind|`

The pipe separated is the content of the grid field.

On possible solution would be to show all of the fields in the grid field in the excerpt selection and you select on of those, rather than the whole fluid field.  Which still leaves you with the issue of having multiple fields, but one possibility would be to just use the first field of that type and no others.

In any case, the current behavior can be problematic.

intoeetive:
I think we should only allow fields that hold text-like data to be used as search excerpt. Not ones that hold array-like data, such as Fluid, Grid, Relationship, or even Checkboxes.

And maybe just the ones that are marked as searchable.

@bryannielsen I'd like to hear your thoughts on this too

PR #2757, opened 2023-01-16
search module updates

### LINKED ISSUE #66: Ordering search results by {view_count_xxx} variables
It would be great if **{view_count_xxx}** variables are allowed to sort search results.

### LINKED ISSUE #83: exp:search:search_results excerpt tag displaying “xhtml” (as in the field formatting choice)
exp:search:search_results excerpt tag displaying “xhtml” (as in the field formatting choice) when there are fewer than 50 words to display AND the textarea (or other field) is within a Fluid field and followed by a Grid.

Fluid Field
- text area (with fewer than 50 words to display)
- grid

We have a fluid field and within it is a textarea followed by a grid field.

Since EE’s Channel settings only show the fluid field as a Search excerpt choice, we couldn’t select the textarea field specifically. When there’s not enough text in the textarea, EE is merging its content with the next field-type in the fluid field (in our case a grid) and it’s wrapping the grid's content with “xhtml” at the beginning and end. 

I think this may have to do with how grid field data is stored, and that EE’s excerpt tag is stripping away html from the grid field that is exposing “xhtml”.

One solution could be changing the channel settings so that we can target a search excerpt field that's within a fluid field, that's also the first of that field type in the entry. So the first textarea, but not the second or third.

What we're doing as a workaround at the moment is using the streeng addon to remove the two "xhtml" instances bookending the search results.

### LINKED ISSUE #458: Variable modifiers don't work in search results
(I'm posting this as a bug, because I'm not sure whether variable modifiers are currently *expected* to work in search result tags. If they're not expected to work, please read as a feature request!)

Variable modifiers appear not to currently work with search result tags. When a variable modifier is used with a tag, the tag is not parsed.

**How To Reproduce**
For example, using the `:limit` variable modifier to shorten the search excerpt just results in an unparsed tag:

```
{exp:search:search_results}
   {excerpt:limit characters='50'}
{/exp:search:search_results}

```
Another example (and the actual use-case which hit this error) is using `:replace` to modify the search result URL (in this case to remove the protocol, in order to provide a cleaner visual):

```
{if page_url}
   {page_url:replace find='/^http(s?):\/\//i' replace='' regex='yes'}
{if:else}
   {auto_path:replace find='/^http(s?):\/\//i' replace='' regex='yes'}
{/if}

```

In this case, the `{page_url}` tag is parsed correctly, and the `:replace` variable modifier works as expected (presumably because this tag is parsed by the Pages module itself, and not by the search module?). But the `{auto_path}` tag is unparsed. *Without* the variable modifier, the `{auto_path}` tag itself is fine.

**Environment Details:**
 - Version: 5.3.2
 - PHP Version: 7.0.33
 - OS: ubuntu 16.04.15
 - Web Server: Apache 2

### LINKED ISSUE #1085: [EE6] Search results excerpt does not take account of expansion of template variables
Where the excerpt from a channel entry shown in search results contains a template variable, this variable is expanded/replaced *after* the search excerpt is generated. This is consistent with the template rendering order shown at https://docs.expressionengine.com/latest/templates/engine.html, but can create some inappropriate effects:

- The resulting excerpt can be significantly longer than the documented excerpt length of 50 words.

- The resulting excerpt will retain any markup present in the template variable. This can cause unintended presentation effects when displaying search results, and contradicts the documentation, which states that "HTML markup is stripped prior to output".

Notwithstanding the standard EE template processing order, search excerpts should be generated — including limiting to 50/n words, and stripping markup — *after* all EE tags have been rendered, including template variables.

EE 6.0.4

### LINKED ISSUE #1851: [EE5/6?] Issue in (simple) search: some custom field/s not searched when search keywords are quoted?
(This is a tricky one to characterize and reproduce. Also, I'm describing what I see in EE5.4.3. I'm not in a position right now to test in EE6, but I don't see anything in the EE6 change logs which directly refers to the issue. So, please read this as something I'm seeing in EE5, which might also be happening in EE6, but needs to be confirmed.)

I'm writing some code to post-process (simple) search results, and I hit some behaviour in simple search results which looks wrong. Specifically, when search terms are quoted, some search hits which should be included are missing, and this might relate to a failure to search all custom fields. To help diagnose, I've exported some of the SQL queries generated by the search module.

A simple example. Searching for a specific last name, **mcguire**, results in 3 hits. One of these occurrences is in a (searchable) custom field. Repeating the search with a quoted search term, **"mcguire"**, results in 2 hits; the missing hit in this case is the one in the custom field. The two SQL search queries generated by these searches are attached. Note that the second (for the quoted search term) is significantly smaller than the first. Specifically, note that the relevant custom field, **326**, is present in the SQL query for the unquoted search term, and missing from the SQL query for the quoted search term.

This behaviour looks to be the same for more complex searches. For example, a search for a specific **firstname lastname** includes the custom field when the search term is unquoted, but doesn't include the custom field when the search term is quoted, as **"firstname lastname"**.

Exported SQL queries follow.

SQL query when searching for **mcguire**:
[search-queries-mcguire-noquotes.txt](https://github.com/ExpressionEngine/ExpressionEngine/files/8149819/search-queries-mcguire-noquotes.txt)

SQL query when searching for **"mcguire"**:
[search-queries-mcguire-quotes.txt](https://github.com/ExpressionEngine/ExpressionEngine/files/8149820/search-queries-mcguire-quotes.txt)

### LINKED ISSUE #1862: [EE5/6] Search module should use results parameter when pagination isn't applied
When results aren't paginated, the search module is hard-coded to return at most 100 hits (see below), and the `results` parameter is ignored.

There's no good reason for this. In cases where pagination isn't being applied, it would make sense to use the existing `results` parameter as a maximum. A hard-coded default if the `results` parameter isn't present would make sense, but the `results` parameter should override this — whether larger or smaller than the default.

Screen Shot 2022-03-01 at 11 56 15 AM

### LINKED ISSUE #1869: [EE6] Simplistic sanitization of search terms produces unexpected/inappropriate/no results
The EE6 search module "sanitizes" search terms using a very simple strategy of replacing certain non-alphanumeric characters with spaces. This appears in `/system/ee/legacy/helpers/search_helper.php` (screenshot below).

The effect of this sanitization is (obviously) to create a modified search term, which potentially gives unexpected results. This is especially the case when a search term is quoted — the typical meaning of which is that the search term is intended to be used exactly.

A real example. The name of a real project on a site that I manage uses an exclamation mark in place of the letter i, in the form **abc!def** (not the real title, but the same form). As a search term, this is modified to **abc def**. When used unquoted, this will search separately for the terms **abc** and **def**, returning inappropriate results, which don't match the original search term. More seriously, when quoted, the search term is modified to **"abc def"**. This modified search term is then searched exactly, and (incorrectly) returns no search results at all. The effect is that this perfectly innocuous project title cannot be searched for.

This behaviour is sufficiently inappropriate/unexpected that it should be considered a bug. It's not reasonable to expect that a typical user will understand that the search term has been modified, resulting in an entirely different search being performed. Using quoted search terms has a specific common expectation in search, and this behaviour of EE violates that expectation.

It's not unreasonable for EE to sanitize search terms if they represent a genuine security threat. But a much less simplistic algorithm is needed, which doesn't modify search terms — especially when quoted — unless absolutely necessary. This should be managed within EE, and not be made the user's problem.

Screen Shot 2022-03-04 at 12 01 32 PM

intoeetive:
Resolved #83 where 'xhtml' string could mistakenly appear in Fluid field search excerpt

Fluid is now respecting each field's "searchable" preference and not including non-searchable fields into excerpt (same as Grid does)

Resolved issue where undesired characters were shown in Grid field excerpt in search

Probably not directly resolves, but related to #767 - with the "searchable" setting for field being respected and "|" stripped from Grid content, the output looks much better, so we might consider it resolved

Resolved #458 where variable modifiers were not fully available in search results

Resolved #1085 where having variables in searchable fields could break search results

Resolved #1851 where some custom fields could be excluded from search 

The solutions here is to only fetch searchable fields for the channels that are searched. This should significantly reduce the size of query for most of the sites.
The bigger sites (50+ searchable fields on channel) would still have the issue - we will write warning to developer log, because searching 50+ fields without index is really bad for performance anyway

Resolved #1869 where unnecessary search term sanitization was performed making the results inaccurate

Resolved #1862 where search results could not display more than 100 entries when not using pagination

Resolved #66 where sorting search results was limited; closes #2853


matthewjohns0n (approved):

robinsowell:
Noting I tested the issue #1085 I don't love stripping globals entirely, but don't have a better idea.  Like full_text still includes them.

PR #2760, opened 2023-01-16
Resolved issues with cache driver not being used

bryannielsen:
There are parts of the core that override the user's choice of cache driver and require the file cache.  This is confusing and particularly problematic in serverless environments where the file driver is not useful.  This change attempts to standardize more functionality and always rely on the user selected cache driver.

### DIFF
--- a/system/ee/legacy/libraries/Cache/Cache.php
+++ b/system/ee/legacy/libraries/Cache/Cache.php
@@ -190,6 +190,20 @@ class Cache extends EE_Driver_Library
     }
 
     /**
+     * Checks whether cache key is writable
+     *
+     * @return	bool
+     */
+    public function is_writable($key = '', $scope = Cache::LOCAL_SCOPE)
+    {
+        if(method_exists($this->{$this->_adapter}, 'is_writable')) {
+            return $this->{$this->_adapter}->is_writable($key, $scope);
+        }
+
+        return true;
+    }
+
+    /**
      * Returns the name of the adapter currently in use
      *
      * @return	string	Name of adapter
--- a/system/ee/legacy/libraries/Cp.php
+++ b/system/ee/legacy/libraries/Cp.php
@@ -390,7 +390,7 @@ class Cp
         $notices = array();
 
         // Show a notice if the cache folder is not writeable
-        if (! ee()->cache->file->is_supported()) {
+        if (! ee()->cache->is_writable() && ee()->cache->get_adapter() == 'file') {
             $notices[] = lang('unwritable_cache_folder');
         }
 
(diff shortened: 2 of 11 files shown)


system/ee/legacy/libraries/Cp.php
```diff
@@ -390,7 +390,7 @@ protected function _notices()
         $notices = array();
 
         // Show a notice if the cache folder is not writeable
-        if (! ee()->cache->file->is_supported()) {
+        if (! ee()->cache->is_writable() && ee()->cache->get_adapter() == 'file') {
```
intoeetive, line 393: I am not sure about this change.

The docs are [requiring](https://docs.expressionengine.com/latest/installation/installation.html#3-set-file-permissions) `/user/cache folder` to be writable, and it has always been like that

system/ee/legacy/libraries/Cache/Cache.php
```diff
@@ -189,6 +189,20 @@ public function is_supported($driver)
...
     }
 
+    /**
+     * Checks whether cache key is writable
+     *
+     * @return	bool
+     */
+    public function is_writable($key = '', $scope = Cache::LOCAL_SCOPE)
+    {
+        if(method_exists($this->{$this->_adapter}, 'is_writable')) {
```
intoeetive, line 199: needs a space after if

intoeetive (changes requested):
Please see comments on code.

Also noting that we need to actually test this when using driver different than `file`

system/ee/legacy/libraries/Cp.php
bryannielsen: This will just always throw a warning if EE is deployed to a serverless host where system/user/cache cannot be writable.  Another option would be to allow the cache folder to be configurable.  Usually the `/tmp` folder is writable even in these environments but still the file cache driver doesn't work because the machine is destroyed after every request.

intoeetive: So does that mean you can set cache driver to "dummy", have `system/user/cache` folder NOT writeable and EE still would work as expected (thinking of query caching, tag caching, template caching)? 
I think we need to actually test that before making this change

bryannielsen: Yes this definitely needs some testing, I just wanted to get the PR open so we could start that process.  I think we should target a later release for this after it has been tested thoroughly.

EE should still function with any cache driver set otherwise I would think that's a fault with the caching system.  In my opinion caching shouldn't be necessary but should provide a performance benefit when enabled.  Anyone choosing to use the `dummy` driver should still have a functioning site but gaining no performance benefit.

intoeetive: The cache directory (`PATH_CACHE`) is also being used for other things (exporting, running 1-click updates etc.). If it's not writable, EE will not function correctly. If some environments can't make it writable, then we need a fix for those cases - not removing the check.

I'm fine with other changes, but not this one

bryannielsen: Yea I did a search on PATH_CACHE and it looks like it's being used in a lot of places to write files where we should really refactor to using the cache driver.  

Updates are something we'll need to think about, they aren't really possible in a serverless environment without a persistent filesystem mount.  So perhaps this warning should be tied to if 1-click updates are enabled in the config?  We really just need some way for people running EE on a serverless environment with a read-only disk to not see this warning on every page load.

PR #2795, opened 2023-01-30
Resolved #2765 where using `ee()->table->set_data()` was not possible outside of Control Panel

### LINKED ISSUE #2765: Table Class set_data method throwing error
I'm writing an add-on that outputs a table of data to the user. It is not for the CP, but it throws the following error:

No such property: 'cp' on ExpressionEngine\Legacy\Facade

ee/ExpressionEngine/Legacy/Facade.php:128

The code is in my mod.addon.php file in a method called results. Nothing fancy.

I removed my code and inserted the demo code from the [Table Class](https://docs.expressionengine.com/latest/development/legacy/libraries/table.html#example-improving-usability-for-large-amounts-of-data) docs into my method and it throws the same error. If I remove just the set_data call, it will output the table with the headers as expected. I'm not sure if it is intentional that the class only be used in the CP, but it seems like this should work in a module method as well.

Using EE 7.2.6

intoeetive:
Resolved #2765 where using `ee()->table->set_data()` was not possible outside of Control Panel

EE6 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/2794

### DIFF
--- a/system/ee/legacy/libraries/Table.php
+++ b/system/ee/legacy/libraries/Table.php
@@ -37,10 +37,12 @@ class EE_Table
     public $empty_cells = "";
     public $function = false;
 
+    private $temp;
+
     /**
      * Constructor
      *
-     * @access	public
+     * @access  public
      */
     public function __construct()
     {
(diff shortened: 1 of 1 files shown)


matthewjohns0n (approved):
Looks good to me 👍🏼

PR #2797, opened 2023-01-30
Resolved #2787 where missing fieldtype would prevent adding new field or entry

### LINKED ISSUE #2787: Unable to create new field if (other) fieldtype is missing
This happens if fieldtype was not correctly uninstalled (e.g. there's still field assigned to this type in `exp_channel_fields`) but the addon files were deleted

If you try to create new field, you get the error
> Error: Unable to load requested fieldtype file: ft.cypress_addon.php.
> Confirm the fieldtype file is located in the /system/user/addons/ directory

Same issue exists if you go to publish new entry - though in that case it might be expected, but only if the field is assigned to given channel

This should be silently ignored instead

intoeetive:
Resolved #2787 where missing fieldtype would prevent adding new field or entry

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Fields/Fields.php
+++ b/system/ee/ExpressionEngine/Controller/Fields/Fields.php
@@ -12,6 +12,7 @@ namespace ExpressionEngine\Controller\Fields;
 
 use ExpressionEngine\Controller\Fields\AbstractFields as AbstractFieldsController;
 use ExpressionEngine\Model\Channel\ChannelField;
+use ExpressionEngine\Error\AddonNotFound;
 
 /**
  * Fields Controller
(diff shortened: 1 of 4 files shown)


bryannielsen:
I tried testing this with Structure by setting up a Structure field, and creating an entry that stored data in that new field.  Then I renamed the structure addon and I get a few errors about the extension not being found, and then the publish tab.  I wonder if we should also apply this treatment to those areas?

intoeetive:
@bryannielsen we definitely need to - but the treatment for extensions and tabs will be different. Please create different issue and we'll work on it

bryannielsen:
There are a few more scenarios where the error is still thrown.  Moving this to a future patch so we can look into a more complete fix

intoeetive:
I am bringing this back into consideration.

The errors on Publish page, or other relevant pages, would be expected - because those would be using the functionality from add-on removed. With the fieldtype removed, there is no reason for "create field" page to error out, because it's not using the removed fieldtype.

So this PR is specifically addressing thatr\ specific issue.

bryannielsen (approved):

PR #2804, opened 2023-01-31
Resolved #2087 to allow for class parameter in `{form:custom_profile_field}` tag

### LINKED ISSUE #2087: In Edit Profile Data Tag -> Custom Profile Field Variable Pair How can i add class to {form:custom_profile_field}
{custom_profile_fields}
        <div class="form-group">
            <label>{lang:profile_field}</label>
            {form:custom_profile_field}   **--> I need to add class to this field like form-control**
        </div>
    {/custom_profile_fields}

Thank You..

jHards:
Allow for `{form:custom_profile_field}` to take an `input-class` parameter to be added as a class to the resultant `<input>` or `<select>` elements. Likely simple to add for other field types but not attempted here.

Also added variables:
- `{lang:profile_field_clean}` to return label without required prefix
- `{lang:profile_field_name}` to return field name returned by `{form:custom_profile_field}`
- `{lang:profile_field_shortname}` to return field shortname

Resolves https://github.com/ExpressionEngine/ExpressionEngine/issues/2087

Documentation to be updated - https://docs.expressionengine.com/latest/member/edit-profile.html#custom-profile-field-variable-pair

Anticipated usage:
```
{custom_profile_fields}
     <div class="form-group">
          <label for="{lang:profile_field_name}" class="col-sm-4 control-label">{lang:profile_field_clean}:</label>
          <div class="col-sm-6">
              {form:custom_profile_field input_class="form-control"}
          </div>
      </div>
{/custom_profile_fields}
```

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member_settings.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member_settings.php
@@ -583,7 +583,6 @@ class Member_settings extends Member
             /**  Parse single variables
             /** ----------------------------------------*/
             foreach ($this->var_single as $key => $val) {
-
                 // Custom member fields
                 $field = ee('Variables/Parser')->parseVariableProperties($key);
                 $fval = $field['field_name'];
@@ -813,7 +812,7 @@ class Member_settings extends Member
                 $temp = str_replace('{lang:profile_field_shortname}', $field->getShortName(), $temp);
                 $temp = str_replace('{lang:profile_field_description}', $field->get('field_description'), $temp);
                 if ($match[4] != '') {
-                    switch($field->get('field_type')) {
+                    switch ($field->get('field_type')) {
                         case "text":
                             $temp = str_replace(
                                 $match[0],
@@ -981,7 +980,7 @@ class Member_settings extends Member
 
                 // Handle arrays of checkboxes as a special case;
                 if ($row['m_field_type'] == 'checkbox') {
-                    foreach ($row['choices']  as $property => $label) {
+                    foreach ($row['choices'] as $property => $label) {
                         $member->$fname = in_array($property, $post) ? 'y' : 'n';
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


jHards:
@matthewjohns0n any update?

jHards:
@matthewjohns0n / @intoeetive any update?

intoeetive:
@jHards thank you for this code submission!

We're planning a number of updates related to Member functionality in EE 7.4, and this is where we'll care of this PR. 

The approach  to custom member fields will be different (we plan to support more different field types for members), so we probably will not be able to accept this PR in its full - but we'll likely take some parts of it.

Thank you for your patience!

jHards:
Thanks for the update!

intoeetive:
I am closing this in favor of https://github.com/ExpressionEngine/ExpressionEngine/pull/2844/commits/aa92dfab3270352667b662958da59c99245225e3 which will become part of member functionality update in EE 7.4
Thank you!

jHards:
> I am closing this in favor of [aa92dfa](https://github.com/ExpressionEngine/ExpressionEngine/commit/aa92dfab3270352667b662958da59c99245225e3) which will become part of member functionality update in EE 7.4 Thank you!

Thanks for this @intoeetive, certainly a more elegant solution! I see you have introduced `{display_form}` which looks to mimic `{form:custom_profile_field}` for now. Are you still planning to allow for a class to be passed to the input element as intended here - this was key for my use case.

Also, I was hoping to include the ability to choose which custom member fields are picked up here, my understanding of the current `{custom_profile_fields}{/custom_profile_fields}` pair tag is that it will return all of them with no way to filter or order?

intoeetive:
For the time being, we're setting this to mimic the behavior of Channel Form tag https://docs.expressionengine.com/latest/channels/channel-form/overview.html#custom_fields

Thus some variables are duplicating the variables that exist under different name (`{display_form}` vs `{form:custom_profile_field}`). There is no way to pass in what fields you need into `{custom_profile_fields}` - but you can add conditionals to check what fields you want (check againt field name or type, e.g. `{if field_name == 'something'}`), as well you should be able to place the field directly using `{field:field_name}` syntaxt (that part is still under development)

To answer your other question, we're not giving a way to add class to input directly. This is due to the complexity of some fieldtypes, which might consist of several inputs, so there is no reliable way to do that in a "generic" way. You can always add class or ID to the outer HTML wrapper for field though!

jHards:
Thanks, I should be able to make that work with the fields now available and some conditionals. Looking forward to trying it out!

PR #2805, opened 2023-01-31
Resolved #2639 where animation was removed from GIF files during upload

### LINKED ISSUE #2639: Using EE Files system to upload animated gif removes animation from image
As per title.
Using either the File / Upload or in-field upload options has same result - animated GIF images are collapsed to static images during the upload process.
This is a change in behaviour from EE5 - it cropped up on a site that makes extensive use of animated gifs, after updating to EE6 they no longer could add new content.

There is a work-around - upload the images directly to the Files folder on the server and then "sync" the images directory from within EE and then attach the found images to the entries concerned, but this is complicated to set up / execute for the design agency concerned... so better would be to fix the issue within EE.

EE 6.4.2 // php 8.0.25

intoeetive:
Resolved #2639 where animation was removed from GIF files during upload

### DIFF
--- a/system/ee/legacy/libraries/Filemanager.php
+++ b/system/ee/legacy/libraries/Filemanager.php
@@ -516,7 +516,7 @@ class Filemanager
             $prefs['file_width'] = $prefs['width'];
         }
 
-        if ($prefs['max_width'] == 0 && $prefs['max_height'] == 0) {
+        if (empty($prefs['max_width']) && empty($prefs['max_height'])) {
             return $prefs;
         }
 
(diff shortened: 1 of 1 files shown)


Yulyaswan (approved):

bryannielsen:
I wasn't able to reproduce the issue.  Uploading animated gifs to an image-only file destination worked fine for me in 6.dev and also in this patch branch.

bryannielsen:
I was able to reproduce this on PHP8+ and the fix is good.  Also just noting that this was not an issue on PHP 7

PR #3031, opened 2023-02-21
Added Members fieldtype to select one or multiple members in channel entry; #1080

### LINKED ISSUE #1080: Members as options
Last year, I did a website for journalists and they asked me to have more than one author for some articles.

It would be great to [populate a select field type](https://docs.expressionengine.com/latest/fieldtypes/select.html#populate-from-channel-field) from a members list.

Even better, would be to be able to use a field group as members custom fields. Or to have a default pre-created channel for members. Both solutions, I guess, are harder to implement.

After to choose a path, we can write a documentation / specification.

intoeetive:
Added Members fieldtype to select one or multiple members in channel entry; closes #1080

Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/601


robinsowell:
@bryannielsen This tested out for me fine.

matthewjohns0n:
I'm having issues with this one - when I filter by Role, it then can become impossible to sort by a different role. I took a video so it's more clear.

https://github.com/ExpressionEngine/ExpressionEngine/assets/11818941/6c206fc5-295a-4951-a6d5-89ba796c6859

intoeetive:
This seems to only be an issue when you have more role than the members that are currently displayed.

Since we're using Relationship component here - I suggest that you just create new issue (but also, you can "fix" it by just scrolling down to the bottom of page) @matthewjohns0n

matthewjohns0n:
I set up this field as the bottom field, so I was unable to scroll to the bottom of the page (already there) so that fix didn't work for me. It's probably fine to open an issue for relationship component here and approve this.

litzinger:
Looks like a useful feature, however, after reviewing the ft.member.php and ft.relationship.php files they're almost identical. Why not extend Relationship_ft and add a little abstraction since this is basically a Relationship field anyway? 

Being a little selfish... I noticed too that this doesn't call any of the relationship hooks, so I can't add any support for this in Publisher. :(

intoeetive:
@litzinger thank you for pointing this out!

We can indeed make the code more DRY and add more extension hooks.

The code is [here](https://github.com/ExpressionEngine/ExpressionEngine/pull/4022/files) and the documentation is [here](https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/601/files#diff-1500e2d94c161c9af981d077352537606f367d52586ac33421a3b68559bea193)

We only added the hooks that were in actual fieldtype file (as relationship parser is not involved here) - are there any other hooks that you need to get added in order to make Publisher work?

litzinger:
That looks wonderful, thanks for considering this change. As for the hooks, I think the crucial ones were what was in the ft file. The others probably won't be necessary.

PR #3033, opened 2023-02-21
ExpressionEngine 6.4.6

intoeetive:
This PR is updating PHPUnit tests (except related to CLI addons generation) to what they are in EE7

As part of this, I also had to make some updates that are already done in EE7, most notable updates to Validation service.

There might be things that have been missed, but the tests are passing now

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,17 +1,11 @@
 name: PHP Unit Tests
 
 on:
-  push:
-    branches:
-      - feature/6.x/unit-tests
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
-      - 6.dev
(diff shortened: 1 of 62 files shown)


bryannielsen (approved):

.github/workflows/unit-tests.yml
```diff
@@ -1,17 +1,11 @@
...
-      - feature/6.x/unit-tests
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
-      - 6.dev
-      - 6.2.0
-      - 6.3.0
-      - 6.4.0
-      - 6.5.0
+      - '*.dev'
```
matthewjohns0n, line 7: @intoeetive Is this okay, or should it be '6.dev'. I'm not really sure if it makes a difference.

PR #2659, opened 2022-12-20
Resolved #2658 where some parameters were not available for member management forms

jHards:
## Overview

Makes Member Management front-end forms more consistent, per feature request. `{exp:member:edit_profile}` is hardcoding the form `id="cform"` and this has been left as-is; otherwise all form tags referenced can now accept `form_class`, `form_name` and `form_id`. Suggest no reason that this isn't made available in v6 too.

Resolves [#2658](https://github.com/ExpressionEngine/ExpressionEngine/issues/2658).

## Nature of This Change

- [ ] 🐛 Fixes a bug
- [x] 🚀 Implements a new feature
- [ ] 🛁 Refactors existing code
- [ ] 💅 Fixes coding style
- [ ] ✅ Adds tests
- [ ] 👽 Adds new dependency
- [ ] 🔥 Removes unused files / code
- [ ] 🔒 Improves security 

## Is this backwards compatible?

- [x] Yes
- [ ] No

PR #3036, opened 2023-02-22
Resolves memberlist form attribute issue that remained after #2659/#2689

jHards:
Previous PR only addressed Member Search Form tag and not Memberlist tag (an oversight). Also updated action id to 'member_search' for both tags, as the current 'do_member_search' appears to be a legacy entry.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member_memberlist.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member_memberlist.php
@@ -128,7 +128,7 @@ class Member_memberlist extends Member
         /** ---------------------------------
         /**  Does the recipient accept email?
         /** ---------------------------------*/
-        $query = ee()->db->query("SELECT email, screen_name, accept_user_email FROM exp_members WHERE member_id = '". ee()->db->escape_str($member_id) . "'");
+        $query = ee()->db->query("SELECT email, screen_name, accept_user_email FROM exp_members WHERE member_id = '" . ee()->db->escape_str($member_id) . "'");
 
         if ($query->num_rows() == 0) {
             return false;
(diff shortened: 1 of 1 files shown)


intoeetive:
@jHards thank you for this PR!

I am not sure that changing one action to another is the thing we'd want to do however - both seems to be functional (but somewhat different) with both legacy and regular templates

jHards:
@intoeetive am I reading the failed test correctly in that it's an issue with the test rather than this PR?

intoeetive:
The test that's failing is `member_list`, so that must be related to the actual changes in code

jHards:
The test is looking for some table tags that it can't find but I wasn't expecting my changes to affect that - what tag is the test trying to load?

intoeetive:
Do you have access to the test run summary page? There's artifacts sections there, where you can download the zip with the screenshots and videos 
https://github.com/ExpressionEngine/ExpressionEngine/actions/runs/4315756812

jHards:
Got it, thanks. Looks like the user the test script is logged in as doesn't have access to via public member profiles?

Member List frontend -- respects the options (failed) (attempt 2)

intoeetive:
It could be that, or it could be something completely different. Since the error is not clear typo or something that you can see from the screenshot, you would need to set up Cypress locally and run the tests to see yourself, and then figure out what's causing the failure

We have a small section on testing written: https://github.com/ExpressionEngine/ExpressionEngine/blob/7.dev/CONTRIBUTING.md#tests

intoeetive:
@jHards I realize you have dedicated a lot of effort to this PR, but looking into it as we come closer to EE 7.4 release, I don't think this is the change we want to include.

It's trying to submit data to search members in place where the code says member search should not be used. The current submission of select dropdowns in `exp:member:memberlist` tag works fine, so I'm going to close this without merging.

Thank you for all your contributions to members functionality in ExpressionEngine - while we don't always accept those, they help us to become better!

jHards:
@intoeetive Thanks for the explanation. Where I was going with this is that I needed to present a pre-filtered, pre-sorted set of member data but I didn't want to output the whole thing with the search form and let users do it themselves. Have I missed something in the docs or is this something that is in your 7.4 pack of member updates?

intoeetive:
@jHards the tag has `role_id`, `limit`, `orderby` and `sort` parameters (which are not documented - thanks for pointing out). If you need to have better options of filtering using tag parameters, please open a feature request. I also believe some of the add-ons are capable to display more granular member lists

jHards:
@intoeetive see [#3649](https://github.com/ExpressionEngine/ExpressionEngine/discussions/3649)

PR #3039, opened 2023-02-24
Resolved #3028 where categories tab drag-and-drop re-ordering was not working

### LINKED ISSUE #3028: [EE v7.2.9] +  [EE v6.4.5] Categories Tab Drag-and-Drop Re-Ordering Not Working
In any entry's Categories Tab enable the Manage Categories toggle. You'll now be able to drag and indent (create a child relationship) with a category, but you will not be able to reorder the categories. This is reproducible in both Chrome and FireFox on MacOS.

This broke right after selected categories began being displayed near the top of the list to help support sites with long lists of categories. Unfortunately, it means that the manage categories option within the Categories Tab is no longer a way to reorder the categories.

Our clients have all been trained to reorder pages in their website's primary navigation from within the Categories Tab, so we are eager to see this functionality return.

Once the Manage Categories Toggle is selected, I would expect the list of categories to return to the order that was set up by our clients regardless of which are selected.

Let me know if have any questions.

- Kurt

Yulyaswan:
Resolved #3028 where categories tab drag-and-drop re-ordering not working
EECORE-2192
EE7 version of #3040

### DIFF
--- a/cp-styles/app/styles/components/_checkboxes-and-radios.scss
+++ b/cp-styles/app/styles/components/_checkboxes-and-radios.scss
@@ -158,7 +158,7 @@ $radio-size: 15px;
 	margin-bottom: 0;
 	position: relative;
   font-weight: normal;
-  pointer-events: none;
+  // pointer-events: none;
 
 	.checkbox-label__text {
 		display: inline-block;
@@ -171,7 +171,7 @@ $radio-size: 15px;
 		transition: background 100ms ease;
 		color: color(text-primary);
 		border-radius: 5px;
-		pointer-events: auto;
+		// pointer-events: auto;
 
 		&:hover {
 			// background: color(bg-10);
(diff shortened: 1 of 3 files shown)


cp-styles/app/styles/components/_checkboxes-and-radios.scss
```diff
@@ -466,3 +466,17 @@ div[data-input-value="category_fields"] {
...
+
+// 		input {
+// 			pointer-events: auto;
+// 		}
+
+// 		.checkbox-label__text {
+// 			pointer-events: auto;
+// 		}
+// 	}
+// }
```
intoeetive: @Yulyaswan should this be removed completely?

Yulyaswan: exactly. removed

intoeetive (approved):

PR #3058, opened 2023-03-02
Resolved #3051 where Menu Manager was doubling rows when creating new Dropdown item

### LINKED ISSUE #3051: [EE v 7.2.10] Menu Manager Dropdown option doubling fields
I've found an annoying bug that has been present on previous builds as well. When you select a dropdown as the menu option type and add a menu entry the rows double. The first row is ignored for its values. If you add another it adds 2 more rows, only the second of the fields capture. 

Screenshot 2023-02-28 at 9 03 17 AM

Screenshot 2023-02-28 at 9 03 27 AM

Screenshot 2023-02-28 at 9 03 33 AM

Yulyaswan:
Resolved #3051 Menu Manager Dropdown option doubling fields

### DIFF
--- a/themes/ee/asset/javascript/src/cp/modal_form.js
+++ b/themes/ee/asset/javascript/src/cp/modal_form.js
@@ -73,12 +73,12 @@ EE.cp.ModalForm = {
 				options.load(that.modalContentsContainer)
 			})
 
-			var timer = setInterval(function() {
-				if ($('.app-modal .grid-field').length) {
-					new Grid.Publish($('.app-modal .grid-field'))
-					clearInterval(timer);
-				}
-			},50);
+			// var timer = setInterval(function() {
+			// 	if ($('.app-modal .grid-field').length) {
+			// 		new Grid.Publish($('.app-modal .grid-field'))
+			// 		clearInterval(timer);
+			// 	}
+			// },50);
 		}
 	},
 


intoeetive:
@Yulyaswan we added this code few month ago and now we are removing it. How can it be properly tested to make sure nothing gets broken?

Yulyaswan:
@intoeetive to check this, you have to check how the grid is initialized in a modal window. It looks like now the script loads faster and we don't need this setInterval

intoeetive:
I tested this in different modals and could not spot any issue, but I'm also a bit hesitant to approve because we don't know why the code was added initially. Maybe it's when working with very large Grids or something along those lines.

Asking @robinsowell to have a look

matthewjohns0n:
I dont feel like I'm able to review this properly, but I did find the previous commit:

https://github.com/ExpressionEngine/ExpressionEngine/commit/933e73a16dceea74729d956ad9662e9150cdadb4

Since its commit message is "init Grid inside modal" we may need to test how this works with Grid in a modal or something along those lines.

intoeetive:
On the team call we decided to move this to next milestone so we could test it better

matthewjohns0n (approved):
I tested that this fixes the issue. I also tested creating a grid in a modal, and it appears to work fine. 

I am approving, but I do not know the full extend of why this code was initially added, and if it will become and issue down the line. So far I cant find any issues.

PR #3060, opened 2023-03-02
Resolved #2775 where approved members could have Pending as secondary role

### LINKED ISSUE #2775: Approve pending member, they stay in pending
So in the CP, make a member, put them in pending.  Approve them.  They go into the correct default member group but they ALSO stay in pending.

And I would SWEAR this came up before, but I can't find it.

intoeetive:
Resolved #2775 where approved members could have Pending as secondary role

(cherry picked from commit a6e000295fab1087ec95ff4f9a64e77308133840)


robinsowell (approved):
works for me now in v7.

PR #3072, opened 2023-03-03
Improved Entry Date field to make it user-friendly to pick a different time, #3021

Yulyaswan:
Improved Entry Date field to make it user-friendly to pick a different time, #3021 
EECORE-2193

For the time block I'm using special HTML element `<input type="time">`
Entering hours and minutes in 12 or 24-hour format depends on **`operating system locale`** (which is essential to understand).
The final result of displaying in the date field will depend on the settings of the site system.


intoeetive (changes requested):
When I change just the time, nothing happens - I need to actually change the date in order for changed time to appear in the field. I think the change needs to happen on blur as well (or as-you-select)

Also if I enable seconds in settings, they show up in picker, but not in the field

intoeetive (changes requested):
@Yulyaswan I had saved entry with entry date set. When I open picker, it sets the time with some offset
image

intoeetive:
@Yulyaswan is this still draft?

Yulyaswan:
@intoeetive yes, I will try to finish before the next release.

robinsowell:
Just confirming, I can't replicate the issue Yuri ran into.  Very odd!  On my end, it all works fine.  Not sure how to reproduce the weird time offset.

intoeetive (changes requested):
@Yulyaswan timezone configured in EE is New York, while my local timezone is obviosly different.  I think this is the reason for behaviour

intoeetive:
Looks like when you select time and then date, the calendar selects wrong date (6/30 instead of 7/4 for me)

robinsowell:
Works well for me and I approved in testing.  But I will note- if you pick the date first, the modal closes and you have to go open it again to pick the time.  That's somewhat annoying, as if I have the time showing, I probably want to set it.  But, also not a deal breaker, so I approve.  Can always tweak later if others are annoyed.

matthewjohns0n:
I agree with Robin on this one - it is a bit annoying to use. It closes after selecting the date, and then you have to click out of the field to open it again.

intoeetive:
On the other hand, you're not forced to pick time, so it does not break the pre-existing user experience

TomJaeger:
So I would say it should only close when you click OFF / out of the date&Time picker as a whole.... So you should be able to select date, not time.. and have it remain open till you click off it...OR you could update just the time or just the date, or both... and it won't close until you click out of it.

Yulyaswan:
@robinsowell @TomJaeger I changed the script
Could you please check, is it like you wanted?

PR #3122, opened 2023-03-15
Fixed #3120 where duplicate status options with large number of statuses in cp

### LINKED ISSUE #3120: Duplicate status options with large number of statuses in cp
On the channel admin page, status tab, in EE 7.2.11 if you have a LOT of statuses, open and closed will have duplicates.  I suspect it's when you have enough to trigger the search box.

In my case, I have 31 statuses, of which 2 are 'open' and 2 are 'closed.

URL is admin.php?/cp/channels/edit/1 and the status select looks like:

Screenshot 2023-03-14 at 2 37 59 PM

Only happens with a lot of statuses.

See also: https://github.com/ExpressionEngine/ExpressionEngine/issues/3112 which I suspect is the same issue.

Yulyaswan:
Fixed #3120 where duplicate status options with a large number of statuses in cp
EECORE-2206
EE7 version of #3170

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Channels/Channels.php
+++ b/system/ee/ExpressionEngine/Controller/Channels/Channels.php
@@ -804,7 +804,10 @@ class Channels extends AbstractChannelsController
             ->pluck('status_id');
 
         // Make sure open and closed are always selected
-        $selected = array_merge($selected, $default);
+        // And they are still haven't included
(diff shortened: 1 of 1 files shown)


intoeetive (approved):
Tested, works well.

@Yulyaswan can you also make the change for EE6, if it's not there already?

PR #3137, opened 2023-03-17
Resolved issue with sidebar member links

robinsowell:
Member sidebar was showing links they didn't always have access too, resulting in errors.


intoeetive (dismissed):
@robinsowell the changes look good, but can you please put attention to PSR-12 notices here? (turning on PHPCS in code editor might also be helpful)

Also it might be nice to make the individual permission checkboxes only show up when the parent toggle is on, as part of this PR

robinsowell:
@intoeetive I think it's good to go now!  Just needs approvals.

matthewjohns0n (approved):

PR #3141, opened 2023-03-20
Resolved #3135, #3047 where the columns in Entry Manager did not dynamically update if some column is not available

### LINKED ISSUE #3047: [EE v7.2.10] Entry Manager Table Column Select Menu missing "Comments" option
EE v6.x.x includes the Comments column menu option so that it may be deselected and removed as a column in the Entry Manager Table at /cp/publish/edit

That menu option is missing in EE v7.2.10 but the column for Comments still shows up. The menu opens over that column, so you can't see it in the screenshot, but it is under there.
 
Screenshot 2023-02-27 at 4 37 38 PM

If I add a new column to the table, the Comments column will go away, but it's not listed in the menu, so there's no way to add it back later if I change my mind.

It could be that EE is not listing Comments it in the menu because we have Comments off for all channels, but if that's the case, then the Comments column shouldn't have been displaying in the table either.

If this isn't a bug, it's very confusing behavior.

### LINKED ISSUE #3135: CP Entries filter, Comment column can be non-removable
So by default, the Comments column shows on the entry filter.  Make sure it's showing- go to Settings- Comment Settings and disallow comments.  Go back to your entries edit page- the comment column still displays.  Click the controller so you can remove it and... Comments is no longer avaible as a checkbox option.

Once you click something else- like Status- and load the page - Comments will disappear.

I think the column is getting stored as one to show and when you disable them, it keeps showing.  Which I don't think is a problem.  But when you disable them, they no longer show in the checkboxes so now you're stuck with the column showing until you resave the filters, at which point it looks like it gets washed out.

IDK how hard it would be to either clear it from the saved columns when you change the allow comments setting or- and I kinda prefer this- IF the comment column is showing, then the checkbox should still be included.   

In any case, it confused a user because they couldn't get rid of it.  EE 7.2.11

intoeetive:
Resolved #3135 where the columns in Entry Manager did not dynamically update if some column is not available

fixes #3047

EE6 version of #3140

### DIFF
--- a/system/ee/ExpressionEngine/Library/CP/EntryManager/ColumnFactory.php
+++ b/system/ee/ExpressionEngine/Library/CP/EntryManager/ColumnFactory.php
@@ -84,9 +84,16 @@ class ColumnFactory
      */
     private static function getStandardColumns()
     {
-        return array_map(function ($identifier, $column) {
-            return self::getColumn($identifier);
-        }, array_keys(self::$standard_columns), self::$standard_columns);
+        return array_filter(
+            array_map(function ($identifier, $column) {
+                if ($identifier != 'comments' || bool_config_item('enable_comments')) {
+                    return static::getColumn($identifier);
+                }
+            }, array_keys(static::$standard_columns), static::$standard_columns),
+            function ($column) {
+                return (! empty($column));
+            }
+        );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Yulyaswan (changes requested):
https://user-images.githubusercontent.com/23382425/227875110-3b4ef1ad-db3e-4e48-a100-3354e5271420.mov

@intoeetive looks like it doesn't work here. Could you please check the video?

intoeetive:
@Yulyaswan added code that was missing - please review/test again

Yulyaswan (approved):

PR #3153, opened 2023-03-22
Raise required MySQL version to 5.6.4 in order to support InnoDB engine for Pro Search indexes

intoeetive:
Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/602


robinsowell:
@intoeetive are we going to be able to do this for v6 as well, or is the db requirement change a deal breaker for that on?

intoeetive:
For v6, this is not needed, as Low Search is separate add-on there. We can make same change for Low Search, if we decide that we need to though

PR #3167, opened 2023-03-29
added new needed_fields_only paramater to the channel entries loop

TomJaeger:
This was added to get around the error of to many columns as the result of a query in the channel entries loop.

`SQLSTATE[0A000]: Feature not supported: 140 Aria table '#sql-temptable' has too many columns and/or indexes and/or unique constraints.`

It is not on by default but requires the use of a new channel entries parameter.

`needed_fields_only="yes"`

While by default this is how fields in the new storage method have operated, it's probably not best to make such a large change to the data available to the channel:entries loop for all of the legacy sites.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
+++ b/system/ee/ExpressionEngine/Addons/channel/mod.channel.php
@@ -2158,7 +2158,6 @@ class Channel
             $sql .= ", wd.entry_id, wd.site_id, wd.channel_id";
 
             if(!empty($legacy_fields)) {
-
                 foreach($legacy_fields as $lField) {
                     $sql .=', wd.field_ft_'.$lField->field_id;
                     $sql .=', wd.field_id_'.$lField->field_id;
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/channel/mod.channel.php
```diff
@@ -2088,8 +2088,16 @@ public function generateSQLForEntries(array $entries, array $channel_ids)
...
-                        m.username, m.email, m.screen_name, m.signature, m.sig_img_filename, m.sig_img_width, m.sig_img_height, m.avatar_filename, m.avatar_width, m.avatar_height, m.photo_filename, m.photo_width, m.photo_height, m.role_id, m.member_id,
+                        m.username, m.email, m.screen_name, m.signature, m.sig_img_filename, m.sig_img_width, m.sig_img_height, m.avatar_filename, m.avatar_width, m.avatar_height, m.photo_filename, m.photo_width, m.photo_height, m.role_id, m.member_id";
+
+        // check if we have param for needed fields only.. default is no
+        if (! $needed_fields_only = ee()->TMPL->fetch_param('needed_fields_only')) {
+
+            // string is weird on this one.. but it keeps the query
+            // well formatted for viewing when fully rendered
+            $sql .="
                         wd.*";
```
liamcrean, line 2098: Missing Comma?

TomJaeger: Good catch..added in!

matthewjohns0n (approved):

PR #3196, opened 2023-04-03
Resolved #3099, #2178 where existing entries did not show up in Entry Manager if the user did not have editing permission on another channel

### LINKED ISSUE #2178: Pagination in CP works even when there is no entries to show
Here is how newly registered author sees the CP:

image

There are no entries to edit, but pagination suggests that there are plenty of them.

### LINKED ISSUE #3099: Channel access: View entries, by others
### Discussed in https://github.com/ExpressionEngine/ExpressionEngine/discussions/3091

<div type='discussions-op-text'>

<sup>Originally posted by **shbchk** March  7, 2023</sup>
Not sure if this is a bug, but there is some inconsistency for sure.

1. Members with the role "OwnEntriesEditors" can edit only entries they created. In the entry list, they can see their own entries only.
2. But pagination suggests that there are a lot of other entries https://github.com/ExpressionEngine/ExpressionEngine/issues/2178
3. And they can see entries created by others in the Relationship fields (if it's enabled).

I wonder if we add an additional action **"View entries, by others"** here?
2c56916e-2591-4b5b-ad80-4a8e3fd3bf0a

</div>

intoeetive:
Resolved #3099 where existing entries did not show up in Entry Manager if the user did not have editing permission on another channel; closes #2178

closes https://github.com/ExpressionEngine/ExpressionEngine/discussions/3091

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/Edit.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/Edit.php
@@ -94,10 +94,6 @@ class Edit extends AbstractPublishController
         }
         $columns = array_filter($columns);
 
-        if (! ee('Permission')->hasAny($this->permissions['others'])) {
-            $entries->filter('author_id', ee()->session->userdata('member_id'));
-        }
-
         $count = $entry_listing->getEntryCount();
 
         // if no entries check to see if we have any channels


intoeetive:
now for the user with these permissions

image

the entry manager looks like this

image

matthewjohns0n:
Not sure if I should just open a new issue, but I noticed while testing this that if set the "Create entries" permission but not the "Edit own entries" permission, I get the `Error: You are not authorized to access this page` error message.

matthewjohns0n:
Another thing I noticed was in this configuration, users can still edit their own entries by going to the URL to edit the entry. Similarly, after saving the entry, it drops them on the edit screen, which is a typical default screen, but not when the user doesn't have access to edit their own entries. I'm creating a separate issue for that since it doesn't seem related.

intoeetive:
> Not sure if I should just open a new issue, but I noticed while testing this that if set the "Create entries" permission but not the "Edit own entries" permission, I get the `Error: You are not authorized to access this page` error message.

I think this might be expected. If you don't have any edit permissions, you don't get to see entry listing. Unless you're talking about the link still being present somewhere?

Yulyaswan (approved):

matthewjohns0n (approved):

PR #3223, opened 2023-04-05
Enabled File field to be used as Pro Variable; #2498

### LINKED ISSUE #2498: Pro variables file field portability
Per report: 
If you save a link to an image using a stand-alone Low Variable the value chosen is written to a file in the _variables template directory and the content comprises the full URL to the image.

If you put an image field into an LV grid and pick / save the same image, LV saves the filedir definition of the image path to the db.

What this means is when you move the site from a staging server to live server, any image links that are set using image fields in grid fields continue to work after the move, but any that were saved as single variables don’t (unless the staging server is visible to the live server).

intoeetive:
Enabled File field to be used as Pro Variable; closes #2498

Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/626

added `file` as reserved field name

------

This PR is allowing File fields to be used as Pro Variables. The variable always needs to be addressed with a pro_variables tag, not just with name.

Due to how "Select File" field is built (allows both single and multiple file selection) we can't just replace the fieldtypes, as data might get lost

This PR adds `file` as reserved word to prevent unexpected behavior when someone still tries to access File variable by name

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -165,6 +165,15 @@ class File_ft extends EE_Fieldtype implements ColumnInterface
     }
 
     /**
+     * Display the field for Pro Variables
+     *
+     */
+    public function var_display_field($data)
+    {
+        return $this->display_field($data);
+    }
+
+    /**
      * Return a status of "warning" if the file is missing, otherwise "ok"
      *
      * @return string "warning" if the file is missing, "ok" otherwise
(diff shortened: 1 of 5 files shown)


dougblackjr:
@intoeetive Love this one!

Question: Since `file` would now be a reserved word, is this backward compatible? Or will this cause breaking changes? We have a few legacy sites that have fields named `file`, just want to check.

intoeetive:
@dougblackjr I think this would not affect sites that have _field_ named `file`, but worth testing

We certainly don't intend to break anything!

intoeetive:
This change is not good, because it's enforcing all `{filedir_` and `{file:` tags to be parsed, even when those should not be parsed.
image

matthewjohns0n (approved):

PR #3235, opened 2023-04-06
Resolved #3222 where File icon in File field does not match field type

### LINKED ISSUE #3222: File icon in File field does not match field type
In the File Manager the icons for  files that are not images vary depending on file type - document, archive etc.

However when you select file in File field, it's displays the generic icon, without respecting file type 
image

Yulyaswan:
Resolved #3222 where File icon in File field does not match field type
EECORE-2213

### DIFF
--- a/cp-styles/app/styles/components/_file-field.scss
+++ b/cp-styles/app/styles/components/_file-field.scss
@@ -241,6 +241,7 @@
     padding: 10px;
     text-align: center;
     border-radius: $border-radius-base;
+    color: var(--ee-secondary-sidebar-item-icon);
 
     img {
       max-width: 200px;
--- a/themes/ee/asset/javascript/src/fields/file/control_panel.js
+++ b/themes/ee/asset/javascript/src/fields/file/control_panel.js
@@ -35,6 +35,10 @@
 			figure.toggleClass('is-svg', data.isSVG);
 			figure.find('img').toggleClass('hidden', ! data.isImage);
 
+			if (! data.isImage && figure.find('i').length) {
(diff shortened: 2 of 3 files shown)


intoeetive (changes requested):
@Yulyaswan I think we're missing `fa-file-video` here

intoeetive (changes requested):
@Yulyaswan if I select 'doc' file that's not in the list of mimes (in my case '.md' file) I get no icon at all. Probably an extra `else` is needed

Also, can we make it so that icon lines would be grey (like they are in file manager) and not black?

intoeetive:
@Yulyaswan I am so sorry :)

If I select a file, then change my mind, click "edit" pencil and select another file, I get 2 icons: 
image

Yulyaswan:
@intoeetive fixed :)

intoeetive (changes requested):
@Yulyaswan I still see this behaviour.
Steps:
- have a field with image in it
- edit the field data, select text file. The image icon removed, text icon added
- edit the field data, select zip fil. The zip icon is added next to next icon

Yulyaswan:
https://user-images.githubusercontent.com/23382425/231180074-8b0f5b4b-9063-4ad2-b500-ae623955adb9.mov

could it be cache? @intoeetive

intoeetive:
@Yulyaswan it could be, but it is not :)
Use the "edit" button (pencil), not "remove"(x)

Yulyaswan:
@intoeetive should be ok now

intoeetive (approved):
👍

PR #3264, opened 2023-04-12
Fixed #2028 where Installer/Updater does not respect the theme

### LINKED ISSUE #2028: Installer/Updater does not respect the theme
Minor issue, but if my CP theme is set to Dark, then I go to upgrade EE, the updater is in light mode. It looks like the updater does look good in dark mode if I manually set data-theme="dark"

Screen Shot 2022-04-18 at 8 27 33 AM

Yulyaswan:
Fixed #2028 where Installer/Updater does not respect the theme
EECORE-2216

EE7 version of #3265

### DIFF
--- a/system/ee/ExpressionEngine/View/_templates/login.php
+++ b/system/ee/ExpressionEngine/View/_templates/login.php
@@ -10,6 +10,15 @@
 		<?=ee()->view->head_link('css/common.min.css'); ?>
 	</head>
 	<body data-ee-version="<?=APP_VER?>">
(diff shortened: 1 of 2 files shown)


intoeetive (commented):
@Yulyaswan can you think how we can avoid "blink" on login page? Because it's relatively small, it shows as white for part of second, then "blinks" into dark. This does not happen on "big" CP pages

intoeetive (approved):
excellent!

PR #3282, opened 2023-04-14
Improved styles for Pages add-on settings

Yulyaswan:
Improved styles for Pages add-on settings

EE7 version of #3283


intoeetive (changes requested):
@Yulyaswan while not used anywhere else currently, this fieldtype is shared, so I think the styles selector need to be more generic, not based on ID coming from field

intoeetive (changes requested):
@Yulyaswan does not seem to accomodate long channel names well: 
image

intoeetive (approved):
Looks much better now, thank you!

PR #3284, opened 2023-04-14
Improved #518 where 'Include in search' option from Relationships field is hidden.

### LINKED ISSUE #518: Relationship fields have 'Include in search' option that doesn't do anything

Screen Shot 2020-09-14 at 11 29 26 AM

The search option doesn't seem to do anything.  It either needs to be removed or the related data needs to pull up the entry when searched.

**How To Reproduce**

**Environment Details:**
 - Version: [e.g. 5.3.2]

Yulyaswan:
Improved #518 where 'Include in search' option from Relationships field is hidden. closes #518
EECORE-2218
EE7 version of #3285

### DIFF
--- a/themes/ee/asset/javascript/src/cp/conditional_logic.js
+++ b/themes/ee/asset/javascript/src/cp/conditional_logic.js
@@ -537,7 +537,6 @@ $(document).ready(function() {
 
         var textInputSelectors = 'input[type=hidden], textarea, button.toggle-btn';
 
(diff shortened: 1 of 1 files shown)


intoeetive (changes requested):
@Yulyaswan this works well, but I just realized that we need to implement same behavior when Relationship is used as column in Grid

(unless you say this needs to be addressed separately)

intoeetive (approved):

PR #3296, opened 2023-04-18
Resolved issue where Control Panel links could sometimes break after submitting idle login popup

### LINKED ISSUE #218: User gets logged out
Following upgrade to EE 5.2.3 I get logged out of the control panel after less than 30 minutes of activity - even while creating content.

**How To Reproduce**
1. Check Keep me logged in.
2. Log into Control Panel
3. Work on creating new entry
4. Less than 30 minutes of activity and the login screen appears

**Error Messages**
None

None

**Environment Details:**
 - Version: 5.2.3
 - PHP Version 5.6.40
 - MySQL Version 5.5.62-cll
 - OS: macOS
 - Web Server: Apache

In my config file I have the following settings:
...
$config['cookie_prefix'] = '';
$config['cookie_path'] = '';
$config['cookie_domain'] = '';
$config['user_session_type'] = 'c';
$config['admin_session_type'] = 's';
...

**Possible Solution**

**Additional context**

intoeetive:
will (hopefully) resolve #218

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -1211,7 +1211,11 @@ GRID_FALLBACK;
         //just to prevent any errors
         if (! defined('BASE')) {
             $s = (ee()->config->item('cp_session_type') != 'c') ? ee()->session->userdata('session_id') : 0;
-            define('BASE', EESELF . '?S=' . $s . '&amp;D=cp');
+            $base = EESELF . '?/cp';
+            if (!empty($s)) {
+                $base .= '&amp;S=' . $s;
+            }
+            define('BASE', $base); // cp url
         }
 
         $this->json = $this->_meta['json'];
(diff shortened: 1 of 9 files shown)


intoeetive:
this has been tested by @Yulyaswan and @robinsowell who confirmed in Slack it does not break anything

we can't be sure it's solving the issue as we don't have reliable way to reproduce it - but looks like it does

ignetic:
Just thought I'd chime in here where I've had this issue for a while now and created a fix which seems to hold. This is by changing global_start.js where it has:

```
	var replaceBase = function(i, value) {
		if (value) {
			return value.replace(oldBaseS[1], newBaseS[1]);
		}
	};
```

to

```
	var replaceBase = function(i, value) {
		if (value && oldBaseS[1]) {
			return value.replace(oldBaseS[1], newBaseS[1]);
		}
	};
```

Fix for replace which adds S=0 to beginning of string where needle may be empty e.g. S=0https://mywebsite.com/

intoeetive:
@ignetic thank you! Would you mind creating a pull request with this fix?

PR #3315, opened 2023-04-21
Resolved #1788 where HTML Buttons were loaded with Channel Form even when not in use

### LINKED ISSUE #1788: Curly brace-reliant HTML buttons breaking channel form jQuery assets
PHP 7.3
EE 6.2.4

I have a channel form that uses a couple RTE fields (CKeditor), so I'm including jQuery assets, (but not CSS). 

None of my other form fields are textarea fields, but I'm seeing a call for the HTML buttons we set up in the EE control panel. One of our standard buttons (to help clients remember how to make sure their email addresses get encrypted), is "email" and we set up something like this...
> Screen Shot 2022-02-04 at 2 11 31 PM
However, that type of HTML button will break the channel form's jQuery, as it gets read as an interrupting <script>. I finally tracked it down at the end of this line...
> Screen Shot 2022-02-04 at 2 13 32 PM
I'm not sure why the HTML buttons are getting called in the first place, but is there a workaround for this? It's not always possible for us to just have "Render URLs and Email addresses as links?" turned on for each channel, so I'd still like to offer my clients a way to encrypt their emails, or create "Contact Foo" links that are encrypted mailto's.

intoeetive:
Resolved #1788 where HTML Buttons were loaded with Channel Form even when not in use

@robinsowell - maybe you could test if this does not break anything

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -52,10 +52,13 @@ class Channel_form_lib
     public $show_fields;
     public $title_fields;
     public $valid_callbacks;
+    public $custom_field_names = [];
 
     public $lang;
     public $api_channel_fields;
     public $form_validation;
+    public $output_js = [];
+    public $extra_js = [];
 
     protected $_meta = array();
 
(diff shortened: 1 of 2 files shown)


robinsowell (approved):
So I was never able to actually cause a js error with this pre-fix, but I did test with both an rte and a textarea with buttons assigned.  Without the fix, even if textarea was not included in the form template, the button js was there.  Without it on the page, it wasn't.  Dates still worked fine either way.  live url title throws a js error either way, I believe completely unrelated to this.

Screenshot 2023-08-14 at 10 14 15 PM

Works fine from basic tests and I understand the logic.  But not tested brilliantly because I wasn't dead sure what clever things to try to break it, so I tried the basic non-clever things.

PR #3323, opened 2023-04-24
Resolved issue where image manipulation could malfunction if the resize protocol was not set

intoeetive:
Resolved issue where image manipulation could malfunction if the resize protocol was not set

fixed by setting fallback to `gd2` as default


bryannielsen:
@intoeetive would it be cleaner to just handle this in `EE_Image_lib::initialize()`?  I see we already have the default set there by `public $image_library = 'gd2';` I think we could just not allow an empty `image_library` value to be set.  What do you think?

intoeetive:
This comes from Portage work-in-progress but i'm not sure exactly why this had to be added...
Your suggestion is good, but i don't know if that would solve the issue, as i don't know what the issue was )
I think this PR will not make anything bad, at the minimum

PR #3330, opened 2023-04-24
Normalize category groups assigment to channels and upload directories

intoeetive:
Normalizing the database structure for cat_group => channel and cat_group => upload_directory relationships

The older cat_group property is still being set, so that legacy add-ons could use it, but all the records are being migdated to new tables and $channel->CategoryGroups should be used in code


intoeetive:
This will eventually become part of https://github.com/ExpressionEngine/ExpressionEngine/pull/3555 but keeping it here for now

PR #3355, opened 2023-05-02
Resolved #2806 where relationships field data was not saved if that was the only change for ChannelEntry model

### LINKED ISSUE #2806: Unable to save relationship data using Channel Model when that is the only update to the entry

If I try to add a relationship to an existing entry using the Channel Model, the data is not saved unless I update another field in the entry as well. 

`$entry = ee('Model')->get('ChannelEntry', $entry_id)->first();`
`// relationship field to update`
`$entry->field_id_XX = array('data' => [$entry_id_to_relate]);`
`$entry->save();`

The above code will not save the related entry. I tried all versions of the data array, ie
`array('data' => array($entry_id_to_relate))`
etc

I stumbled on the issue when I tried to update the title, just to see if the entry wasn't saving for some reason. When I added the title to the update before saving, the entry updated the title as well as the relationship field

**How To Reproduce**

1. Get an existing entry from a channel with a relationship field in it using the Channel Model
2. Attempt to _only_ update the relationship field
3. Save the entry
4. The relationship data shouldn't have saved, if this is reproducable
5. If you reproduced it, update the same entry's title along with the relationship data to see if they both save correctly

**Environment Details:**
 - Version: [7.1.5]
 - PHP Version [7.4.33]
 - MySQL Version [8.0.31-0ubuntu0.22.04.1]
 - OS: [Linux: Ubuntu 22.04.1 LTS]
 - Web Server: [Apache 2.4.52]

intoeetive:
Resolved #2806 where relationships field data was not saved if that was the only change for `ChannelEntry` model

UPD: the solution here was changed, see the last comment

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -114,7 +114,7 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
 
         if (isset($model) && is_object($model)) {
             $name = $this->field_name;
-            $model->$name = '';
+            $model->$name = null;
         }
 
         ee()->session->set_cache(__CLASS__, $cache_name, array(
--- a/system/ee/ExpressionEngine/Library/Data/Entity.php
+++ b/system/ee/ExpressionEngine/Library/Data/Entity.php
@@ -648,7 +648,9 @@ abstract class Entity extends MixableImpl implements Publisher
         if ($new_value !== $old_value) {
             if (! $this->hasBackup($name)) {
                 $this->setBackup($name, $old_value);
-            } elseif ($new_value === $this->getBackup($name)) {
+            } elseif ($new_value === $this->getBackup($name) && !is_null($new_value)) {
+                // if both values are NULL, we might be dealing with complex fields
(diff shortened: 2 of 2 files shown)


intoeetive:
Apparently my change is breaking `ExpressionEngine\Service\Model\OneToOneModelTest.php` PHPUnit test

I am not sure what that means and what the fix should be - @bryannielsen maybe you could have a look when you get time (I'd say this is low priority)

intoeetive:
@bryannielsen @matthewjohns0n can either of you look into this? I'd love to get this into 7.4 if it's not breaking anything

intoeetive:
The problem: empty string saved to field is not same as null, therefore 'backupIfChanged' is called. But it then determines that empty is equal to empty, so it marks the field as clean. If that's the only change, the saving gets skipped.

The solution is to keep relationship column value as null... I don't know though if that would have anhy consequences - like, are we somewhere checking that relationship field is empty string? I think that's really unlikely though

bryannielsen (approved):
This looks good for relationship fields, in the original issue @intoeetive you mentioned this would address grids too but I'm still seeing those not save if they're the only change on the model

matthewjohns0n (approved):

PR #3370, opened 2023-05-05
Create update:file-usage command

matthewjohns0n:


intoeetive:
@matthewjohns0n it's failing the tests, can you check?

intoeetive (changes requested):
I have tested this on default theme install and got few issues:
- the Update File Usage Utility in the CP does not work at all. It keeps sending AJAX requests, but nothing happens.
- the CLI seem to have worked (but I did not try it on really large website). However in the logs there is no record 'Started updating file usage' (unless that's intentional) and `update_file_usage_completed` shows literally as `update_file_usage_completed`, not as proper language string

matthewjohns0n:
@intoeetive Can you review this again? Those issues should be wrapped up now.

PR #3376, opened 2023-05-08
Added `primary_role_id` parameter for `exp:channel:entries` to work same as older `group_id`; #1353

### LINKED ISSUE #1353: Rename `group_id` parameter of Channel Entries
[`group_id`](https://docs.expressionengine.com/latest/channels/entries.html#group_id) is a too generic parameter name. Couldn't it be renamed to `member_role_id`?

The current version could be removed from docs until eecms version 7.

intoeetive:
Added `primary_role_id` parameter for `exp:channel:entries` to work same as older `group_id`; closes #1353

User Guide https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/650

EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/3377


bryannielsen:
Is this only filtering based on `members.role_id`?  Should we be using a more generic term like `member_role_id` and then filter based on the multiple roles a member could have in `members_roles.role_id`?

intoeetive:
@bryannielsen that would be a nice feature, however it's completely not what this PR is about. This PR is merely adding an alias to already existing parameter that checks ONLY primary role to have more clear meaning in EE6/7 terms

bryannielsen (approved):

PR #3386, opened 2023-05-12
Resolved #587 where addon description was truncated without ability to see entire description

Yulyaswan:
Improved #587 Where addon description truncated without ability to see entire description; resolved #587

EE7 version of #3387

PR #3387, opened 2023-05-12
Improved #587 Where addon description truncated without ability to see entire description

### LINKED ISSUE #587: Addon description truncated without ability to see entire description
Add-on excerpt text gets truncated with no option to view all text or click to addon info page.
Screen Shot 2020-10-22 at 1 45 56 PM

Yulyaswan:
Improved #587 Where addon description truncated without ability to see entire description; resolved #587 
EE6 version of #3386

### DIFF
--- a/cp-styles/app/styles/components/_add-on-card.scss
+++ b/cp-styles/app/styles/components/_add-on-card.scss
@@ -116,9 +116,9 @@
 
 .add-on-card__desc {
   position: relative;
-  white-space: nowrap;
-  overflow: hidden;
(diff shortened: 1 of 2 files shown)


matthewjohns0n:
My issue with this is that it allows add-on devs to put any length in the description, which has the potential to make the CP look pretty bad:

Screenshot 2023-06-22 at 3 37 00 PM

Maybe there's a reasonable amount of description that we can allow before cutting it off.

intoeetive:
moving into Draft until @TomJaeger tells us what to do :)

PR #3391, opened 2023-05-15
Resolved #532 where index template could be deleted but then immediately recreated

### LINKED ISSUE #532: Deleting a template group's index results in a confirmation but also fails
In EE 5.3.2 I tried to delete the index template of a template group. I got a message that it was successful, but it still existed.

**How To Reproduce**

1. Create template group.
2. Delete the index template.
3. You get a confirmation message, but will fail.

Screen Shot 2020-09-25 at 2 34 52 PM

I actually presume this is the intended behavior, in which case the message should be an error instead of a confirmation.

intoeetive:
Resolved #532 where index template could be deleted but then immediately recreated

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Design/AbstractDesign.php
+++ b/system/ee/ExpressionEngine/Controller/Design/AbstractDesign.php
@@ -513,6 +513,7 @@ abstract class AbstractDesign extends CP_Controller
             $column[] = array(
                 'name' => 'selection[]',
                 'value' => $template->template_id,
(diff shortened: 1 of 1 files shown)


bryannielsen:
I don't know if this is the right solution, removing the `index` template from being selected for bulk actions seems like it could cause issues down the road if/when `delete` is not the only bulk action

intoeetive:
@bryannielsen I'm sure we can find a lot of things that are not compatible with potential future updates :)
But I think we should aim to make the current UI work, and tackle the problems as they appear

bryannielsen (approved):

PR #3394, opened 2023-05-15
Resolved #3383 where loading File Grid fields with many rows was slow

Yulyaswan:
Improved #3383 where large file grids slow loading all images; Resolved #3383


robinsowell:
Nice!  I've got a user testing it out now.

intoeetive (approved):

PR #3395, opened 2023-05-15
Resolved #262 where it was not possible to change case of template group

### LINKED ISSUE #262: Minor glitch in Template Manager: trying to change case of template group name gives error
This is a minor glitch in the Template Manager. Attempting to change the name of a template group in such a way that the characters are the same but the case of one or more characters is different produces an error:

`The template group name you submitted is already taken`

**Environment Details:**
 - Version: [e.g. 5.2.2]

**Possible Solution**
Validation of new template group name should be case-sensitive, to identify that new template group name isn't the same as old. This should be clever enough to allow the change of a group name, without allowing the creation of a new group name which is the same as another except for case. A group name can clash with another, but not with itself!

### LINKED ISSUE #4424: Changing the case of a template group name, but not the name itself, is not allowed - even if changed name matches actual group folder name.
A previous developer created a site with many template groups using Capitalised names. This lead to a variety of problems, as EE seems to respect the case of a group when doing routing - so if the group is called `Cats` then `/cats` won't find it.

If you try to correct this capitalisation within the CP you get validation errors when you try to remove the capitalisation - on basis that the 'new' name is already in use - as presumably the validation used by EE is case insensitive and so is picking up on the group you are trying to rename... 

This means even if you manually rename the groups in a file manager the only way to get the EE template group titles to match is to manually edit the db ... which is less than ideal.

This issue is a resurfacing of an earlier one (#262) - that one got as far as having a solution proposed but it looks like the implementation of the solution did not happen due to uncertainty about consequential issues arising in other parts of EE.

**Environment Details:**
 - Version: 7.4.11
 - PHP Version 8.3

**Possible Solution**
Make the validation of group names case sensitive...

intoeetive:
Resolved #262 where it was not possible to change case of template group

Closes #4424

### DIFF
--- a/system/ee/ExpressionEngine/Model/Template/TemplateGroup.php
+++ b/system/ee/ExpressionEngine/Model/Template/TemplateGroup.php
@@ -12,8 +12,6 @@ namespace ExpressionEngine\Model\Template;
 
 use ExpressionEngine\Service\Model\Model;
 
-use ExpressionEngine\Library\Filesystem\Filesystem;
-
 /**
  * Template Group Model
  */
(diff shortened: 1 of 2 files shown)


TomJaeger:
I think this one needs further discussion as a whole as there are many things at play.

- Routing
- File system case sensitive or not
- Structure
- Pages
- Coilpack
- Future changes on the routeing front as a whole

PR #3449, opened 2023-05-24
Update cli prompt lang

TomJaeger:
Added prompts that were missing


matthewjohns0n (commented):
We dont reuse the lang entries anywhere else, so I ended up adding a `command_update_env_ask_config_value` lang entry. Approved 👍🏼

PR #3452, opened 2023-05-25
Resolved #3439 where using File field limited to upload directory on other site would result in JavaScript error

### LINKED ISSUE #3439: File field can break publish page in MSM
Testing 7.3 and I LOVE you can have file directories shared across sites.

BUT in testing, I was reminded how it's not a great experience if you assign a file field to a channel and the file field does not have access to a directory on that site.  It just breaks itself and all of the js:

Screenshot 2023-05-23 at 2 33 31 PM

This isn't really a new bug, but with the ability to use directories across msm sites, maybe either do not even show fields as in the field options if they aren't valid on the site or show a warning or something.  But it was a bad user experience and for a moment I thought there was some new bug introduced.

intoeetive:
Resolved #3439 where using File field limited to upload directory on other site would result in JavaScript error

image

@robinsowell please review. We need to test it in different combinations (sites, member role permissions, existing files etc.) to make sure it does not cause data loss


robinsowell:
I'm running into issues, particularly in a matrix field.

I download the branch, did default install.  Made a file field called 'blog_file' which is a file field limited to the 'blog' directory.  This is all on the default site.  Also note, I'm showing version 7.2.17.

I made a second msm site, 'dos'.  Made a channel in dos, assigned the 'blog' group and my new 'blog_file' field.

Go to add entry in my dos channel.  No js errors- which is good and an improvement.  My blog field shows:


Screenshot 2023-06-12 at 10 36 59 AM

That's correct.  My grid field could be a bit better on messaging, but I still don't trigger js errors nor can I do a bad upload:

Screenshot 2023-06-12 at 10 37 08 AM

HOWEVER- if I now go make a file upload directory in the dos site, now when I go to add an entry in dos channel, I still don't have js error but my file field looks like:

Screenshot 2023-06-12 at 10 35 25 AM. 

If I do drag/drop to it I get:

Screenshot 2023-06-12 at 10 43 04 AM

Weirdly, my matrix field with the file field actually gives me the option of the file field on the dos site:

Screenshot 2023-06-12 at 10 41 28 AM

Which it should not.  If I go to the field settings while on dos, it doesn't show any directory selected, which is what happens when the directory chosen isn't available on the site you are currently on:

Screenshot 2023-06-12 at 10 42 33 AM

I had similar issues if I just dropped the updated files into my 7.3 install to test.

robinsowell (dismissed):
See comments, it's doing odd things still.

robinsowell (approved):

robinsowell:
@intoeetive I can't get it to break again.  I like seeing the debug in the console:

Screenshot 2023-07-10 at 11 27 55 AM

That's useful.  I will grant, my head is spinning from all of the possible permutations testing this, but I'm calling it good.

I do still get a radio with nothing selecting in the actual field settings when an unavailable directory is chosen, but I'm going to put that in as a seperate (minor) bug- I think a flag or... something belongs there.  Possibly a validation error.

PR #3469, opened 2023-05-30
Added visual indication of future entry date and past expiration date in Entry Manager; #2912

intoeetive:
Added visual indication of future entry date and past expiration date in Entry Manager; closes #2912

Based on #1603 by @mithra62 

This PR adds appropriate hourglass icon with a tooltip:
- to entry date column ONLY if entry date is in future
- to expiration date column ONLY if expiration is in the past
image


intoeetive:
@TomJaeger because this is UI change, we need your approval

matthewjohns0n (dismissed):

PR #3470, opened 2023-05-30
Improved getting real IP address of visitor when using load balancers

intoeetive:
based on #446 by @litzinger 

I am not sure this is best possible order of headers, but I also don't know if that matters a lot

User guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/689

Because this is adding new config file, setting milestone to 7.4


bryannielsen (dismissed):

intoeetive:
@matthewjohns0n this is adding new composer dependency for working with IP address ranges. Those files are pure math, but please review from security standpoint

bryannielsen (approved):

PR #3476, opened 2023-05-31
Improved Control Panel accessibility for using screen readers

Yulyaswan:
common changes for the whole site. Does not include color contrast enhancement

EE7 version of #3511


intoeetive (changes requested):
@Yulyaswan we need to make all labels use lang() function and be placed in lang files, so that those would be translatable

system/ee/ExpressionEngine/Service/Thumbnail/Thumbnail.php
```diff
@@ -68,34 +68,34 @@ public function __construct(File\FileSystemEntity $file = null)
             } elseif ($file->isEditableImage() || $file->isSVG()) {
                 $this->url = $file->getAbsoluteThumbnailURL();
                 $this->path = $file->getAbsoluteThumbnailPath();
-                $this->tag = '<img src="' . $this->url . '" alt="' . $file->title . '" title="' . $file->title .'" class="thumbnail_img" />';
+                $this->tag = '<img src="' . $this->url . '" alt="' . $file->description . '" title="' . $file->title .'" class="thumbnail_img" />';
```
intoeetive: I think many people would use file description for longer description, so we need to run it through `ee()->functions->word_limiter` and make sure HTML is stripped off, if we want to use it

system/ee/ExpressionEngine/Service/Thumbnail/Thumbnail.php
```diff
@@ -68,34 +68,34 @@ public function __construct(File\FileSystemEntity $file = null)
...
                 switch ($file->file_type) {
                     case 'doc':
-                        $this->tag = '<i class="fal fa-file-alt fa-3x"></i>';
+                        $this->tag = '<span class="sr-only">' . lang('file_icon') . '</span><i class="fal fa-file-alt fa-3x"></i>';
                         if (strpos($file->mime_type, 'pdf') !== false) {
-                            $this->tag = '<i class="fal fa-file-pdf fa-3x"></i>';
+                            $this->tag = '<span class="sr-only">' . lang('pdf_file_icon') . '</span><i class="fal fa-file-pdf fa-3x"></i>';
                         } elseif (strpos($file->mime_type, 'html') !== false || strpos($file->mime_type, 'css') !== false || strpos($file->mime_type, 'xml') !== false) {
-                            $this->tag = '<i class="fal fa-file-code fa-3x"></i>';
+                            $this->tag = '<span class="sr-only">file icon</span><i class="fal fa-file-code fa-3x"></i>';
```
intoeetive: not translated here

system/ee/ExpressionEngine/Library/CP/FileManager/Columns/Thumbnail.php
```diff
@@ -41,7 +41,7 @@ public function renderTableCell($data, $field_id, $file, $viewtype = 'list')
                 $url = ee('CP/URL')->make('files/directory/' . $file->upload_location_id, ['directory_id' => $file->file_id]);
                 $file_thumbnail = '<a href="' . $url . '">' . $thumb->tag . '</a>';
             } elseif (ee('Permission')->can('edit_files')) {
-                $file_thumbnail = '<a href="' . ee('CP/URL')->make('files/file/view/' . $file->file_id) . ($file->isImage() ? '" class="imgpreview" data-url="' . $thumb->url : '') . '">' . $thumb->tag . '</a>'; 
+                $file_thumbnail = '<a href="' . ee('CP/URL')->make('files/file/view/' . $file->file_id) . ($file->isImage() ? '" class="imgpreview" data-url="' . $thumb->url : '') . '" alt="' . $file->description . '">' . $thumb->tag . '</a>';
```
intoeetive: see the comment below

system/ee/ExpressionEngine/Addons/pro/Service/FrontEdit/FrontEdit.php
```diff
@@ -295,7 +295,7 @@ public function loadFrontEditAssets($output)
                     . '" data-editableurl="'
                     . $fieldEditUrl
                     . '" data-entry_id="ENTRY_ID" data-site_id="SITE_ID" data-size="WINDOW_SIZE" title="FIELD_NAME">'
-                    . '<img src="' . $pencilUrl . '" width="24px" height="24px" style="cursor:pointer !important; filter: drop-shadow(0 1px 3px rgba(0,0,0,.20)) !important; vertical-align: bottom !important; border-radius: 0 !important; width: unset !important;" />'
+                    . '<img src="' . $pencilUrl . '" width="24px" height="24px" alt="pencil icon" style="cursor:pointer !important; filter: drop-shadow(0 1px 3px rgba(0,0,0,.20)) !important; vertical-align: bottom !important; border-radius: 0 !important; width: unset !important;" />'
```
intoeetive: not translated

frontedit/src/js/components/DockDragHandle.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <div class="cursor-move-ee-44E4F0E59DFA295EB450397CA40D1169" v-on:dblclick="toggleCollapsed">
-        <img v-if="storeHost" v-bind:src="storeHost +'img/drag-handle.svg'" class="svg-ee-44E4F0E59DFA295EB450397CA40D1169 ee-pro-drag-handle-ee-44E4F0E59DFA295EB450397CA40D1169" />
+        <img v-if="storeHost" v-bind:src="storeHost +'img/drag-handle.svg'" class="svg-ee-44E4F0E59DFA295EB450397CA40D1169 ee-pro-drag-handle-ee-44E4F0E59DFA295EB450397CA40D1169" alt="drag handle icon"/>
```
intoeetive: not translated

system/ee/ExpressionEngine/View/_templates/default-nav.php
```diff
@@ -13,7 +13,7 @@
     <div class="main-nav__toolbar">
         <?php if (isset($header['toolbar_items']) && $header['toolbar_items']): ?>
             <?php foreach ($header['toolbar_items'] as $name => $item): ?>
-                <a class="button button--secondary icon--<?=$name?>" href="<?=$item['href']?>" title="<?=$item['title']?>"><span class="hidden"><?=$item['title']?></span></a>
+                <a class="button button--secondary icon--<?=$name?>" href="<?=$item['href']?>" title="<?=$item['title']?> button"><span class="hidden"><?=$item['title']?></span></a>
```
intoeetive: "button" part of title is not translated

intoeetive (changes requested):
@Yulyaswan overall good, I found few small issues and we also need to have merge conflicts resolved

intoeetive (dismissed):

intoeetive:
@Yulyaswan the code change are good (I did not actually test though, just reviewed the code), but I think the changes in language need better English wording for both keys and values.

@matthewjohns0n can you have a look at changes (mainly lang file) and provide updates as needed?

system/ee/ExpressionEngine/View/_shared/footer.php
```diff
@@ -110,6 +110,10 @@
 } ?>
 <?php echo implode('', ee('CP/Modal')->getAllModals()); ?>
 
+        <?=ee()->javascript->set_global([
+            'lang.password_icon' => lang('password_icon')
+        ])?>
+
```
intoeetive: I don't think we are supposed to set JS globals in view file; this is where we render those already, not set.
If this is something that we need on ALL pages, we could do that in legacy/libraries/Core in run_cp function. But perhaps we could just do this in a non-javascript way?

system/ee/ExpressionEngine/View/_shared/form/buttons.php
```diff
@@ -79,6 +79,7 @@
 		<button class="<?=$class?>" <?=$button['attrs']?> <?=$disabled?> name="<?=$button['name']?>" type="<?=$button['type']?>" value="<?=$button['value']?>" data-submit-text="<?=lang($button['text'])?>" data-work-text="<?=lang($button['working'])?>"><?=$button_html?><?=$button_text?></button>
 		<?php if ($i == 0 && count($submits) > 1) : ?>
 		<button type="button" class="<?=$class?> dropdown-toggle js-dropdown-toggle saving-options" data-dropdown-pos="bottom-end">
+            <span class="sr-only"><?=lang('save_btn')?></span>
```
intoeetive, line 82: I don't think this label is correct here, as it's for the dropdown arrow, not for save button

system/ee/ExpressionEngine/View/_shared/header.php
```diff
@@ -3,7 +3,7 @@
     <head>
         <?=ee()->view->head_title($cp_page_title)?>
         <meta http-equiv="content-type" content="text/html; charset=utf-8">
-        <meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"  name="viewport">
```
intoeetive: Just making sure there will be no undesired consequences on different screen resolutions if we remove this (e.g. in modals, slide-outs, jump menu behaviour etc)

intoeetive (commented):
@Yulyaswan please see few extra comments I've spotted. I also wonder if we need to merge latest 7.dev now because of already made changes in JS?

system/ee/ExpressionEngine/View/_shared/form/buttons.php
Yulyaswan: it's just label that says: 
`'save_btn' => 'Button that opens options for saving the page',`

system/ee/ExpressionEngine/View/_shared/footer.php
Yulyaswan: moved to `system/ee/legacy/libraries/Cp.php`
This part is used in Member, Role pages, login page and login modal window

system/ee/ExpressionEngine/View/_shared/header.php
Yulyaswan: `Maximum-scale is an optional attribute for the viewport, and it defines the maximum zoom that website users are able to do. Setting a value for the 'maximum-scale' attribute could prevent the user from scaling properly, meaning that users may not be able to zoom in/out on mobile devices, and may cause accessibility issues.`
`The user-scalable attribute defines whether users can scale (zoom) the page content on mobile. By default, scaling is on, but if user-scalable is set to '0' or 'no' then this will prevent users from zooming, which negatively affects user experience (in most cases).`
Screenshot 2023-08-07 at 13 39 44
The recommended, 'safe' viewport setting is which we have on site:
`<meta name="viewport" content="width=device-width, initial-scale=1">`

intoeetive: @Yulyaswan this does break things :(
On File Manager page on iPhone 12 (Chrome simulator) in horizontal mode the main sidebar is not being automatically collapsed, making it nearly impossible to use.

intoeetive (changes requested):
@Yulyaswan - see the comment.

All other changes seem to be fine, so if we can skip that change for now, I can approve and get this merged.

intoeetive (approved):

PR #3482, opened 2023-06-01
Add field group shortname to update routine

### LINKED ISSUE #3480: 7.3 field group names required for fieldlet tags, but have spaces
We don't have a short name for field groups, and we do allow spaces, so it's not weird to have field group names like 'Something with a space'.  

If you use this in a fluid field as a group, your template looks like

```
{my_fluid}
       {my_fluid:Something with a space}
        {fields}
```
Aside from being weird, the closing ```{my_fluid:Something with a space}``` is unparsed, though everything else works fine.

Probably best to make a short name and require that.

matthewjohns0n:
Adds short name to field group. 

WIP.

Fixes #3480.


intoeetive:
Made further changes:
- renamed property to short_name to be consistent with the similar name in other places
- use built-in alphaDash validation
- added beforeValidate setter to it wouldn't break existing code
- fixed updater to use direct DB queries, so it would be break-proof if the model changes in future
- added column in installer schema
- updated the actual fluid parser to use short_name
- added new variables
- last but not least, updated the docs

PR #3487, opened 2023-06-02
Fixes many typos in the core

matthewjohns0n:


system/ee/ExpressionEngine/Addons/spam/upd.spam.php
```diff
@@ -443,7 +443,7 @@ private function updateChannelSpam_2_00_00()
     private function updateForumSpam_2_00_00()
     {
         // Legacy forum stores raw SQL. Get rid of all UPDATE queries, since we have no
-        // way to determine what order they should run in, and UPDATEs (edits) to topics or posts
+        // way to determine what order they should run in, and UPDATEEs (edits) to topics or posts
```
TomJaeger: two E's here

system/ee/ExpressionEngine/Controller/Utilities/SyncConditionalFields.php
```diff
@@ -237,7 +237,7 @@ public function evaluate()
             'message_type' => 'success',
             'channel_id' => $channel_id,
             'entries' => $entries->pluck('entry_id'),
-            'entries_proccessed' => $entries->count()
+            'entries_processed' => $entries->count()
```
TomJaeger, line 240: We need to confirm through the core that the updated array key is used. @matthewjohns0n

system/ee/ExpressionEngine/View/_shared/sidebar/navigation/list_item.php
```diff
@@ -9,7 +9,7 @@
         </span>
     <?php endif; ?>
     <?php if ($addlink) : ?>
-    <a href="<?=$addlink?>" class="dropdown__item-button button button--secondary button--xsmall"<?=$addlinkAttibutes?>><i class="fal fa-plus"></i><span class="hidden"><?=lang('btn_create_new') . ' ' . $text?></span></a>
+    <a href="<?=$addlink?>" class="dropdown__item-button button button--secondary button--xsmall"<?=$addlinkAttributes?>><i class="fal fa-plus"></i><span class="hidden"><?=lang('btn_create_new') . ' ' . $text?></span></a>
```
TomJaeger, line 12: Not sure we can change this one? @matthewjohns0n

system/ee/ExpressionEngine/Addons/spam/upd.spam.php
matthewjohns0n: Fixed

system/ee/ExpressionEngine/Controller/Utilities/SyncConditionalFields.php
matthewjohns0n: I did verify this one and made sure it was updated everywhere.

system/ee/ExpressionEngine/View/_shared/sidebar/navigation/list_item.php
matthewjohns0n: I did update this everywhere, but since it is a shared form view, we probably cant make this change.

matthewjohns0n:
After discussing this with @TomJaeger, we're taking this out of the 7.3.0 release since it requires a bit more manual checking of the changes.

intoeetive (commented):
I think we should NOT change variable / function names and CSS class names, even if they are misspelled. At least not in huge PR like this

If this would be update in comments only, I'd be happy to approve

PR #3505, opened 2023-06-12
Enabled showing field short names on Entry Publish page; #3340

intoeetive:
Closes https://github.com/ExpressionEngine/ExpressionEngine/discussions/3340

It has been requested few times to be able to view field short names directly on entry edit page, with the ability to copy them for usage in templates

This PR is adding that feature, see video for details https://www.dropbox.com/s/045xp7p2pfk12f7/Rec_12.31.40%2020230615.mp4?dl=0

The setting is per-role, it is ON by default for superadmins and existing editors on migrated sites; it's off by default for the new roles


intoeetive:
@Yulyaswan we need to have some CSS / HTML / JS here. Look for calls of `getNameBadge` function and `publish/partials/field_name_badge` view file. I think it could go on same line with Label, but in small font, grey and italic - but up to you if you have different ideas. Also maybe we could show some icon on hover and copy the content on click (with some visual indication - e.g. the icon could change)

Please have in mind this is also used in Fluid, Grid, and Grid inside Fluid

bryannielsen (approved):

matthewjohns0n:
I really like this feature! The only thing I don't like is that when you click to copy it, the text turns green, and then stays green. Then when you click to copy another it also turns green and stays green. I would like to see it fade back to the original color after a few seconds. 

EDIT: 
I just watched the video and saw that it does fade. This also happens locally, I just didn't wait long enough. So nevermind on that one!

TomJaeger:
I just tested this as well.  Its currently about a 12 count (my counting in my head) -- Auto save kicks in before the green fades.  Can we change this to about 3 seconds.

intoeetive:
@TomJaeger changed to 2 seconds (was 10)

TomJaeger:
This is much better.. Instead of the green, can we go with the same color the active tab on the publish screen gets? (Looks like a purple similar to the save button color in the snow theme)

intoeetive:
color changed

PR #3515, opened 2023-06-13
Added support for `STRAIGHT_JOIN` in ActiveRecord; #3073

robinsowell:
Added straight as a join type

Not adding it as a flag/keyword, though first commit has both

NOTE - removed the flag from the final commit per discussion

per user request, adding a straight join flag and join type.

used as join('channel_data', "channel_data.entry_id = channel_titles.entry_id", "straight");<br>
SELECT `exp_channel_titles`.`entry_id` FROM (`exp_channel_titles`) STRAIGHT_JOIN `exp_channel_data` ON `exp_channel_data`.`entry_id` 

<br><br>
used as ->straight()<br>
SELECT STRAIGHT_JOIN `exp_channel_titles`.`entry_id` FROM (`exp_channel_titles`

docs pr: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/678

### DIFF
--- a/system/ee/legacy/database/DB_active_rec.php
+++ b/system/ee/legacy/database/DB_active_rec.php
@@ -32,7 +32,6 @@ class CI_DB_active_record extends CI_DB_driver
     public $ar_wherein = array();
     public $ar_aliased_tables = array();
     public $ar_store_array = array();
-    public $ar_straight = false;
 
     protected $ar_group_count = 0;
     protected $ar_empty_group = true;
@@ -232,22 +231,6 @@ class CI_DB_active_record extends CI_DB_driver
         return $this;
     }
 	
-    /**
-     * STRAIGHT
-     *
-     * Sets a flag which tells the query string compiler to add STRAIGHT_JOIN
-     *
-     * @access	public
-     * @param	bool $val Set to FALSE to remove STRAIGHT from a query
-     * @return	CI_DB_active_record The active record object
-     */
-    public function straight($val = true)
-    {
-        $this->ar_straight = (bool) $val;
-
-        return $this;
-    }
-
     /**
      * From
      *
@@ -1752,9 +1735,6 @@ class CI_DB_active_record extends CI_DB_driver
             $sql = $select_override;
         } else {
             $sql = (! $this->ar_distinct) ? 'SELECT ' : 'SELECT DISTINCT ';
-			if ($this->ar_straight) {
-				$sql .= 'STRAIGHT_JOIN ';
-			}
 
             if (count($this->ar_select) == 0) {
                 $sql .= '*';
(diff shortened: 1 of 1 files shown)


bryannielsen:
@robinsowell is this syntax for the second usage correct?

```
SELECT STRAIGHT_JOIN exp_channel_titles.entry_id FROM (exp_channel_titles
```

I don't have much experience with Straight Joins but in the little research I did it seems like it should be used in a similar way as other Joins.  I can't find any references where it is part of the SELECT clause, maybe you can send me some links on that?

robinsowell:
@bryannielsen Does this help? https://blog.sqlauthority.com/2014/03/12/mysql-how-to-do-straight-join-in-mysql/

I was puzzled by the two different approaches at first, but poked around and added it for someone who had a really big speed improvement using it in one particular circumstance- thing it was actually in a solspace add-on.  But we needed it in native AR to apply it.

It did/does work for the end user, so even though obscure, I added it.

bryannielsen:
@robinsowell was it just the `SELECT STRAIGHT_JOIN` that helped the customer?  I just haven't seen that used much and I think we would rather encourage specifying the join type on the `join()` instead.  I'm leaning towards removing the `->straight(bool)` for selects and just keeping it as a new join type option unless there's a benefit to having the other.  If we do want to keep it I think I would remove the boolean argument so you're either forcing a query to use straight joins or not, or is there a use case for where you would use `->straight(false)` to turn off the straight join behavior on a query?

intoeetive:
Robin to say final word, but I think straight_join is a modifier for select, not actually join type. And it needs to be treated sane way as 'distinct'

bryannielsen:
@intoeetive I don't think that's actually the intended usage though.  The docs only mention using it as a join type - https://dev.mysql.com/doc/refman/8.0/en/join.html but I think it's possible to use it through a select, just like you can technically join tables without a join statement.  I just think if there's no benefit to this approach we shouldn't bother supporting it as it just causes more confusion

robinsowell:
In truth, this was a pure FR from a user- I think they wanted to use it on a Solspace Calendar query that was being super slow for them (had to be something to do with the way mysql was doing the join), and having the flag was a billion times easier to implement.  I'll see if I can find the support discussion.

robinsowell:
Agreed- I think adding the straight_join as a join type makes sense.  I'll drop the flag as it looks like it's non-recommended.

TomJaeger:
My gut is that we fall in line with the existing join_type param on the method. and add it there.

https://docs.expressionengine.com/latest/development/legacy/database/active-record.html#jointable-cond-type--

robinsowell:
@bryannielsen When you get a chance, can you take a look at this one again? I pulled out the flag/keyword, so it's just adding a straight join type.  Tested doing 

```
ee()->db->select('channel_name');
ee()->db->from('channels');
ee()->db->join('comments', 'comments.channel_id = channels.channel_id', STRAIGHT);
ee()->db->join('members', 'comments.author_id = members.member_id', STRAIGHT);
$query = ee()->db->get();
```

PR #3520, opened 2023-06-14
Resolved #3463 where "choose wisely" was a poor choice

### LINKED ISSUE #3463: "Choose wisely" for select dropdowns lacks professionalism
EE v 7.2.17

Wondering if there's a technical reason the select dropdown fields default to "Choose wisely" instead of something like "Please select" ? 

I'm sure there's a way for us to change this in the bowels of EE software, but that's not very practical for all the installs we have.

We train a lot of clients who are nervous about the web publishing process. Seeing "Choose wisely", even for something as simple as choosing a button color, can create some hesitation for these folks. In general, the choice of words also doesn't come across as very professional, in my opinion.

bryannielsen:
We are changing the blank select state from "Choose wisely" to "Please select" by introducing a new `please_select` language variable.  We are leaving in place the old `choose_wisely` language variable in case it is being used by others but this can be removed in the next major release.

### DIFF
--- a/cp-styles/pages/entry.php
+++ b/cp-styles/pages/entry.php
@@ -722,7 +722,7 @@ include(dirname(__FILE__) . '/_wrapper-head.php'); ?>
 					<div class="select">
 						<div class="select__button">
 							<label class="select__button-label">
-								<!-- Choose Wisely -->
+								<!-- Please select -->
 								<div class="status-tag st-open">Open</div>
 								<input type="hidden" name="">
 							</label>
@@ -748,7 +748,7 @@ include(dirname(__FILE__) . '/_wrapper-head.php'); ?>
 					<div class="select">
 						<div class="select__button">
 							<label class="select__button-label">
-								Choose Wisely
+								Please select
 								<input type="hidden" name="">
 							</label>
 						</div>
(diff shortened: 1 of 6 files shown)


intoeetive:
I have heard a lot of times from agencies that with even a little change in the UI (like the labels names differently) is requiring them to instruct the editors again.

So I don't think we should not just change the label for everyone, especially in patch release. I think we need to make it into system configuration variable (with "Choose wisely" being set for existing installs and "Please select" for new installs), and for the select fields maybe even make configurable in each field settings

bryannielsen:
I think that's a good policy in general but I do feel like this is really poorly worded and this change can only make it clearer for editors.  I can remember my first encounter with this "Choose wisely" label caused me to second guess whether it was safe to interact with this input.

intoeetive:
I think we need @TomJaeger decide on this one

intoeetive:
@TomJaeger we need your opinion here. I think we need to:
- make this text configurable as override in config file
- keep using 'Choose Wisely' for existing installs so that user would not be confused with change
- use 'Please select' for new installs

TomJaeger (dismissed):

PR #3523, opened 2023-06-14
Resolved #3522 where grid fields in a pro variable would not parse when set to early parsing

### LINKED ISSUE #3522: 7.3 issue with Pro Variables grid field not parsing
This worked in 7.2, it breaks in 7.3.  It only happens if the pro variable grid field has early parsing toggled ‘on’.  If I turn it off, it works. 

My template looks like:

```
{exp:pro_variables:parse var="lv_gridtest"}
  the file: {lv_gridtest:g_file}<br/>
  the text: {lv_gridtest:g_text}<br/>
{/exp:pro_variables:parse}
```
This outputs:
Screenshot 2023-06-14 at 11 35 43 AM

My grid field had 2 rows:

Screenshot 2023-06-14 at 11 36 51 AM

matthewjohns0n:
In EE 7.3+, the template parser sees `{lv_gridtest:g_text}` in the template, and also sees that `lv_gridtest` is a global variable.

It thinks that `g_text` is a modifier to `lv_gridtest`, and processes it as such. Since `lv_gridtest` is a global variable at this point, but it is a grid, the value is set to `''`.

The template parser processes the tag here and ends up stripping the tag out before it ever makes it to pro vars.

Resolved #3522 where grid fields in a pro variable would not parse when set to early parsing

### DIFF
--- a/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
+++ b/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
@@ -418,6 +418,19 @@ class LegacyParser
                     $content = $value;
                     $var_props = $this->parseVariableProperties($modified_var, $prefix);
 
+                    // Is pro variables installed?
+                    if (ee('Addon')->get('pro_variables')->isInstalled()) {
+                        // Get early parsed pro variables
+                        $early = ee()->pro_variables_variable_model->get_early();
+                        $early = pro_flatten_results($early, 'variable_data', 'variable_name');
+
+                        // If this is an early parsed pro variable, we skip it for now
+                        // and let it get parsed later
+                        if (array_key_exists($var_props['field_name'], $early)) {
+                            continue;
+                        }
+                    }
+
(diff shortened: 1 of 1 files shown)


matthewjohns0n:
@intoeetive It would be great if you could review this. You have the most knowledge of the modifier parsing, so I was hoping you'd notice if something was off.

This currently works, and even lets you use modifiers like this:

```
{exp:pro_variables:parse var="lv_gridtest"}
  the file: {lv_gridtest:g_file}<br/>
  the text: {lv_gridtest:g_text:rot13:encrypt}<br/>
{/exp:pro_variables:parse}
```

intoeetive:
@matthewjohns0n I don't quite like it.

It creates additional overhead by running extra query on each parser run and also feels hacky - if we ever bundle in early parsing in the core, it will stop working.

I am proposing another solution in https://github.com/ExpressionEngine/ExpressionEngine/pull/3525, please review it

(also closing this PR to avoid confusion)

PR #3525, opened 2023-06-15
Resolved #3522 where early parsed Grids in Pro Variables would not render properly

### LINKED ISSUE #3522: 7.3 issue with Pro Variables grid field not parsing

intoeetive:
Resolved #3522 where early parsed Grids in Pro Variables would not render properly

This change reverts to the behavior that existed before multiple modifiers feature was added - if the modifier does not exist (is actually a column name, or something completely fake) , the modified variable is not parsed. With the change that we now only check first modifier

### DIFF
--- a/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
+++ b/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
@@ -420,20 +420,29 @@ class LegacyParser
 
                     // in order to support multiple modifiers, we'll do this in a loop
                     if (isset($var_props['all_modifiers']) && !empty($var_props['all_modifiers'])) {
+                        $modifiersCounter = 0;
                         foreach ($var_props['all_modifiers'] as $modifier => $params) {
                             // is the modifier valid?
                             $method = 'replace_' . $modifier;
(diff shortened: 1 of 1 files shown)


matthewjohns0n (approved):
@intoeetive I just wanted to think this through - is there ever a case that this will skip a modifier that it shouldn't skip? (are we introducing a bug)?

I can't think of a case that would break here but just wanted to try to get ahead of something before it crops up if possible. 

I agree that adding queries is not ideal for the parser, so this looks like a better solution. I tested this with my example PV grid setup, as well as a few other PV variables using modifiers and it works for me so I'm approving.

intoeetive:
I don't think any such case would be possible. The only case where this will not be working is when you have something like {grid_field:some_modifier:column_name:another_modifier} but it's also not valid syntax in EE

PR #3526, opened 2023-06-15
Color Contrast Theme

Yulyaswan:
As discussed before, I created this PR (Draft) to improve color contrast. I used the `light snow theme` as an scss file to change colors. This work is based on Wave app and suggestion from it. I mainly used HEX colors (not variables), because, in my opinion, we don't need to do this kind of work for different themes, like dark or pink.

If @TomJaeger will approve these changes, then it will make sense to create a new theme, something like `"contrast-color theme"` using `cp-styles/app/styles/themes/_theme-light-snow.scss` as an example for colors.

IMPORTANT: Don't merge this PR anywhere. After merging this PR will break our Snow theme, and we don't want this.  This is just an example.

github issue 1590


Yulyaswan:
The theme is ready for review

robinsowell:
Wow- this is really cool!  I just patched it into a 7.5.8.

Couple of comments- frontend edit windows lose some of the high contract on the labels:
Screenshot 2025-03-19 at 5 04 12 PM

Left is cp, right is front end edit.  I suspect that should stick?

Also- and I think this is because we're riding on the snow theme- when you select it in the jump menu everything works, but it doesn't show as selected in the dropdown options:

Screenshot 2025-03-19 at 5 12 39 PM

I'm on high contrast, but it's not reflected in the dropdown.  Like I say, I suspect this isn't a 'real' bug, but mentioning it.

TomJaeger (approved):

PR #3547, opened 2023-06-19
Resolved issue where Structure would show deprecation notice when using PHP 8.2

robinsowell:
str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated


intoeetive (dismissed):

matthewjohns0n (dismissed):

intoeetive:
@robinsowell what has been the reason to move `reqiure_once` into `__construct`? We keep those out in all other add-ons

intoeetive (approved):

robinsowell:
@intoeetive Thanks you!

PR #3553, opened 2023-06-21
Added CLI command to run Content Reindex; #3428

intoeetive:
Added CLI command to run Content Reindex; #3428

also renames "search reindex" to "content reindex" as that better explains what it does

docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/683


bryannielsen:
This is great!  I'm wondering if we should change the command to `content:reindex` instead of `sync:reindex` so it's a little more descriptive?  Otherwise it sounds like we're syncing a reindex which doesn't make much sense.  If we were to make that change I think also changing the command class to be `ContentReindexCommand` instead of `ReindexCommand` would be good.  Being a little more specific might prevent a collision down the road if we have a broader `reindex` command that does more than just content.

intoeetive:
@bryannielsen please have a look at existing commands:
image
The concept is that the first part is the TYPE of operation performed (e.g. backup, update, - or sync) and the second part is the actual object we operate with. There is an exception here though, cache: clear is opposite.
But the general idea is that the commands are grouped. Content reindexing fall more in line with `sync` command, although I agree it does sound too good. If we want to be more descriptive, we could do `sync:content-reindex` or `sync:content`. `reindex:content` is probably what makes most sense, but `reindex` for content is realy similar to sync does for files. But then again, none of there is fully self-explainatory. Also, very few Linux command names make sense, you could never get idea of what `grep` does by looking at its name )
So options are:
- sync:reindex
- reindex:content
- content:reindex
- sync:content
- sync:reindex-content
- sync:content-reindex
- sync:content-index
- sync:searchable-content

CreativePoweredGeek:
Any idea when we might see this?   Looks like you have made progress.

CreativePoweredGeek:
> Any idea when we might see this? Looks like you have made progress.

??  trying to plan my timeline...  this is (at the moment) the last peice to my puzzle.

TomJaeger:
We're working to nail down our timeline for the 7.4 release currently.  We can get you a developer preview sooner then later though if that would be beneficial?

CreativePoweredGeek:
That would be great.  Thanks.

matthewjohns0n (approved):

PR #3555, opened 2023-06-22
Allow making category group required for channel; Allow restricting category selection to single category per group

intoeetive:
This PR allows marking category groups assigned by channel as:
- required (min 1 category from group needs to be selected for entry)
- allow / disallow multiple selection (default allow multiple selection and uses checkboxes, if disallowed is using radio buttons instead)
image

There are cases when "allow multiple" is unchecked, but we still need to display checkboxes; In such case we'll display warning:
a) when there are already several categories from this group assigned
image
b) when parent category needs to be selected automatically and child categories exist (because the setting can be changes and categories can be edited independently, we only can do this check on entry page)
image

It is also refactoring the categoryGroup<->Channel assigment by moving that to pivot table

Docs: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/722

Closes https://github.com/ExpressionEngine/ExpressionEngine/discussions/2865 discussion


intoeetive:
@Yulyaswan I've added hidden inputs and the code to display the toggles properly on initial load depending on state, but I really don't know what to do next. I think the `key` for input needs to be set to something and then maybe extra code written to get the hidden fields populated as you move the toggles?

matthewjohns0n (approved):

PR #3467, opened 2023-05-29
Resolved #3456 where manipulated images could have wrong rotation

intoeetive:
Resolved #3456 where manipulated images could have wrong rotation

PR #3569, opened 2023-06-27
Resolved #3456 where image manipulation could have wrong orientation

### LINKED ISSUE #3456: Uploaded images showing incorrect rotation
7.x issue

Basically, iphones use exif data to properly oriented images.  People upload an image that looks correctly oriented and it ends up sideways.

I know this has come up before and at one point 'fixed' but the fix no longer works and it's just not a good user experience, especially for non-technical users.  I believe any resizing/alteration loses exif and the image is just wrong.  

Here's the prior fix:
https://github.com/ExpressionEngine/ExpressionEngine/pull/277

Here's prior discussion:
https://github.com/ExpressionEngine/ExpressionEngine/issues/105

Here's 
an example- I'm also including a zip of the two files.  I was testing 7.3 and used the two test files in my tests- you can see the sideways orientation ends up sideways:

Screenshot 2023-05-25 at 11 39 45 AM

MAYBE it works with the original image (if browser reading exif) but the thumbs and file manips don't read it and strip it and you end up sideways and people end up confused.

[dylan-Images.zip](https://github.com/ExpressionEngine/ExpressionEngine/files/11566800/dylan-Images.zip)

bryannielsen:
This fix should address the issue where an orientation was detected in exif data and corrected for but the fixed image was not being used in subsequent manipulations and thumbnails.

If approved this PR closes #3467 and resolves #3456


intoeetive:
This seems to work for me - althought the other PR also worked - (which I'll close to avoid confusion), but I'd like extra confirmation from @robinsowell

robinsowell (approved):
Testing this killed me.

PR #3576, opened 2023-06-29
Resolved #3568 where deselecting all template access for Role in one MSM site would also deselect it on other sites

### LINKED ISSUE #3568: 7.3.5 MSM Template access permissions can get wiped editing Roles
This is hard to reproduce- I'm going to send over a database that you can use.  I could NOT always reproduce it and I dug a fair bit.  I give up.  

But- with the database (was originally a 7ish v, not ancient), go to Default site.  Go to Roles, select 'Editor' to edit.  Go to the 'Template Access' tab.  Click 'Check all' to select all of the groups/templates.  You might have to click it once to select, once to unselect, once to select again.  

Then switch sites to Dos, go to Roles, edit 'Editor' go to 'Template Access' tab, make sure they're all selected, save.  Go back to Default and check the Editor Template Access settings- they were unset.

Note- this came in via slack, I could not immediately replicate.  I added to templates to the Dos site, so that I had some duplicate template_group/template names.  Then I could replicate.  This is my main testing site, but I uninstalled all third party add-ons, so hopefully nothing weird in play.

I tried replicating on a totally fresh and empty install, made the like 2 templates, a 'dos' site, tried editing the 'Member' role- never could trigger it.  

Oh- and I also tried editing template access permissions in the Template manager for the individual files.  Never could trigger an error that way.

intoeetive:
Resolved #3568 where deselecting all template access for Role in one MSM site would also deselect it on other sites

EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/3577

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Members/Roles/Roles.php
+++ b/system/ee/ExpressionEngine/Controller/Members/Roles/Roles.php
@@ -540,6 +540,11 @@ class Roles extends AbstractRolesController
 
         // template_access
         $template_ids = [];
+        // ensure templates from other sites are always in
+        if (!$role->isNew() && bool_config_item('multiple_sites_enabled')) {
+            $template_ids = $role->AssignedTemplates->filter('site_id', '!=', ee()->config->item('site_id'))->pluck('template_id');
+        }
+        // add posted template IDs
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
@intoeetive MSM templates look good in testing.

PR #3593, opened 2023-07-10
Implemented check for duplicate group names in Template Manager

intoeetive:
image

EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/3602

### DIFF
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -2230,6 +2230,9 @@ class EE_Template
         ee()->db->select('group_id');
         ee()->db->where('group_name', ee()->uri->segment(1));
         ee()->db->where('site_id', ee()->config->item('site_id'));
+        // always get the first template group to avoid dealing with duplicates
+        ee()->db->order_by('group_id', 'asc');
+        ee()->db->limit(1);
         $query = ee()->db->get('template_groups');
 
         // Template group found!
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -4324,6 +4324,18 @@ class EE_Template
             return false;
         }
 
+        // if site_id not known (on update), get it from site short name
+        if (empty(ee()->config->item('site_id'))) {
+            $site = ee('Model')->get('Site')
+                ->filter('site_name', ee()->config->item('site_short_name'))
+                ->first();
+            if (!is_null($site)) {
+                ee()->config->set_item('site_id', $site->getId());
+            } else {
+                return false;
+            }
+        }
+
         ee()->load->library('api');
         ee()->legacy_api->instantiate('template_structure');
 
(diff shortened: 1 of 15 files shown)


robinsowell:
@intoeetive I like the look and the warning, and I'm approving, because I can't really break it.  But I do want to bring one thing up in case you see a potential issue I am missing.

When you edit/delete a template group in the cp, or click the link to view the templates in that group, it's using the template name in the url and not the id number.

If I have a duplicate template group- say 'layout' and I click the one that's lower in my sidebar and that has the duplicate warning on it to delete- it's actually deleting the first group.  I suspect because the names are identical, it's just killing the first one.  This is try if I try to view the templates by clicking the group link- I click the second layout but once I'm viewing the templates, the active one is the first 'layout'.

Note BOTH 'layout' groups share the same folder.  If I add a new file to the layout folder, it goes in the database, but associated with the second 'layout' group only.  And now if I click the link in the cp to view the templates in layout, I don't see my new file.  

In this screen cap, you can see I have a 'layout' group with group ids 3 and 42.  I stuck a new template in the layouts template folder.  Visited the cp.  That new template was created and the group_id associated with it is 42.  I... don't see it in the cp, because in the cp I am seeing/edit the info for group_id 3.

Screenshot 2023-07-10 at 1 36 23 PM

If I delete the template group via the cp, I delete template group 3, regardless of when one I'm next to when I click the trash can.  That's... actually good, because if I made any new templates, they're assigned to id 42.

This is a very long way of saying I can't actually break anything using this to delete duplicate template groups.  But there is a level of complexity going on here I wanted to point out in case you do see a potential issue.

robinsowell (dismissed):
Approving, but see comments for a discussion of findings.  I couldn't break anything so... approved.

intoeetive:
@robinsowell pushed an update that should resolve the issues.

Now when the files are added, they are synchronized into the oldest template group. 
When the groups are viewed / edited / removed, the group_id is now also passed in URL, so people can still make the changes when there are duplicates.

I don't really like the fact that we now pass BOTH group_name and group_id in the URL, but that's the only way we can make that work without breaking existing bookmarks people might have, because group_name can also be numeric

robinsowell (approved):

robinsowell:
@intoeetive Testing the template manager stuff and it works much better.  Added files do go into the oldest, the oldest group keeps its settings.  Deleting the newer one actually deletes the newer one, and it is clearing out the exp_templates db table.

My only lingering concern would be for people who already had this problem and who might have templates assigned to the 'new' group already.  Not sure if that's a possible issue or not.

But aside from that potential concern, all is working well for me now and I can clean out the dupes without it causing any issues for me.

system/ee/legacy/libraries/Template.php
```diff
@@ -4330,6 +4345,7 @@ public function sync_from_files()
             $groups = ee('Model')->get('TemplateGroup')
                 ->with('Templates')
                 ->filter('site_id', ee()->config->item('site_id'))
+                ->order('group_id', 'desc') // we need older groups in case of duplicate names
```
robinsowell, line 4348: mediate duplicate issue frontend

PR #3614, opened 2023-07-19
Move the site_id and site_short_name check earlier into the code

intoeetive:


system/ee/ExpressionEngine/Library/Advisor/Advisor.php
```diff
@@ -17,9 +17,18 @@ public function postUpdateChecks()
     {
         // if we call this from updater, site_short_name might be not set
         // grab the first site and set it
-        if (empty(ee()->config->item('site_short_name'))) {
-            $siteQuery = ee('db')->select('site_name')->from('sites')->order_by('site_id', 'asc')->limit(1)->get();
-            ee()->config->set_item('site_short_name', $siteQuery->row('site_name'));
+        if (empty(ee()->config->item('site_short_name')) || empty(ee()->config->item('site_id'))) {
```
bryannielsen, line 20: This looks good, any chance we can simplify it by going back to what you had earlier with grabbing the first Site without any where clauses?

```
        if (empty(ee()->config->item('site_short_name') || empty(ee()->config->item('site_id'))) {
            $siteQuery = ee('db')->select('site_id, site_name')->from('sites')->order_by('site_id', 'asc')->limit(1)->get();
            ee()->config->set_item('site_short_name', $siteQuery->row('site_name'));
            ee()->config->set_item('site_id', $siteQuery->row('site_id'));
        }
```

We can do the `$siteQuery->num_rows()` check if you think it's necessary but an install without a row here is probably pretty rare and problematic in lots of other ways?

bryannielsen:
I'm going to close this PR because I made these changes in #3618 after reverting another PR so the history is a little cleaner.

PR #3615, opened 2023-07-19
Enabled synchronizing upload directory via CLI; #3044

intoeetive:
Enabled synchronizing upload directory via CLI; #3044

User Guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/702

### DIFF
--- a/system/ee/ExpressionEngine/Cli/Cli.php
+++ b/system/ee/ExpressionEngine/Cli/Cli.php
@@ -142,6 +142,7 @@ class Cli
         // Sync
         'sync:conditional-fields' => Commands\CommandSyncConditionalFieldLogic::class,
         'sync:file-usage' => Commands\CommandSyncFileUsage::class,
+        'sync:upload-directory' => Commands\CommandSyncUploadDirectory::class,
 
         // Update
         'update' => Commands\CommandUpdate::class,
--- a/system/ee/ExpressionEngine/Cli/Commands/CommandSyncUploadDirectory.php
+++ b/system/ee/ExpressionEngine/Cli/Commands/CommandSyncUploadDirectory.php
@@ -1,0 +1,146 @@
+<?php
+/**
+ * This source file is part of the open source project
+ * ExpressionEngine (https://expressionengine.com)
+ *
+ * @link      https://expressionengine.com/
+ * @copyright Copyright (c) 2003-2023, Packet Tide, LLC (https://www.packettide.com)
+ * @license   https://expressionengine.com/license Licensed under Apache License, Version 2.0
+ */
+
+namespace ExpressionEngine\Cli\Commands;
+
+use ExpressionEngine\Cli\Cli;
+
+/**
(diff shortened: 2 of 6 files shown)


bryannielsen:
@intoeetive is there any reason we wouldn't want to allow this to be called without specifying a specific upload location?  I'm thinking if someone needed to call this on a cron job it might be helpful to not require calling each upload location as a separate command.  Also as locations are added/removed any scripting would need to update too.  Just a thought, we don't necessarily need to tackle that now.

intoeetive:
@bryannielsen we're making it work same way as it works in the CP. It might make sense toy have it also worrk globally (loop through all upload locations) but I'd say that can be separate FR

bryannielsen (approved):

matthewjohns0n:
I'm having a bit of a hard time with this for several reasons.

1. I don't currently have any manipulations. It shouldn't be required if there are no manipulations available.

2. I have missing images. Because the images are missing, the CLI is throwing a bunch of errors that are not being caught. In the CP, you can run the Sync even when there are missing images and it allows it to run. Here is a screenshots of errors:

Screenshot 2023-11-15 at 6 07 29 PM

3. I tested with an S3 adapter and got these errors:

Screenshot 2023-11-15 at 6 07 10 PM

4. This command is not possible to run without user interaction if the user does not want to use manipulations. If you do not provide the `--manipulations` flag, it asks you to provide manipulations, which you can skip. However, you can not skip it without the interaction. This is a problem because this command will most likely be scripted. 

5. I agree that we should add a `--all` flag, which just loops over the upload directories and syncs all of them.

bakin1999:
image

image
yeah I only got the error about the sizing. $allSizes is an empty array at this point.  Is the same as the first error on matt's screenshot as well.


As for Matt's number 4.  You can just run this with all manipulations passed and it will not require an interaction. Like this:
php eecli.php sync:upload-directory --manipulations=all   This works without manipulations but runs into it not finding the array size key

* I never saw Matt's number 2 issue come up however, I removed a file, but never got it to show that error yet.

intoeetive:
@matthewjohns0n I think the error is actually because you have {base_path} used in server path, and it's set to DOCUMENT_ROOT or similar, so you get wrong path. Since this is common case, I added error check.

Also the other issue are resolved now.

Please test again

bakin1999:
Hmm well upgrading dp versions my command does not work at all now.

The output of the command
image

Settings for the blog upload directory
image

Settings for my base path that is somehow not being found now
image


This same command worked on DP 17 but not on DP 18 and up. Somehow it is falling into the new error check part.  

P.s. Oh, the screenshot shows just id 1 but I tried this with the blog id as well and got same error message

intoeetive:
@bakin1999 you're absolutely right. I was so happy about my fix that made it way more strict than it should be :) All should be good now

bakin1999:
great is working again!

PR #3627, opened 2023-07-21
Added Utility and post-update check for duplicate Template Groups

intoeetive:
User guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/718

This PR is adding post-update check for duplicate template groups, as well as stand-alone utility to check and fix.

If duplicates are found, message is shown post-update, same as with other checks.

The Utility lists all duplicates and offers renaming and removing any of those - they can pick the earlier one or the later one.

I had to add support for group_id to some functions in template manager (edit and remove) so that we could address exactly the groups selected.

The template manager is listing only the earlier template groups (the ones that are used on front-end) - but also display banner if duplicates exist letting them know they need to use utility. It also adds record to developer log

Upon deleting, they are brought back to the list of duplicates; upon editing they are brought to template manager (this is not easy to change so I'd like to keep even if that's not ideal)

Partailly based on work in https://github.com/ExpressionEngine/ExpressionEngine/pull/3593

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Design/AbstractDesign.php
+++ b/system/ee/ExpressionEngine/Controller/Design/AbstractDesign.php
@@ -104,7 +104,13 @@ abstract class AbstractDesign extends CP_Controller
             }
         }
 
+        $groupNamesListed = [];
+
         foreach ($template_groups->all() as $group) {
+            if (in_array($group->group_name, $groupNamesListed)) {
+                // duplicates found, show alert but do not add to sidebar
+                continue;
+            }
             $item = $template_group_list->addItem($group->group_name, ee('CP/URL')->make('design/manager/' . $group->group_name));
 
             $item->withEditUrl(ee('CP/URL')->make('design/group/edit/' . $group->group_name));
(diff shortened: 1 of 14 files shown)


robinsowell:
Worked fine for me, but deleting a bunch of them was tedious.

UI wise, I'd have preferred checkboxes, so I could just check off the ones I want deleted and submit once.  Of course, that would require a check/warning if they were deleting ALL of a certain name. 'You will no longer have a group called 'blogs' are you sure you want to delete them all' sort of thing.

Other nice things- if the duplicates have an actual difference in the settings OR the templates assigned to them, indicate that in some way.  Or even, have some extra ui so the 'groups' of dupes stand out (alternate row colors between groups of groups or something) and a column to indicate if there is a difference in the templates in the groups and a column to indicate a difference in the group settings.

I had a lot of them, didn't really want to manually compare each one, but wasn't sure if deleting one was going to do 'very bad things'.  Some way to alert folks of the likelihood of 'very bad things' would be helpful.

robinsowell (dismissed):
Suggested some changes, but wouldn't object to be overruled.

robinsowell (approved):
As long as they backup first- and we do clearly warn to do that- I think it all looks good.

Issue #2564, opened 2022-11-26
File manager performance


shbchk:
EE 7.2.1
Over 10 000 images in the system, stored locally, most of them in one directory.
Compatibility Mode is off.

It takes almost half a minute to load the file manager (in modals too):

![649258b2-a070-41da-a1be-c67a50b6ac18](https://user-images.githubusercontent.com/12234507/204106352-cf9a0a7f-30a6-4013-b84c-d6f43b26b0d1.png)

shbchk:
Well, I guess it is time to give a try to the new Cloud Files adapter and move those thousands of files to Digital Ocean Spaces :)

shbchk:
Oh my god, how slow those are! I never tried S3 before and thought it might be faster, but it's really not.

When I tried to open the Sync Directory page for a directory of 10000 images that I've uploaded to Digital Ocean Spaces (_it was very long and painful_), I got this error:

`Fatal error: Maximum execution time of 300 seconds exceeded in C:\OpenServer\domains\vurchel.local\sstm\user\addons\cloud_files\vendor-build\guzzlehttp\guzzle\src\Handler\CurlMultiHandler.php on line 131`

intoeetive:
@shbchk can you please try the change in linked PR and tell us if this works for you.

This change is for Local adapter only. Unfortunately we don't have good solution currently for cloud storage having thousands of files in single upload directory

shbchk:
@intoeetive it worked great for me! Thank you!

shbchk:
@intoeetive Hi, Yuri! Should I open a new ticket for this issue with cloud storage and a large number of files in a single directory? Is there a chance for this to be fixed?

intoeetive:
I believe @bryannielsen is working on update for Cloud Files that should work better

shbchk:
That's great! Thanks, I will wait. I hope waiting will not be long :)

shbchk:
@bryannielsen hi! I can't yet transfer all my tens of thousands of image files to S3, right?

https://github.com/ExpressionEngine/ExpressionEngine/pull/3630 - we're waiting for this?

PR #3630, opened 2023-07-24
Allow filesystem to eager load select paths

bryannielsen:
Update filemanager to use this functionality instead of attempting to eager load the entire contents.  The old behavior was problematic with remote filesystems containing many files.  Related to #2564

This change allows adapters to implement functionality in an `eagerLoadPaths($paths = [])` method when applicable.

CachedAdapter contains a lot of duplicate code from Flysystem's `Cached\CachedAdapter` which it extends because the parent's private properties do not lend themselves to a simpler subclass.


system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
```diff
@@ -293,22 +293,38 @@ function ($key) {
         $data = array();
         $missing_files = false;
 
-        $destinationsToEagerLoad = [];
+        // Group files by destination and eager load first
+        $eagerLoads = array_reduce($files->asArray(), function ($carry, $file) use ($member) {
+            if ($file->UploadDestination->adapter != 'local' && $file->UploadDestination->exists()) {
+                if(!array_key_exists($file->upload_location_id, $carry)) {
```
intoeetive: Would it be better to swap these lines? I think most of the times the files would be sharing upload destination, so checking whether we already have the info before going into UploadDestination details should be more effective. Also might be worth checking if UploadDestination is not null - for edge cases when the folder record has been removed, but the file record stays - just so it wouldn't bomp out

system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
```diff
@@ -293,22 +293,38 @@ function ($key) {
...
+            $files = array_filter($files, function ($file) {
+                return !$file->isDirectory();
+            });
+
+            foreach($files as $file) {
+                $paths[] = $file->getAbsolutePath();
+                $paths[] = $file->getAbsoluteManipulationPath('thumbs');
+            }
+
+            $uploadDestination->getFilesystem()->eagerLoadPaths($paths);
```
intoeetive: do we need to check if `eagerLoadPath` is available method?

intoeetive (commented):
@bryannielsen - see the couple of comments I had

system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
bryannielsen: I think this is fine the way it is, that line 298 is just the old line 306 (https://github.com/ExpressionEngine/ExpressionEngine/pull/3630/files/6135f94929ff29b3d01e8a8b44d200d8a65e8b75#diff-cd5b889f5e6fcff6954cdafc0347f4ecdc532ad6e03ce5c5d115ce956fc32644L306) and the UploadDestination relationship gets eager loaded so the performance impact should be trivial.  We would also have to do this same check later when $carry hasn't been set yet and I think this way is a little more readable.

system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
bryannielsen: The filesystem has an implementation for eagerLoadPaths() so we don't need to check here, but inside that function it checks to see whether the current adapter has an implementation before calling it

intoeetive (changes requested):
When using CloudFiles 1.0 (did not test 1.1) the subfolders don't show up. They work fine on current 7.dev

bryannielsen:
Good catch Yuri, this latest push should resolve that issue

intoeetive (changes requested):
@bryannielsen on my test, I has S3 directory with 5 images and a subfolder. Without your changes, the load time is ~3 sec. With your changes, it's 5-6 sec. Tested multiple times, numbers staid the same. So IDK if it really makes any good, at least without hundreds of files in the directory (which is rather edge case).
I don't think we should make it better for 5% of people and worse for everyone else.
Also - unrelated - your code is missing some spaces (see PSR-12 notes on Files tab)

bryannielsen:
@intoeetive did you upgrade to the latest Cloud Files?  The performance should be better and more consistent in 1.1.0+ whether you have 5 or 5,000 files in a bucket.  Without this change and the updated version of Cloud Files we can have control panel timeouts on the file listing page.  The experience may be slightly degraded with older versions of Cloud Files that aren't setup to run parallel requests.

I think I can make a change to eagerLoadPaths() so that if an adapter does not support this we do the old eager load behavior of listing the entire directory.  Which will preserve better load times for smaller buckets but still break things on large buckets.

intoeetive:
@bryannielsen I updated to CP 1.1 but the picture did not change.

I then tried uploading more files (~80) and the load time increased for current 7.dev (4 sec) but still remains 5-6 sec when using this branch.
So perhaps this change only makes difference when you have large number of files (100+). If that is so, I think we need to check  for files number and only apply this method for that case

intoeetive (dismissed):
thinking about if, if people are using S3, they probably have large number of files - otherwise they have no reason to use it

so this change is probably fine, as long as it does not affect local adapter

matthewjohns0n:
My tests were consistent with Yuri's. I have 13 images in an S3 bucket. (Cloud Files 1.1.0).

`7.x`: ~1.9s average
this branch: 3.0s average

bryannielsen:
Interesting, I'll add mine here too for comparison

28 files in S3
7.x - 1.2 seconds
this branch - 1.5 seconds 

@robinsowell maybe we can bother you for another data point

robinsowell:
My results don't show a lot of difference between the two, but the older version is a touch faster on average.  Mostly focused on my largest directory- around 200 image files.  

'Finish' old version ~ 1.8
'Finish' new version ~ 2.2s

New was faster a few times, tried in FF and FF incognito.  So.... IDK.

robinsowell:
Just pondering, I keep browser cache disabled when testing.  I wonder if that messes with the results.

robinsowell:
@bryannielsen totally unrelated to testing this, I put a channel form on the frontend to test some wygwam thing and I'm getting a php memory error- my memory set to 128.

```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /Users/robinsowell/Sites/738/system/user/addons/cloud_files/vendor-build/aws/aws-sdk-php/src/data/s3/2006-03-01/endpoint-rule-set-1.json.php on line 6```

robinsowell:
Circling back, it looks like it's only happening with my wygwam field and only in the channel entry form.  And it happens in the currently released v7 as well.  SO nothing to do with this pr.  Looks like a wygwam issue.

intoeetive:
@bryannielsen there's now a code conflict

PR #3640, opened 2023-07-28
Resolved #3634 where having MSM-shared upload directory would cause Pages to be saved to wrong site

### LINKED ISSUE #3634: MSM and Pages / Structure saving new pages data to wrong site
First report was issue with Pages module, but I tested and Structure is also affected

At first I thought this was a validation issue.  I have a page_url on site 'a' 'my_test_url'.  I then go to site b, make a new entry, put the same uri in.  That should be allowed across sites, but it's coming back with a validation error that the uri is a duplicate.

On further investigation- NON duplicate uris don't appear to be saved.

On further further investigation, uris on site 'b' ARE saved, but they are saved under site 'a'.  This would also explain the duplication warning- they really are duplicates because my site 'b' url is attempting to save in site 'a'.

Debugging it- in tabs.pages.php- if I am on site 'b' and I output the $site_id in the display function, it is correct:

```
    public function display($channel_id, $entry_id = '')
    {
		
		//var_dump(ee()->config->item('site_id'));
```

This is also why it appears site 'b' isn't saving the uri- it saved it to site 'a' it does not exist in site 'b' and so there is nothing to display.

If I dump the site id in ``function prepareSitePagesData`` it shows the wrong site- site 'a'.

Note- this test was done with all validation commented out.  So I'm going to go undo that and test more.  But user was getting the behavior on a totally clean copy.

STRUCTURE

So with Structure, if you're on site b, it's not saving the page uri to the wrong site, but it's not saving it at all, at least not in the pages array and it's not reflected in the entry display in the cp.  It might LOOK like it saves it at first, because if it's not set, it apparently populates with the url title.  But try editing the page uri to something else and... it doesn't stick.  Also, if you look at the raw data, saving on site a does go into exp_sites and saving on site b does not:
Screenshot 2023-07-25 at 12 48 17 PM

intoeetive:
Resolved #3634 where having MSM-shared upload directory would cause Pages to be saved to wrong site

### DIFF
--- a/system/ee/legacy/models/file_upload_preferences_model.php
+++ b/system/ee/legacy/models/file_upload_preferences_model.php
@@ -114,7 +114,7 @@ class File_upload_preferences_model extends CI_Model
         $site_id = $upload_destination['site_id'];
         $overrides = array();
 
-        if ($site_id != ee()->config->item('site_id')) {
+        if ($site_id != 0 && $site_id != ee()->config->item('site_id')) {
             $overrides = ee()->config->get_cached_site_prefs($site_id);
         }
 


robinsowell (approved):
Fixed it, confirmed by user as well.

PR #3641, opened 2023-07-28
Resolved #3639 where is was not possible to save loaded revision with removed Grid rows

### LINKED ISSUE #3639: Grid validation can throw a 'not authorized' message trying to save an old version
Replicated on 7.3.8, only happens to logged in superadmin

Enable versioning for a channel.  Go make an entry with a grid field.  Save it.  Add a row, save.  It will go into exp_entry_versioning like this:

```s:10:"field_id_7";a:1:{s:4:"rows";a:1:{s:8:"row_id_2";a:2:{s:9:"col_id_11";s:0:"";s:9:"col_id_12";s:0:"";}}}s:10:"```

Edit the entry, delete the row, save it.

Load the version that has the saved grid field, try to save, now you get a validation error and basically no way to get rid of it rather than load a new version.

Screenshot 2023-07-27 at 2 24 33 PM

When you deleted the row, it removed it from exp_channel_grid_field_x and in grid/libraries/Grid_lib.php around line 250:

```
       if (isset($data['rows'])) {
            foreach ($data['rows'] as $key => $row) {
                if (substr($key, 0, 6) == 'row_id') {
                    $row_key = str_replace('row_id_', '', $key);

                    if (! in_array($row_key, $valid_rows)) {
                        if (ee('Permission')->isSuperAdmin()) {
                            return array('value' => '', 'error' => lang('not_authorized'));
                        } else {
                            unset($data['rows'][$key]);
```

If you're a superadmin and it's submitting a row that no longer exists, you get a not authorized error.  Otherwise, it just deletes it.  I'm not sure why it's different for superadmin, but it's definitely confusing.  

Note- IF before you save, you delete the orphaned row, you can save then.  Basically, you manually clear it out and your good.  But otherwise, once you save, you get the validation error, there's no trashcan to delete it and you basically have to refresh the page or load another version. 

ETA: SOOOO I thought the solution would be to just unset the data like we do for non-superadmins, but then I realized that for versions, that means you lose the data in the grid if the row was deleted in another version.  This is exactly what versions DO NOT WANT.

Solution- if the data in the version array looks like this:
```
s:11:"field_id_23";a:1:{s:4:"rows";a:1:{s:9:"new_row_1";a:2:{s:9:"col_id_39";s:0:"";s:9:"col_id_40";s:0:"";}}}s:11
```
It's ok.  EE is treating it like a new, unsaved row and not trying to match it to an existing column.  Maybe when we either save the version or more likely when we unserialize the array and set it up, we could treat either all or any missing grid rows as 'new_row_x' instead of 'row_id_1'?

intoeetive:
Resolved #3639 where is was not possible to save loaded revision with removed Grid rows

I think it does not quite make sense to sinlently remove Grid rows if the revision was loaded with those rows - so the goal probably was to restore those.

This PR is also making slight change to revisions UI - we previouly kept displaying "you are viewing a revision..." banner even after saving - which wasn't true; and the button to load the current version when you're on revision wasn't active.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/grid/libraries/Grid_lib.php
+++ b/system/ee/ExpressionEngine/Addons/grid/libraries/Grid_lib.php
@@ -250,14 +250,21 @@ class Grid_lib
         }
 
         if (isset($data['rows'])) {
+            $total_rows = count($data['rows']);
             foreach ($data['rows'] as $key => $row) {
                 if (substr($key, 0, 6) == 'row_id') {
                     $row_key = str_replace('row_id_', '', $key);
 
                     if (! in_array($row_key, $valid_rows)) {
-                        if (ee('Permission')->isSuperAdmin()) {
-                            return array('value' => '', 'error' => lang('not_authorized'));
+                        if (ee('Request')->get('version')) {
+                            //if we have loaded version, restore the fields that are missing
+                            $keys = array_keys($data['rows']);
+                            $values = array_values($data['rows']);
+                            $index = array_search($key, $keys);
+                            $keys[$index] = 'new_row_' . $total_rows + (int) $row_key;
+                            $data['rows'] = array_combine($keys, $values);
                         } else {
+                            // otherwise assume someone else removed the row, and they know what they are doing
                             unset($data['rows'][$key]);
                         }
                     }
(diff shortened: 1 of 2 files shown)


robinsowell (approved):
Boom, totally worked.  AND SO MUCH BETTER TO NOT LOSE THE DATA!

Thanks Yuri!

intoeetive:
I'll have to close this in favor of more complex solution: https://github.com/ExpressionEngine/ExpressionEngine/pull/3659

PR #3652, opened 2023-08-01
Resolved #3636 where searching in File Manager while having Upload Directory selected did not include files in subfolders

### LINKED ISSUE #3636: Searching in File Field type with Subfolder - No results
Using a File Fieldtype, there is no way to search for files within subfolders. I'm not sure if it's a bug or a feature that needs to be added. It becomes cumbersome to find files when they are located multiple levels deep in subdirectories.

**Environment Details:**
 - Version: [7.2.8]
 - PHP Version [8.0.27]

intoeetive:
Resolved #3636 where searching in File Manager while having Upload Directory selected did not include files in subfolders

Ideally I would want to list the subfolders when when a file is found, but that would require more complex and expensive Model queries
This PR is just listing found files without mentioning which subfolder those are in - and this is the same behavior that currently exists on File Manager homepage, so we're probably fine with it.

### DIFF
--- a/system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
+++ b/system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
@@ -72,7 +72,6 @@ trait FileManagerTrait
         //limit to subfolder, show breadcrumbs
         if (! empty($uploadLocation)) {
             $directory_id = (int) ee('Request')->get('directory_id');
-            $files->filter('directory_id', $directory_id);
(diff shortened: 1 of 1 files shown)


gamebreakers808:
Is this related to the inability to see a new sub folder in the file manager after creating it? I'm using 7.3.8. Thanks!

intoeetive:
@gamebreakers808 I'm not quote sure what is the issue that you have. Please create separate bug report with exact steps how to replicate (note that the folders created on filesystem would not appear in EE filemanager automatically; you need to run upload directory sync for them to appear)

matthewjohns0n (approved):

PR #3665, opened 2023-08-04
Resolved #3663 where subdirectories were assigned to wrong site when using MSM

### LINKED ISSUE #3663: File manager- subfolder doesn't show in MSM secondary sites
Odd one!  So with MSM, on my primary site, I create a subfolder and it shows up properly.

I go to a secondary MSM site, create a subfolder (regular folder, not global, set to allow subfolders)- it says it created it, but it doesn't show up in the list.  It DOES show up if I click the 'upload' button where it shows a list of folders and I can even upload to it that way.  It's just not displaying in the file manager.

7.3.8.

intoeetive:
Resolved #3663 where subdirectories were assigned to wrong site when using MSM


intoeetive:
@robinsowell would you mind testing? For existing folders, you'd need to run update to 7.3.9
I think would be also nice to test this if table prefix is different from exp_

bryannielsen (dismissed):

bryannielsen (approved):

robinsowell:
@intoeetive Just finished testing- did use non-exp prefix.  Upgrade got everything showing in the right place.  There is one bug- I'm not sure it's due to this or not.  The subfolders when empty show that 'no permission' message in the dropdown area.  This is true of existing subfolders and new subfolders, happens in the default side and the add-on sites.

Screenshot 2023-08-09 at 10 02 39 AM

Note- I can actually upload to the subfolder using the dropdown and once I have one in there, drag and drop is fine.  The message is just wrong.

robinsowell (commented):
Have a display issue with empty subfolders showing you don't have permission to upload.

intoeetive:
@robinsowell we've got that issue with message on subfolders fixed in other PR, which is already in 7.dev (but not in this branch)

PR #3672, opened 2023-08-08
Resolved #3667 where all members with access to Template Manager could access System Templates

### LINKED ISSUE #3667: EE6 - Limited access to template variables gives unlimited access to system templates
If you give a member role access to just "template variables" when they view CP they also see the "Template Group" listing (reporting no templates found) and the system templates (which it appears they can edit)

image

**Environment Details:**
 - Version: 6.4.13
 - PHP Version 8.0.29

**Possible Solution**
Not showing the redundant elements / not giving edit access to system templates.

intoeetive:
Resolved #3667 where all members with access to Template Manager could access System Templates

Because this is permissions issue that we need to fix in patch release, I am making System Templates access dependant on "Global template settings" permission (which already is giving access to template routes)

As we make updates to template manager in the future, we need to make these separate permission settings.

EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/3673


intoeetive:
@bryannielsen I have updated the checks. Also removed check whether Member odule is installed when using legacy member template - since that is built-in module that can't be removed, that's hardly the possible case. Same updates done for EE6

bryannielsen (approved):

PR #3678, opened 2023-08-09
Added 2 Structure settings to customize Pages UI

swierczek:
Added 2 new settings (at the bottom) to toggle elements on the Pages UI:

image

Default view:

image

Status displayed, "Hide from nav" hidden:

image

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/views/index.php
+++ b/system/ee/ExpressionEngine/Addons/structure/views/index.php
@@ -145,8 +145,17 @@ if ($cp_asset_data or count($tabs) > 1) {
                     }
                 }
 
-                if ($page['hidden'] == 'y') {
+                if (isset($settings['show_entry_status']) && $settings['show_entry_status'] == 'y') {
+                    $status = ee('Model')->get('Status')->filter('status', '==', $page['status'])->first();
+                    echo $status->renderTag();
+                }
+
+                if (isset($settings['show_nav_status']) && $settings['show_nav_status'] == 'y' && $page['hidden'] == 'y') {
                     echo ' <span class="hidden-page">(hidden)</span>';
+                }
+
+                if (isset($page['extra_html']) && $page['extra_html']) {
+                    echo ' <span class="extra-html">'.$page['extra_html'].'</span>';
                 }
 
                 echo '</span>', "\n";
--- a/system/ee/ExpressionEngine/Addons/structure/views/index.php
+++ b/system/ee/ExpressionEngine/Addons/structure/views/index.php
@@ -146,8 +146,7 @@ if ($cp_asset_data or count($tabs) > 1) {
                 }
 
                 if (isset($settings['show_entry_status']) && $settings['show_entry_status'] == 'y') {
-                    $status = ee('Model')->get('Status')->filter('status', '==', $page['status'])->first();
-                    echo $status->renderTag();
+                    echo isset($status_tags[$page['status']]) ? $status_tags[$page['status']] : '<span class="status-tag st-'.str_replace(" ", "-", strtolower($page['status'])).'" style="">'.$page['status'].'</span>';
                 }
 
                 if ((!isset($settings['show_nav_status']) || $settings['show_nav_status'] == 'y') && $page['hidden'] == 'y') {
--- a/system/ee/ExpressionEngine/Addons/structure/views/index.php
+++ b/system/ee/ExpressionEngine/Addons/structure/views/index.php
@@ -146,7 +146,7 @@ if ($cp_asset_data or count($tabs) > 1) {
                 }
 
                 if (isset($settings['show_entry_status']) && $settings['show_entry_status'] == 'y') {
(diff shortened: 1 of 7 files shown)


system/ee/ExpressionEngine/Addons/structure/views/index.php
```diff
@@ -145,10 +145,19 @@
...
+                if (isset($settings['show_entry_status']) && $settings['show_entry_status'] == 'y') {
+                    $status = ee('Model')->get('Status')->filter('status', '==', $page['status'])->first();
+                    echo $status->renderTag();
+                }
+
+                if (isset($settings['show_nav_status']) && $settings['show_nav_status'] == 'y' && $page['hidden'] == 'y') {
                     echo ' <span class="hidden-page">(hidden)</span>';
                 }
 
+                if (isset($page['extra_html']) && $page['extra_html']) {
```
swierczek, line 156: Added this as a way to use the `structure_index_view_data` hook to add even further UI elements as needed. For instance, additional Publisher tags:

image

system/ee/ExpressionEngine/Model/Status/Status.php
```diff
@@ -141,6 +141,8 @@ public function getSelectOptionComponent($use_ids = false)
 
     public function renderTag()
     {
+        ee()->lang->loadfile('admin_content');
```
swierczek, line 144: Without this, the status tag is rendered as `open` and `closed`, not `Open` and `Closed`.

system/ee/ExpressionEngine/Addons/structure/views/index.php
```diff
@@ -145,10 +145,19 @@
                     }
                 }
 
-                if ($page['hidden'] == 'y') {
+                if (isset($settings['show_entry_status']) && $settings['show_entry_status'] == 'y') {
+                    $status = ee('Model')->get('Status')->filter('status', '==', $page['status'])->first();
+                    echo $status->renderTag();
+                }
```
intoeetive: I think this part needs to be optimized. If you have 50 entries, that would mean 50 extra queries, most of which will be duplicates. It should be a lot better to eager load statuses before sending variables to view file (and then maybe run `array_map` on result while calling `renderTag` to get extra variable for status badge)
Also there can be cases where status has been removed from system, but entries still exist with that status - we need to make sure to not break things if that's the case.

system/ee/ExpressionEngine/Addons/structure/views/module_settings.php
```diff
@@ -49,6 +49,24 @@
                 </select>
             </td>
         </tr>
+        <tr class="odd">
+            <td><?=lang('show_nav_status')?></td>
+            <td>
+                <select name="show_nav_status">
+                    <option value="y"<?=set_select('yes', 'Yes', ($settings['show_nav_status'] == '' || $settings['show_nav_status'] == 'y' ? 'y' : ''));?>><?=lang('yes')?></option>
```
intoeetive, line 56: This defaults to "yes", but the rest of code does not seem to respect that. By default, the nav status is not show (which I think it should be) - however when I visit module settings, it shows as "yes", which is not correct

intoeetive (changes requested):
@swierczek Thank you! I think this is great addition.

I noticed however it does not look good UI-wise, not like on your screenshots - perhaps some changes were not pushed? 
image

Please also see other comments that I posted.

Additionally, would you mind updating patch version number in addon.json file?

system/ee/ExpressionEngine/Addons/structure/views/index.php
swierczek: I can absolutely optimize this by getting all Statuses outside of the loop and then referencing them here, but how would you suggest handling cases where the status has been removed, but entries still exist with that status? How do other parts of the CMS handle this?

intoeetive: Here's one example (not eager loading here though) :
```
!empty($entry->getStatus()) ? $entry->getStatus()->renderTag() : $entry->status
```

swierczek: Before (Status exists):
image

After (Status no longer exists, but entry is still assigned to it):
image

Is this a sufficient fix for missing Statuses? I added "New one" as a new Status, assigned the entry to it, then deleted the status, and now the UI appears like this. Alternatively, I could output it as just text like the Entries list does:

image

But I think the tag style here helps differentiate it in the UI. Maybe I'll add some default tag styling to structure.css to make it black text/border instead.

swierczek: Here's the updated style if a Status no longer exists. I think this is slightly cleaner for UI consistency:

image

swierczek:
I believe all requested changes have been fixed now. I added a cache busting parameter to structure.css, which I suspect is the reason you weren't seeing updated styles.

intoeetive (approved):
@swierczek I think this is good to go now - thank you! We'll probably experiment a bit with the UI before including it, but I do hope we'll be able to release this ASAP.

I believe this is great addition!

swierczek:
I accidentally deleted my forked repo before this was merged. What is the best way to proceed to get this merged? Should I recreate the PR in a new fork and resubmit, or are you able to make these updates some other way based on this current PR?

PR #3700, opened 2023-08-14
add markers to identify the partial name

intoeetive:
https://github.com/ExpressionEngine/ExpressionEngine/discussions/2867

### DIFF
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -403,6 +403,8 @@ class EE_Template
                             'Snippet "' . $variable . '"'
                         );
 
+                        $replace = $this->addSourceMarker($replace, $variable, $added_globals['template_type']);
+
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/Template.php
```diff
@@ -403,6 +403,8 @@ public function parse(&$str, $is_embed = false, $site_id = '', $is_layout = fals
                             'Snippet "' . $variable . '"'
                         );
 
+                        $replace = $this->addSourceMarker($replace, $variable, $added_globals['template_type']);
```
intoeetive, line 406: Because this has potential to break templates (e.g. comment inside a comment, or when partial is used as a parameter for HTML tag), we need more precise control here.

At the minimum, we need to add config override and only add markers when it's set. Ideally, we might want to be able to control this on template level, like we do with front-end editing tags.

Or make it better code overall, aware of context and not break things

PR #3702, opened 2023-08-14
Added {exp:member:role_groups} template tag

swierczek:
Add `{exp:member:role_groups}` template tag to get info about which Role Groups a member is assigned to.

Documentation will still be needed if this is accepted, which I am happy to update too. I didn't want to do all of the work if this will be declined by the mods though.

Demo template:
```
{!-- existing functionality --}
<h1>Roles</h1>
{exp:member:roles}
	<p>role ID: {role_id}</p>
	<p>role name: {name}</p>
{/exp:member:roles}

{!-- new functionality --}
<h1>Role Groups</h1>
{exp:member:role_groups}
	<p>role group id: {role_group_id}</p>
	<p>role group name: {role_group_name}</p>
{/exp:member:role_groups}
```

Output for member not assigned to any Role Groups:
image

Output for member in a Role Group:
image

### DIFF
--- a/system/ee/ExpressionEngine/Addons/member/mod.member.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member.php
@@ -2620,6 +2620,42 @@ class Member
     }
 
     /**
+     * Member Role Groups list
+     */
+    public function role_groups()
+    {
+        $member_id = (!ee()->TMPL->fetch_param('member_id')) ? ee()->session->userdata('member_id') : ee()->TMPL->fetch_param('member_id');
+
+        $member = ee('Model')
+            ->get('Member', $member_id)
+            ->with('RoleGroups')
+            ->all()
+            ->first();
+
+        if (!$member) {
+            return ee()->TMPL->no_results();
+        }
+
+        $vars = [];
+        foreach($member->RoleGroups->asArray() as $roleGroup) {
+            if ($roleGroup->group_id === 0 && $roleGroup->name === null) {
+                continue;
+            }
+
+            $vars[] = [
+                'role_group_id' => $roleGroup->group_id,
+                'role_group_name' => $roleGroup->name,
+            ];
+        };
+
+        if (!$vars) {
+            return ee()->TMPL->no_results();
+        }
+
+        return ee()->TMPL->parse_variables(ee()->TMPL->tagdata, $vars);
+    }
+
+    /**
      * Member roles list
      */
     public function roles()
--- a/system/ee/ExpressionEngine/Addons/member/mod.member.php
+++ b/system/ee/ExpressionEngine/Addons/member/mod.member.php
@@ -2637,7 +2637,7 @@ class Member
         }
 
         $vars = [];
-        foreach($member->RoleGroups->asArray() as $roleGroup) {
+        foreach ($member->RoleGroups as $roleGroup) {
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/member/mod.member.php
```diff
@@ -2619,6 +2619,42 @@ public function custom_profile_data($typography = true)
...
+            ->with('RoleGroups')
+            ->all()
+            ->first();
+
+        if (!$member) {
+            return ee()->TMPL->no_results();
+        }
+
+        $vars = [];
+        foreach($member->RoleGroups->asArray() as $roleGroup) {
```
intoeetive: Is there a specific reason to call `asArray` here? `RoleGroups` is already iterable

system/ee/ExpressionEngine/Addons/member/mod.member.php
```diff
@@ -2619,6 +2619,42 @@ public function custom_profile_data($typography = true)
...
+                continue;
+            }
+
+            $vars[] = [
+                'role_group_id' => $roleGroup->group_id,
+                'role_group_name' => $roleGroup->name,
+            ];
+        };
+
+        if (!$vars) {
```
intoeetive: Casting array to boolean is not PHP-version-proof. It should be better to use `empty()` check instead

intoeetive (changes requested):
@swierczek I think it's good addition! We're planning to roll out some Member functionality updates in 7.4, so it might go there as well.

I added a couple of comments, and also the code needs to be made PSR-12 compatible (e.g. a space is needed after foreach)

system/ee/ExpressionEngine/Addons/member/mod.member.php
swierczek: Nope, I didn't realize it was iterable 👍

matthewjohns0n:
This tag isnt working for me - I'm testing with a super admin account and have one role group called "all" which contains all roles. 

This is the output of the example template tag:
Screenshot 2023-11-15 at 3 48 38 PM


When I dig into the code, I'm seeing that during the loop, `$roleGroup->group_id` is 0 and it hits the continue. I'm not sure why I'm not getting the correct role groups here.

system/ee/ExpressionEngine/Addons/member/mod.member.php
```diff
@@ -2619,6 +2619,42 @@ public function custom_profile_data($typography = true)
...
+                continue;
+            }
+
+            $vars[] = [
+                'role_group_id' => $roleGroup->group_id,
+                'role_group_name' => $roleGroup->name,
+            ];
+        };
+
+        if (!empty($vars)) {
```
matthewjohns0n, line 2651: This should be inverted - fixed here: https://github.com/ExpressionEngine/ExpressionEngine/commit/f8bf5acdb9e6fbde67d0d06e0a4d9721a77bfcf1

matthewjohns0n:
After the above fix, it is parsing it as if there is no member role group. This is still incorrect because I should have one member role group. 

Current output: 

Screenshot 2023-11-15 at 3 57 12 PM

intoeetive:
It appears that besides member roles assigned to groups, role groups can be assigned directly to members (don't ask me why! :) - although you can't assign those directly to superadmins.

The template code was only respecting the second, and not the first - so I added that.

@matthewjohns0n please re-test

PR #3710, opened 2023-08-19
Resolved issue where using undefined layout variables in tag parameters would cause unexpected results

aaronkirkham:
This PR fixes an issue where passing undefined layout variables in to template tag arguments would result in `ee()->TMPL->fetch_param` giving back unclean layout variables, this is because `_cleanup_layout_tags` is called after the function which parses template tags.

Example:
```php
// inside our layout template.html
{exp:addon:tag arg="{layout:some_variable_which_doesnt_exist}"}

// then inside our addon Tag.php
dump(ee()->TMPL->fetch_param('arg'));
```

The above would result in an unclean layout variable being output (`string(41) "{layout:some_variable_which_doesnt_exist}"`) instead of an empty string - which forces me to manually check and replace any input passed to my tag which comes from potentially undefined layout variables.

Example 2:
```php
// inside layout template.html
{exp:addon:tag arg="{if '{layout:some_variable_which_doesnt_exist}'}true{if:else}false{/if}"}
```

The above would result in strange behaviour - `truefalse"}` being output if we just return the `fetch_param` back from the tag.

Example 3:
```php
// inside layout template.html
{exp:addon:tag arg="{layout:some_variable_which_doesnt_exist:trim:url_slug}"}
{exp:addon:tag arg="{if '{layout:some_variable_which_doesnt_exist:trim:url_slug}'}true{if:else}false{/if}"}
```
This again would have the same effect as above as, but because these are parsed in a slightly different way, we must have a valid key for `some_variable_which_doesnt_exist` in the `layout_conditionals` array, or the conditionals won't be run.

---

To fix these issues, I'm giving all referenced layout variables (which are undefined) a default value of an empty string. This is done by matching all the layout variables (excluding `layout:contents`), finding out which of the referenced variables are undefined in the `$layout_vars` array, then adding in the variable name with an empty value.

### DIFF
--- a/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
+++ b/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
@@ -408,7 +408,7 @@ class LegacyParser
         if (preg_match_all('/' . LD . 'layout:(?!\bcontents\b|\bset\b)([^!]+?)' . RD . '/', $str, $matches)) {
             foreach ($matches[1] as $match) {
                 // ensure that the match string has a modifier character in it
-                if (($modifier_pos = strrpos($match, ':')) !== false) {
+                if (($modifier_pos = strpos($match, ':')) !== false) {
                     $key = substr($match, 0, $modifier_pos);
 
                     // ignore if layout:variable is already defined
--- a/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
+++ b/system/ee/ExpressionEngine/Service/Template/Variables/LegacyParser.php
@@ -403,25 +403,6 @@ class LegacyParser
     {
         $conditionals = [];
 
-        // get all the left over layout variables in the string (excluding layout:contents and layout:set)
-        // the ones left here (after Template parseLayoutVariables) should only be layout variables with modifiers
-        if (preg_match_all('/' . LD . 'layout:(?!\bcontents\b|\bset\b)([^!]+?)' . RD . '/', $str, $matches)) {
-            foreach ($matches[1] as $match) {
-                // ensure that the match string has a modifier character in it
-                if (($modifier_pos = strpos($match, ':')) !== false) {
-                    $key = substr($match, 0, $modifier_pos);
-
-                    // ignore if layout:variable is already defined
-                    if (isset($vars['layout:' . $key])) {
-                        continue;
-                    }
-
-                    // set the undeclared (but referenced) variable to an empty string
-                    $vars['layout:' . $key] = '';
-                }
-            }
-        }
-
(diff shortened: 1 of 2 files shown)


intoeetive:
@aaronkirkham thank you for your submission! I think this is great addition

@matthewjohns0n can you just have a quick look at this as well (I think it's good to go though)

matthewjohns0n (commented):
This looks good to me, but in some cases we actually want the developer to know that we loaded a template and the tag wasn't set. This could potentially lead to a lot of confusion if someone thinks a layout variable is set and we're no longer showing an error.

@intoeetive What are your thoughts on this? Do you think it would be reasonable to log this in the developer log, or would that just fill it up too fast? Not sure what a good compromise would be, or if we should just approve as-is.

aaronkirkham:
@matthewjohns0n Wonder if instead of using the developer log, we could output the array of variables that were marked as undefined in the ee-debugger section? Not sure if it's too quiet that way, but at least it would let you look at which undefined variables were referenced in the layout.

As an example:

Screenshot 2023-08-29 003732

intoeetive:
@matthewjohns0n since this is specifically addressing layout variables, I think we're fine with treating variables that are not set as empty variables/ Having a debug / profiler record is really the only thing that could be helpful, the `truefalse"}` string output is really not a good indicator that a variable is not set :)

@aaronkirkham I think logging this into template profiler with `log_item` as you suggest would be super helpful!

aaronkirkham:
@intoeetive I've added that in now.

I also made a slight change to the other code to strip any trailing bits of the layout variables e.g. when using `{layout:some_undefined_var index="1"}` now the part after the space will also be dropped when generating the key.

intoeetive:
@aaronkirkham looking into this again, I think the regex is suboptimal.
Instead of checking for space and colon later in the code, we can just modify the regex check like this:
`if (preg_match_all('/' . LD . 'layout:(?!\bcontents\b)([^!]+?)(}|\s|:)/', $str, $matches)) {`
What do you think?

aaronkirkham:
@intoeetive That was my initial though, but we'd still want to match if they contain a `:` or a space initially. If it's the only reference to that variable in the layout, it then wouldn't be assigned an empty string value as the regex doesn't match it.

intoeetive:
@aaronkirkham I think it's still matching correctly - the 2nd capturing group (your `$matches[1]`) remains unchanged. See this test: https://regex101.com/r/pnEFJl/1

aaronkirkham:
@intoeetive Oh you're right! That definitely feels like the better way of doing that - thank you!

I've commit the changes. (replaced the literal `}` character with the RD constant to be consistent)

intoeetive (approved):

PR #3717, opened 2023-08-22
Resolved #2814 where Date picker wasn't always showing calendar when manually entering date into a date field

### LINKED ISSUE #2814: Date picker shows `undefined NaN` when user manually enter date into a date field
EE 7.2.8

I got several date fields on a page, so, sometimes it is much easier to copypaste the selected date from the first field and manually correct them later, without using a date picker.

But when I do this, the date picker shows NaN:
date picker

Here is the video for a better understanding of the issue:
[video-convert-1675324248798.webm](https://user-images.githubusercontent.com/12234507/216263843-35b5cf69-1e63-460c-bb36-6cf5cd8f41be.webm)

Yulyaswan:
Resolved #2814  where Date picker shows undefined NaN when user manually enters date into a date field

EE7 version of #3728

### DIFF
--- a/themes/ee/asset/javascript/src/cp/date_picker.js
+++ b/themes/ee/asset/javascript/src/cp/date_picker.js
@@ -248,14 +248,73 @@ EE.cp.datePicker = {
 				var timestamp = $(this.element).data('timestamp');
 
 				if ( ! timestamp) {
-					d = new Date(Date.parse($(this.element).val()));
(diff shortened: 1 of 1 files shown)


intoeetive (approved):
This works fine for me.

Noting that while this only covers 2 date formatting cases out of 4 supported, for 2 other other ones this is not needed, they work already without the fix

PR #3719, opened 2023-08-23
Feature/7.x/multiple custom menus

### LINKED ISSUE #3270: Custom menu permissions in CP only look at primary role
I made a custom menu and set permissions so only superadmins have access.

I assigned a member to primary role of 'Editor' and then in role assignment, they're assigned to both Editor and Superadmin roles.

If I login as this member in the CP, I don't see my custom menu, even though they have superadmin as a secondary role.

Replicated EE 7.2.14

### LINKED ISSUE #4922: Menu manager menus should be MSM aware
Having added MSM to an existing site which extensively uses the Menu manager on the primary site, when you're working on the MSM site (via the primary site CP) the menu items persist and depending where the link goes links are often "dead" or inaccessible because they point to something on the primary site.

Would it make more sense for menus to be MSM aware, so each MSM site has it's own set of links?

intoeetive:
Make custom menus in CP specific to MSM sites; https://github.com/ExpressionEngine/ExpressionEngine/discussions/3572 

Allow multiple custom menus per role; https://github.com/ExpressionEngine/ExpressionEngine/discussions/3572
Display custom menu heading in navigation; https://github.com/ExpressionEngine/ExpressionEngine/discussions/3646
Use icons for the add-ons in sidebar custom menus; #2859
Deprecate ee()->menu legacy library

closes #3270
closes #4922

This initially seemed like a small change, but turned out into changing how models are related etc, so we can only do it in 7.5

Needs to have tests written


ic360-com:
Wondering if we might want to call the Menu something for reference, but actually display another Heading.
eg. "Client's Menu"... but displayed Heading could be "Quick Edits"

intoeetive:
@Yulyaswan we need a small adjustment here so that the add-on icons in the custom menu in sidebar would have the same alignment and indent
image
Add few add-ons in Settings->Menu Manager to see. The file is `system\ee\ExpressionEngine\View\_shared\sidebar\navigation\custom.php`

PR #3730, opened 2023-08-29
Resolved issue where the user was redirected to different page after saving Front-End Editing Settings

RK311y:
## Overview

The `CP/URL` used for redirection after submitting the **Front-End Editing Settings** is incorrect.

The settings form is visible via the URI: `/cp/settings/pro/frontedit`. 

The redirect URL after submitting the form is currently [`/cp/settings/pro/general`](https://github.com/ExpressionEngine/ExpressionEngine/blob/0949dd1aa989e7ba97321d02c72dfecc0cb83dd8/system/ee/ExpressionEngine/Addons/pro/Controller/Settings/Frontedit.php#L56).

### Suggested changes

Instead of defining the redirect URL, I have proposed code that utilizes that class property [`$base_url`](https://github.com/ExpressionEngine/ExpressionEngine/blob/0949dd1aa989e7ba97321d02c72dfecc0cb83dd8/system/ee/ExpressionEngine/Addons/pro/Controller/Settings/Frontedit.php#L20).

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro/Controller/Settings/Frontedit.php
+++ b/system/ee/ExpressionEngine/Addons/pro/Controller/Settings/Frontedit.php
@@ -53,7 +53,7 @@ class Frontedit extends Settings\Pro
                     ->defer();
             }
 
-            ee()->functions->redirect(ee('CP/URL')->make('settings/pro/general'));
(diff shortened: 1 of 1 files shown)


intoeetive (approved):
@RK311y totally makes sense. Thank you!

PR #3731, opened 2023-08-29
Resolved #3124 where installing Structure did not properly populate records for existing Pages

intoeetive:
Resolved #3124 where installing Structure did not properly populate records for existing Pages

Before this change, the existing pages all were added as Listings. Which I think is not correct.

Now they are added as Pages.

One this that I noticed is that if you have Pages URI set as `some/thing` in Structure it will become just `thing` (the existing routes would work though until that is re-saved). I don't think we can reliably set parent ID here because in Pages you can have parts of URI names as anything, while in Structure both `some` and `thing` need to be URIs of existing entries. Also this seems to be pre-existing issue.

Since this is major change in data structure, we need to make sure to test in on site with very complex usage of Pages and see if nothing gets broken (thinking mostly existing routes, but also the CP)

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/upd.structure.php
+++ b/system/ee/ExpressionEngine/Addons/structure/upd.structure.php
@@ -122,13 +122,16 @@ class Structure_upd
         $this->create_table_structure_members();
         $this->create_table_structure_nav_history();
 
-        // populate listings
-        $this->populate_listings();
-
         // Insert the root node
         $data = array('site_id' => '0', 'entry_id' => '0', 'parent_id' => '0', 'channel_id' => '0', 'listing_cid' => '0', 'lft' => '1', 'rgt' => '2', 'dead' => 'root', 'updated' => date('Y-m-d H:i:s'));
         $sql = ee()->db->insert_string('structure', $data);
         ee()->db->query($sql);
+
+        // convert existing Pages to, um, pages
+        $this->populate_pages();
+
+        // populate listings
+        $this->populate_listings();
 
         // Insert the action id
         $action_id = ee()->cp->fetch_action_id('Structure', 'ajax_move_set_data');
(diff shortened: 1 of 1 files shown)


robinsowell:
Ah- interesting regarding structure that you can't save a / in the uri and they're (the segments) all independent entries.  I tried saving an entry with this for the url - how-does-it-work/idk/dude and it didn't object but it saved it as how-does-it-work-idk-dude.  I didn't realize that was the case- and I agree, we can't take a guess at creating parents.

I wonder if that would be a better way of adjusting/transforming the existing page uris that have slashes?  While I didn't setup up my test to have something/cat/first and something/dog/first, what would happen if you did that?  Would you have two structure uris with 'first' as the value?  If you did the 'replace / with -' in existing page->structure transformation, it would eliminate that possibility and likely be more informative.

So- that's my one suggestion- instead of making an existing page uri with a / just the last segment, transform the / to a -, as structure would do saving it in the entry editor.

I also have a question about the Structure CP- on the validation page, I see no actionable issues, but I do have this:

Screenshot 2023-10-11 at 3 03 51 PM

It's a little confusing- it looks like the two columns should be reversed or I'm not understanding what it's showing me.

Otherwise- it's awesome seeing all of my existing pages in the Structure page in the cp and having the channels properly set to 'Page' type already.

And existing pages that had a / in them do still work, so nice.

The one other thing I find confusing- and this probably relates to that 'URL Mismatches' confusion I had above- I put a structure nav tag on my page, and the urls all used the old pages urls- so the ones with the / in them.  I expected them to use the new urls- the ones that had been converted from pages format to structure format.  So I expected how-does-it-work-idk-dude type instead of how-does-it-work/idk/dude.

It's confusing in part because if I go edit that entry, in the structure uri field before I save, I'd have, well right now I'd have 'dude'.

Meh- hope all of the above makes sense.  But basically, I tend to think existing page uris with / should have those convert to - instead of truncating and I kind of expected a full conversion without saving the entry.  In other words- I wouldn't really expect URL Mismatches and I think I'd expect my structure nav to reflect what a structure url would look like.  

But- I also kind of get why it works that way.  Tricky.

intoeetive:
@robinsowell I made a deeper dive and found that the fact only last segment is shown in the URI input actually has nothing to do with installer. It takes the record directly from `site_pages` field, splits it with `/` and takes the last part. So if we would want to replace `/` with `-` we'd have to modify `site_pages`, which will also make current URLs that have slash in thm to not work anymore. So I would rather not do that.

robinsowell (approved):

PR #3773, opened 2023-09-28
Updated ColorPicker fieldtype so it could be used with Channel Form

intoeetive:


system/ee/ExpressionEngine/Addons/colorpicker/ft.colorpicker.php
```diff
@@ -93,6 +93,10 @@ public function save($data)
      */
     public function display_field($data)
     {
+        if (REQ != 'CP') {
+            return '<input type="color" name="' . $this->field_name . '" value="' . $data . '" />';
```
matthewjohns0n, line 97: If we return early like this, will the Channel Form actually show all the color picker settings set for the field?

intoeetive: It will not. The HTML Color input does not support all those settings. But if we don't return, the whole form will just break

matthewjohns0n: Okay - this seems like a good solution for getting this supported in channel forms, although I'd imagine we eventually need to add support for it to use the field settings. I'm resolving this conversation and approving.

matthewjohns0n (approved):

PR #3775, opened 2023-09-28
Improved styling of Slider fieldtypes when used with Channel Form

intoeetive:


bryannielsen (approved):

intoeetive:
@Yulyaswan we need a fix here
The slider is using the system font (like if the page would have no CSS at all), not the font specified for all other form elements
image

Yulyaswan:
@intoeetive added styles. Please have a look
Screenshot 2023-11-15 at 18 03 19

intoeetive:
Looks good, thanks!

PR #3789, opened 2023-10-02
fix php8.2 warnings found during a visitor testing release

bakin1999:


intoeetive (approved):

intoeetive:
@bakin1999 I've approved th PR, but it looks like Visitor is still calling `password_safety_check` which has been deprecated (but we decided to not log deprecation message just yet). If you're updating it, it's worth to get rid of that method

PR #3805, opened 2023-10-09
Logging improvements

intoeetive:
Added configuration setting to disable logging of CP visits

Resolved issue where member deletion notification were not sent when deleting member via CP

Record member deletions to CP Logs; #3392

Record entry deletions to CP Logs; #3392


intoeetive:
This might be not necessary once we integrate Monolog

PR #3833, opened 2023-10-17
Make channel layout JS more specific and only operate on layout tabs

swierczek:
We utilize the `cp_js_end` hook to add an SSO button to the `#idle-modal` in new tabs:

image

But because of those new tabs in the modal, we're experiencing a bug on the Channel Layout page when adding new Tabs. The JS for this page currently operates on the entire DOM, so when it tries to determine the index for a new tab, it increments it not by one, but by 3 because it sees the 2 tabs that we added to the `#idle-modal` too. See how the new `rel` attribute is `t-9` when the previous tab's `rel` is `t-6` - I'd expect the new `rel` to be `t-7`.

image

The issue that this causes is that when we create a new tab, we are unable to immediately drag fields to that tab, because the relevant section isn't created:

image

---

This PR adds an ID to the Form:

image

And then adjusts the JavaScript to only see the tabs in this form, rather than any tabs in the DOM, so the new tab ID is correct:

image

And the new tab immediately has a section to drag/drop into:

image

### DIFF
--- a/system/ee/ExpressionEngine/View/channels/layout/form.php
+++ b/system/ee/ExpressionEngine/View/channels/layout/form.php
@@ -2,7 +2,7 @@
 
 <div class="panel">
 <div class="form-standard has-tabs publish" data-publish>
-	<?=form_open($form_url, 'class="ajax-validate"')?>
+	<?=form_open($form_url, 'class="ajax-validate" id="layout-form"')?>
   <div class="panel-heading">
     <div class="form-btns form-btns-top">
 		<div class="title-bar title-bar--large">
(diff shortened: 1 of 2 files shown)


intoeetive:
@swierczek any specific reason you removed the code that was there for the debug section?

swierczek:
@intoeetive `// remove debug - it has tabs and we don't want fields to end up in them` - its purpose previously was to ensure that the rest of this JS didn't see its tabs. But that's not needed anymore now that the rest of the JS in this file is only looking at the tabs in the form, not all tabs on the page.

intoeetive:
Got it, thanks!

@Yulyaswan - can you review this PR when you get a chance?

Yulyaswan (approved):

PR #3854, opened 2023-10-23
Include built-in properties and categories into conditional field rules

### LINKED ISSUE #4812: A stray ">" appears when displaying a Field's Conditional Logic
When editing a field, and that field has conditional logic, the conditional logic area displays an floating > in the interface.  See screenshot.

It is not immediately clear where this is coming from, but perhaps it's in the react code?

**How To Reproduce**
Steps to reproduce the behaviour:

1. Create a field.  Make it conditional.
2. Make the condition be that another field, Contains, some text.
3. Save, then reload or edit that field: admin.php?/cp/fields/edit/350
4. View that area of the page and lo, there is an extra > visible in the interface.

Image

**Environment Details:**
 - Version: 7.5.10

### LINKED ISSUE #5119: Conditional logic can lead to the appearance of data loss...
Issue- had a case where a conditional was set up, ran sync, and a it appeared there was data loss.  

Simplified example:

I did a really simple test, have a field SEO Description.  It’s on a LOT of channels.  I made it conditional on having content in a text field called ‘my_textarea’.  That field?  Is only in a few channels.  After I sync the conditiona, the SEO Description disappears from the cp on all channels that don’t have the ‘my_textarea’ field- because… there is no content in my_textarea.  This happens on the frontend as well, because according to conditional logic, there is no content in 'my_textarea'.

But it's unexpected because the conditional is checking a field that doesn't even EXIST in that channel.  I was basically expecting it to ignore it.

I do think we need to at least warning people before they run a sync that the field they are making conditional exist in some channels where the conditional will always return 'false'.

I don't think we should just change the logic to ignore the conditional check when a field that something is conditional on doesn't exist, because that alters existing behavior.

I think my favorite solution would be an option in the conditionals to say 'ignore the conditional check if one or more fields used in the conditional doesn't exist.  And then it's up to the user and doesn't break backwards compatibility. Or- give them a list of channels to apply the conditional in, and it only works in those channel.  Or some such.

What's happening is logical, but it can also be very unexpected, especially if you're keeping things DRY and reusing fields.

intoeetive:
Include entry title and url_title into conditional field rules; #2929
Include entry status into conditional field rules; #2929
Include categories into conditional field rules; #2929
Include channel into conditional field rules; fixes #5119

Fixes #4812


PDXMan1:
This is exactly what need! I'll be keeping a watch for the next build and hope this makes it. Thank you for your work on this!

matthewjohns0n (dismissed):

PR #3855, opened 2023-10-23
Include short field name in keyword search in channel settings

Yulyaswan:
Include short field name in keyword search in channel settings; #3822


intoeetive (commented):
In most cases, `item.instructions` does not exist or is empty, should we check for that before checking against searchTerm? Also, should we lowercase searchTerm before doing the search, instead of doing that for each loop in the cycle?

Other than those performance notices, this has been tested and and it works

intoeetive (approved):

PR #3861, opened 2023-10-24
Resolved #3860 where certain "Save" buttons did not scale properly on mobile devices

### LINKED ISSUE #3860: UI issue on moblie on Entry page
So... this one is a bit of a bear to reproduce.  I've only seen it happen with Publisher add-on in use:

EE_Admin

It's jamming the user, as "when trying to edit EE entries on a small screen, the “View (Published)” and “Save As (Draft)” buttons are really unusable. They are both inside a single container that does not expand. "

The reached out to Publisher/BM about it, but per them, it's actually an EE Core issue, as Publisher is using all the built in styles and markup for these elements.

But... you pretty much have to install Publisher to actually see the issue, at least from me experimenting around on my tiny phone.

lnicely:
Resolves https://github.com/ExpressionEngine/ExpressionEngine/issues/3860

### DIFF
--- a/cp-styles/app/styles/components/_filter-bar.scss
+++ b/cp-styles/app/styles/components/_filter-bar.scss
@@ -380,3 +380,14 @@
 	  }
 	}
 }
+
+@media only screen and (max-width: $tablet-width) {
+	.filter-bar__item.publisher-save-as {
+          position: static;
+        }
+
+        .filter-bar__item.publisher-save-as > button {
+		  margin-top: 2em;
+          position: relative;
+        }
+  }
--- a/themes/ee/cp/css/common.min.css
+++ b/themes/ee/cp/css/common.min.css
@@ -28515,6 +28515,16 @@ input[type=button].btn-block {
   display: block;
 }
 
+@media only screen and (max-width: 768px) {
+  .filter-bar__item.publisher-save-as {
+    position: static;
+  }
+
+  .filter-bar__item.publisher-save-as > button {
+    margin-top: 2em;
+    position: relative;
+  }
+}
 .tab-bar {
   display: flex;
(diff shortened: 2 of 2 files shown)


intoeetive (changes requested):
@lnicely thank you for the contribution!

However in the form that you submitted it would not persist, because the `common.min.css` is gettting built automatically out of `scss` source files in `cp-styles/app/styles` folder. So the change needs to be done in appropriate file there and then the build process needs to be run with `npm run build:css` command.

We don't have this documented well unfortunately. If you're unsure what exactly needs to be done, just let us know and we'll take care of that. Thanks!

Yulyaswan (changes requested):
I would suggest these styles. in this case we will not have to change the appearance of this block
```
@media only screen and (max-width: $tablet-width) {
	.filter-bar__item.publisher-save-as {
          position: static;
        }

        .filter-bar__item.publisher-save-as > button {
          position: relative;
        }
  }
```

https://github.com/ExpressionEngine/ExpressionEngine/assets/23382425/83199e1f-44dd-4b0f-ab84-df52304f5195

lnicely:
Thanks @Yulyaswan! That'll work. In the case where a Publisher draft exists, an additional 'View As' button appears. If we add `margin-top: 2em;` to `.filter-bar__item.publisher-save-as > button`, that takes care of the issue where the 'Draft' button isn't fully shown.

Yulyaswan (approved):

intoeetive:
@lnicely since this has been already marked as approved, there's no need to re-request review.

We have scheduled it for one of upcoming 7.x patch releases, although we can't give any ETA at the moment

Thank you!

intoeetive (approved):

PR #3862, opened 2023-10-25
Resolved #3841 where some characters in entry title might cause automatic url_title generation to not work

### LINKED ISSUE #3841: An entry titled "Selçuk Ören" cannot be created when `Enforce automatic URL title` is enabled
EE 7.3.12

How to reproduce:

1. Enable `Enforce automatic URL title` option in channel settings
2. Try to create a new entry, with title `Selçuk Ören`.

Video:
[video-convert-1697647638117.webm](https://github.com/ExpressionEngine/ExpressionEngine/assets/12234507/48a76c6f-3580-4f91-948c-09afe628c492)

### LINKED ISSUE #3870: `url_title` changes after saving an entry when `Enforce automatic URL title` is enabled
How to reproduce:

1. Enable `Enforce automatic URL title` in channel settings
2. Create an entry titled "Anča". 
3. You'll see that `url_title` value is "anca" before saving and "anaa" after saving.

And also, not sure if I need to post this separately, but Pro Search can't find entry "Anča" using "anca" keyword.

[Video](https://github.com/ExpressionEngine/ExpressionEngine/assets/12234507/49e9bd62-5663-4bbd-9747-a5a6ee65476a)

intoeetive:
Resolved #3841 where some characters in entry title might cause automatic url_title generation to not work
Resolved #3870 where converting text to ASCII might have not always worked properly

When adding PHP8.2 compatibility, we replaced `utf8_decode` with `mb_convert_encoding` here, which caused mapping to foreign character not work properly.

This fix is making direct use of `mb_ord` if it's available, and falling back to deprecated `mb_convert_encoding` if multibyte is  not available (this will throw deprecation error, but people should really have multibyte support if they run PHP 8.2 in 2023)

### DIFF
--- a/system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
+++ b/system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
@@ -501,6 +501,9 @@ And if you made it to this &#x1F573;&#xFE0F; you did pretty good.']
                 ],
                 'Sample-Title-to-Turn-Into-a-Slug-including-💩-tags-quotes-and-high-ascii-ssae-and-seps____in....content'
             ],
+            // ['ExpressionEngine®', [], 'expressionengine'],
+            ['Anča', [], 'anca'],
+            ['The General’s Room', [], 'the-generals-room'],
         ];
     }
 
--- a/system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
+++ b/system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
@@ -462,11 +462,7 @@ And if you made it to this &#x1F573;&#xFE0F; you did pretty good.']
      */
     public function testUrlSlug($content, $params, $expected)
     {
-        // minimal map
-        $config['foreign_chars'] = [
-            '223' => "ss", // ß
-            '230' => "ae", // æ
-        ];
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
```diff
@@ -501,6 +501,9 @@ public function urlSlugProvider()
                 ],
                 'Sample-Title-to-Turn-Into-a-Slug-including-💩-tags-quotes-and-high-ascii-ssae-and-seps____in....content'
             ],
+            // ['ExpressionEngine®', [], 'expressionengine'],
```
bryannielsen: @intoeetive I added these tests to cover the issues described but this one won't pass, the formatter leaves the (R) symbol.  But when I test within the control panel it does get removed so I don't know how important it is to behave the same way and I'm not sure where the control panel behavior is defined (it seems like JS).

bryannielsen (approved):

system/ee/ExpressionEngine/Tests/ExpressionEngine/Service/Formatter/TextFormatterTest.php
intoeetive: @bryannielsen [®](https://www.htmlsymbols.xyz/unicode/U+00AE) is in our Emoji map, and we allow emojies in URL Slug. So I updated your unit tests script

bryannielsen (approved):

PR #3880, opened 2023-11-08
Resolved #3847 where it was not possible to identify the the publish tab with validation error if the tab was hidden

### LINKED ISSUE #3847: Entry save fails with error box but no highlighted issue
Client running EE 6.4.13 - long established site - reports new error.
When editing entry that has file fields within grids grid fields. One field is populated successfully and saved. When a  file is attached to a second grid within the same entry, the file can be uploaded OK (confirm it arrives in Files area etc.) but attempts to save the entry fail with red box error "We were unable to update this entry, please review and Ix errors below.". Except there are no errors highlighted below.

The issue is quite possibly a configuration issue, the reason for posting is that the error message report is unhelpful - in that the location of the error is not highlighted.

**How To Reproduce**
Error is repeatable within site both on live server and staging server, but seems isolated to this site.

**Error Messages**

Edit Entry

**Environment Details:**
 - Version: 6.4.13
 - PHP Version 8.0.3

**Possible Solution**
Fix the error reporting to highlight the problem field.

intoeetive:
Resolved #3847 where it was not possible to identify the the publish tab with validation error if the tab was hidden

### DIFF
--- a/system/ee/ExpressionEngine/View/publish/partials/publish_form.php
+++ b/system/ee/ExpressionEngine/View/publish/partials/publish_form.php
@@ -29,7 +29,7 @@
             <div class="tab-bar__tabs">
             <?php
             foreach ($layout->getTabs() as $index => $tab):
-                if (! $tab->isVisible()) {
+                if (! $tab->isVisible() && ! $tab->hasErrors($errors)) {
                     continue;
                 }
                 $class = '';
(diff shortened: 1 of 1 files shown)


Yulyaswan (dismissed):
I can successfully save the page even if there is an empty required field in the hidden tab. The field is not conditional.


Uploading Screen Recording 2024-02-19 at 12.48.05.mov…

intoeetive:
@Yulyaswan the video was not uploaded - but I also wasn't able to save, got the expected error messages. The steps to replicate the original problem are in last comment in referenced issue

intoeetive:
Upon testing in 7.4, we were not able to replicate the exact reported issue.

If the required field is on a hidden tab, the validation error shows up when creating new entry, but not when editing existing entry. When editing existing entry and the required field is on hidden tab, the entry just saved.

At this point, I'm not sure that what is the desired behaviour and whether we want to change anything. Asking @matthewjohns0n @bryannielsen @robinsowell to weigh on this

PR #3881, opened 2023-11-08
Added possibility to change field label using Publish Layouts

Yulyaswan:
Video: 
https://github.com/ExpressionEngine/ExpressionEngine/assets/752126/0465ea15-2029-418d-836b-2e7adbd11640


intoeetive:
@Yulyaswan I added the backend part, but I think we need a UX update.

1. It is currently not clear that the field name can be edited. I suggest maybe showing pencil icon on hover? (like we do for copying field short name on entry page). Or maybe change mouse to text cursor, not hand.
2. If you press enter in the input, the page gets submitted, but the changes are not saved. We need to prevent that, and make pressing Enter work same as 'blur' currently (and not submit page)

Yulyaswan:
@intoeetive I changed as you asked. Please check

intoeetive (dismissed):
@Yulyaswan loving it! Thank you

PR #3893, opened 2023-11-11
adds `response_send_output` hook

### LINKED ISSUE #3892: Add `response_send_output` hook to Response object
To allow for tooling like the EE Debug Toolbar, we only need a single hook made available. A discussion on implications should take place though

eric-lamb62:
Just adds a simple hook to start the discussion

closes #3892

### DIFF
--- a/system/ee/ExpressionEngine/Core/Response.php
+++ b/system/ee/ExpressionEngine/Core/Response.php
@@ -93,10 +93,6 @@ class Response
      */
     public function send()
     {
-        if (ee()->extensions->active_hook('response_send_output') === true) {
-            $this->body = ee()->extensions->call('response_send_output', $this->body);
-        }
-
         if (! $this->body) {
             foreach ($this->headers as $name => $value) {
                 $GLOBALS['OUT']->headers[] = array($name . ': ' . $value, true);
--- a/system/ee/ExpressionEngine/Core/Response.php
+++ b/system/ee/ExpressionEngine/Core/Response.php
@@ -93,6 +93,10 @@ class Response
      */
     public function send()
     {
+        if (ee()->extensions->active_hook('response_send_output') === true) {
+            $this->body = ee()->extensions->call('response_send_output', $this->body);
+        }
+
         if (! $this->body) {
             foreach ($this->headers as $name => $value) {
                 $GLOBALS['OUT']->headers[] = array($name . ': ' . $value, true);
--- a/system/ee/legacy/core/Output.php
+++ b/system/ee/legacy/core/Output.php
@@ -341,6 +341,13 @@ class EE_Output
         }
         // --------------------------------------------------------------------
 
+        if (ee()->extensions->active_hook('before_response_send_output') === true) {
+            $output = ee()->extensions->call('before_response_send_output', $output);
+            if (ee()->extensions->end_script === true) {
+                return;
+            }
+        }
+
         // Do we need to generate profile data?
         // If so, load the Profile service and run it.
         if ($this->enable_profiler == true && (! (AJAX_REQUEST or ee('LivePreview')->hasEntryData()))) {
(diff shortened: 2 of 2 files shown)


intoeetive (commented):
@mithra62 thank you for contributing!

A couple of thoughts:
- would it better to name the hook `before_response_send_output` ?
- would it be worth adding ability to also modify the headers (although I'm not sure if that might involve potential security breach, @matthewjohns0n to weigh on this) and ability to terminate the output using `ee()->extensions->end_script` like in other placed?

eric-lamb62:
@intoeetive 

Oh yeah, I can see renaming the hook though the inverse, `after_response_send_output`, opens some questions. Also agree on the headers on both counts of useful and possibly a security implication. 

Biggest thing though is the `Response` object may not be the best place for this hook. From my testing (and I could be doing it wrong), it appears every request goes through the `EE_Output` object (templates and CP) which is called within the `Response` object. In fact, I can find no instances where a request doesn't end at https://github.com/ExpressionEngine/ExpressionEngine/blob/a92f5691cced14aecd62b6bce8b62d9e40023d43/system/ee/ExpressionEngine/Core/Response.php#L106

Further, considering the `$GLOBALS['OUT']->_display()` calls `echo` within the method, and has no return, we could probably refactor to move the `return` after the `after_response_send_output` call. Or move the hooks into the `EE_Output` object calls. 

Definitely some considerations to be made...

intoeetive:
`EE_Output` might be better place indeed. This is where _elapsed time and memory usage_ variables are parsed, so adding a hook that would be used for extra debugging there kinda makes sense. Also this is where EE native Profiler is being added

eric-lamb62:
@intoeetive I made a change to move the hook(s) to the `EE_Output` object in the `_response()` method. I also added the `end_script` and broke it up into both `before` and `after` variants. 

https://github.com/ExpressionEngine/ExpressionEngine/blob/749cd7fbc76ce8c0764ca2d9be273cbc92896668/system/ee/legacy/core/Output.php#L344

https://github.com/ExpressionEngine/ExpressionEngine/blob/749cd7fbc76ce8c0764ca2d9be273cbc92896668/system/ee/legacy/core/Output.php#L410

Added bonus, we don't have to worry about passing headers with this so that's nice 😸

intoeetive:
I think this is good - although asking @TomJaeger to say final word on this

eric-lamb62:
I definitely have to run some more tests on my end; not entirely sure of all the implications with this change (I have a specific goal in mind for this addition). It's almost 1:30am here though so will have to get into later today after a quick rest.

eric-lamb62:
Insomnia's a jerk... just ran my tests; everything came back green on my end.

matthewjohns0n:
@mithra62 I can imagine this to be potentially useful. I'm just wondering, what was your intended use case for this? 

As far as security goes, extensions already have a lot of power. There are many hooks that a malicious third party could use with great effect. I don't think this is any "worse" than some of the existing hooks, and I would say that in general, add-on code already gets read and scrutinized once it's out there. It is equally important to install trusted add-ons, since they are given a lot of power.

Maybe a better solution is for the add-on store to scan add-ons upon submission for certain things like what hooks they are using and publish that on the add-on page. That would allow developers to be a little more critical about what add-ons they install on their sites.

eric-lamb62:
Thanks @matthewjohns0n! It's specifically for this 
https://github.com/mithra62/ee_debug_toolbar

matthewjohns0n:
Ahhh gotcha - I'm downloading the toolbar now to play around with, I've been meaning to for a while. I actually assumed that it was for that, but I didn't realize you had to manually add that hook.

eric-lamb62:
Oh! Hell yes. Lemme know what you think :)

TomJaeger:
All good here as well.

intoeetive (approved):

PR #3897, opened 2023-11-13
Update tables to fix missing primary keys

### LINKED ISSUE #3891: All EE tables need a primary key or fail when sql_require_primary_key set.
Note,  MySQL Galera clusters using row-based replication requires that all tables have a primary key that is unique.  So those folks really don't have an option there.

Looks like we have the following tables without:

exp_consent_request_version_cookies
exp_dashboard_layout_widgets
exp_dock_prolets
exp_file_usage.

See also https://github.com/ExpressionEngine/ExpressionEngine/issues/3535

Won't allow create/alter of affected tables.

bryannielsen:
Add missing primary keys, resolves #3891

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro/upd.pro.php
+++ b/system/ee/ExpressionEngine/Addons/pro/upd.pro.php
@@ -184,6 +184,13 @@ class Pro_upd extends Installer
             if (!ee()->db->table_exists('dock_prolets')) {
                 ee()->dbforge->add_field(
                     [
+                        'dock_prolets_id' => [
+                            'type' => 'int',
+                            'constraint' => 10,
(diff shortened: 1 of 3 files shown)


intoeetive (changes requested):
@bryannielsen I think we also need to have keys added in main mysql_schema file for new installation

Also, are these the only tables that need this? I feel like there should be more (including new channel_category_groups probably)

bryannielsen:
@intoeetive I updated the mysql_schema file and pro install to include these new keys.  I also looked through all the tables and after this update runs every table has a primary key.  `channel_category_groups` already has a primary key (channel_id, group_id)

intoeetive (approved):

PR #3900, opened 2023-11-14
Support color picker swatches in channel forms

bryannielsen:


intoeetive (approved):

intoeetive:
Test run cancelled because I just merged 7.dev back into branch and previous run was successull (and this part is not covered with tests anyway)

PR #3909, opened 2023-11-23
Enabled pressing `Ctrl + A` to add new elements in the CP; Renamed "New..." to "Add..." in the Control Panel UI

intoeetive:
Enabled pressing `Ctrl + A` to add new elements in the CP;
Renamed "New..." to "Add..." in the Control Panel UI

Ctrl+N is reserved combination in Chromium-based browsers, so I had to rename "New member" to "Add member", "New Category" to "Add category" etc

Because of this UX change I think it needs to be included into minor release, not patch


matthewjohns0n (commented):
Unfortunately, I personally use CMD+A when editing content, since it is "select all", so I would prefer we not take over that shortcut. Would you be open to branstorming a different key combination?

intoeetive:
We can consider another shortcut, but I believe within fields you'll have it still working as previously, and it's not too likely someone will want to select all text on CP page

PR #3917, opened 2023-11-29
Improved styles for Entry Status dropdown on Publish pages

Yulyaswan:
Improved Entry status styles; Fixes #2856 

This is implementing styles that were originally planned for 6.0 in https://packettide.atlassian.net/browse/EECORE-557

Screenshot 2023-11-29 at 22 05 32


intoeetive (approved):
worked well on my tests

PR #3922, opened 2023-12-01
Resolved #987 where pagination could be wrong on Entry Manager if Categories column present

### LINKED ISSUE #987: Wrong limit on Entry Manager page when Categories column present
I have a test site running ee 6.0.3, imported data using Zealous Web's Import / Export addon, had a few problems and they are now fixed, however before I reinstalled the updated version I deleted the 31 entries from the first import, cleared all caches and then installed the new version of the app and imported the 31 entries again which had been altered slightly to include statuses and URL_titles.

Imported all 31 perfectly apart from the fact that the channel entries page (cp/publish/edit) only shows 13 entries when it is set to show 25, the drop down set to show 50 only shows 25, set to 75 shows them all, the weird bit is the last option says show all 62 entries, there are only 31 ?

intoeetive:
Resolved #987 where pagination could be wrong on Entry Manager if Categories column present

### DIFF
--- a/system/ee/ExpressionEngine/Library/CP/EntryManager/Columns/Categories.php
+++ b/system/ee/ExpressionEngine/Library/CP/EntryManager/Columns/Categories.php
@@ -19,16 +19,6 @@ use Mexitek\PHPColors\Color;
  */
 class Categories extends Column
 {
-    public function getEntryManagerColumnModels()
-    {
-        return ['Categories'];
-    }
-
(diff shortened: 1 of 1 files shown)


robinsowell:
Tested out by slack user and it fixed it with no issues!  (I never did manage to replicate it- I don't know why not.)

robinsowell (approved):

Issue #1643, opened 2021-12-10
Issues with the new Fluid field hooks


litzinger:
Sorry it's taken so long to get around to actually using the new hooks as they were implemented, and thanks for adding them, but their current implementation falls a little short. My original submission included passing the FluidField instance, if available at the time of the hook call, as the first parameter of the hook call. Having the full instance available provides a lot of information, such as the fluid_field_id, and entry_id being enacted upon. It looks like the `fluid_field_add_field` and `fluid_field_get_all_data` hook signatures were changed.

Originally submitted as

`fluid_field_get_all_data($entryId, $fluidFieldId)`
`fluid_field_add_field(FluidField $fluidField, $fieldTableName, $fieldData)`

but they're currently

`fluid_field_get_all_data(FieldData $fieldData)`
`fluid_field_add_field($fieldTableName, $fieldData)`

When I call the `fluid_field_add_field` from what I can tell when setting a break point, is that with the new method signature I have no way of knowing what the entry_id or the fluid_field_id is.

Originally submitted the signatures as:

```
public function fluid_field_get_all_data($entryId, $fluidFieldId): Collection|array
public function fluid_field_get_field_data($fieldId, $fieldDataId): array
public function fluid_field_add_field(FluidField $fluidField, $fieldTableName, $values): array
public function fluid_field_update_field(FluidField $fluidField, $fieldTableName, $values): array
public function fluid_field_remove_field(FluidField $fluidField): void
```

Ideally they'd all have `FluidField $fluidField` as the first parameter. It's important to know the `entry_id`, `field_id`, `fluid_field_id`, and the state of any array data of the field when the hook is called.

For example with the current implementation of `fluid_field_add_field` I have no idea what the `entry_id` or the `fluid_field_id` is :(

<img width="977" alt="Screen Shot 2021-12-10 at 2 34 54 PM" src="https://user-images.githubusercontent.com/53965/145640647-2518e014-0df9-4014-9603-7236899241f0.png">

intoeetive:
@dougblackjr it looks like you worked on this originally, do you have any insight?

PR #3927, opened 2023-12-05
Added fluid field information to extension hook; #1643

intoeetive:
Added fluid field information to extension hook; #1643

### DIFF
--- a/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
@@ -398,6 +398,15 @@ class Fluid_field_ft extends EE_Fieldtype
         $query->set($values);
         $query->where('id', $fluid_field->field_data_id);
         $query->update($fluid_field->ChannelField->getTableName());
+
+        if (ee()->extensions->active_hook('fluid_field_after_update_field') === true) {
+            ee()->extensions->call(
+                'fluid_field_after_update_field',
+                $fluid_field,
+                $fluid_field->ChannelField->getTableName(),
+                $values
+            );
+        }
     }
 
     private function addField($order, $group, $field_id, array $values)
@@ -436,6 +445,16 @@ class Fluid_field_ft extends EE_Fieldtype
 
         $fluid_field->field_data_id = $id;
         $fluid_field->save();
+
+        if (ee()->extensions->active_hook('fluid_field_after_add_field') === true) {
+            ee()->extensions->call(
+                'fluid_field_after_add_field',
+                $fluid_field,
+                $fluid_field->ChannelField->getTableName(),
+                $values,
+                $id
+            );
+        }
     }
 
     private function removeField($fluid_field)
--- a/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
@@ -422,7 +422,8 @@ class Fluid_field_ft extends EE_Fieldtype
             $values = ee()->extensions->call(
                 'fluid_field_add_field',
                 $fluid_field->ChannelField->getTableName(),
-                $values
+                $values,
+                $fluid_field
             );
         }
 
--- a/system/ee/ExpressionEngine/Addons/fluid_field/Model/FluidField.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/Model/FluidField.php
@@ -119,7 +119,8 @@ class FluidField extends Model
             $rows = ee()->extensions->call(
                 'fluid_field_get_field_data',
                 $this->field_id,
-                $this->field_data_id
+                $this->field_data_id,
+                $this
             );
         } else {
             ee()->db->where('id', $this->field_data_id);
@@ -143,7 +144,8 @@ class FluidField extends Model
             $field_data = ee()->extensions->call(
                 'fluid_field_get_all_data',
                 $field_data,
-                $this->fluid_field_id
+                $this->fluid_field_id,
+                $this
             );
         }
 


litzinger:
I see some updates are being made to the Fluid stuff. I have some other recommendations: https://gist.github.com/litzinger/5f7f8f4abba565ff976b22868f95db99

Most of those are additions, which could wait, but this one in particular feels like should be made sooner than later. The add and update hook signatures are different. This changes makes them consistent:
https://gist.github.com/litzinger/5f7f8f4abba565ff976b22868f95db99?permalink_comment_id=4632953#gistcomment-4632953

intoeetive:
@litzinger thank you for the suggested changes!

I don't think we could change the `fluid_field_add_field` signature - as if someone's using it, it will break their code. So how much I regret it, we'll likely have to stick with different parameters order =(

I'll be happy to add `after_update` and `after_add` hooks. 

I don't think we need to separately pass `$this->field_data_id, $this->fluid_field_id, $this->entry_id` in `fluid_field_add_field` because `$fluid_field` will be the model instance that already has those data. Please let me know if I miss something there.

litzinger:
I don't think you missed anything. Understandable about the hook parameter order, just a little awkward ¯\_(ツ)_/¯ 

I have a bunch of notes in that gist to update my branch once 7.4 is released. I'll package everything in a single PR. Something I added to that gist as a reminder to myself is that one of the hooks appears to be in the wrong location and should probably be moved.

https://github.com/litzinger/ExpressionEngine/commit/614fdfb7779ddeb2c9708ca0cef4b069dd8ac138

litzinger:
> I don't think we need to separately pass `$this->field_data_id, $this->fluid_field_id, $this->entry_id` in `fluid_field_add_field` because `$fluid_field` will be the model instance that already has those data. Please let me know if I miss something there.

We either need to pass those individually, or pass `$this` so it has access to the entire object. The only parameter it gets now is `$this->field_id`, so either we pass those separately, or we change it to something like this:

```
->call('fluid_field_get_field_data', $this->field_id, $this)
```

But that is inconsistent with how all the other Fluid hooks work... none of the others pass `$this`. So it needs either the whole object passed, or just the relevant fields... in this case the data_id, field_id, and entry_id.

intoeetive:
I was talking about `fluid_field_add_field` which is in `ft.` file.

For `fluid_field_get_field_data` we probably need to have those extra data (or whole model reference, which is likely better) added

intoeetive:
@litzinger I have made updated to the code and to the docs (see referenced PR).

Would you mind checking and seeing if that's any good?

system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
```diff
@@ -436,6 +445,16 @@ private function addField($order, $group, $field_id, array $values)
...
         $fluid_field->field_data_id = $id;
         $fluid_field->save();
+
+        if (ee()->extensions->active_hook('fluid_field_after_add_field') === true) {
+            ee()->extensions->call(
+                'fluid_field_after_add_field',
+                $fluid_field,
+                $fluid_field->ChannelField->getTableName(),
+                $values,
+                $id
```
litzinger, line 455: This looks good. So far I have basic text fields working fine in Fluid and Publisher  with most of these changes, but one more small change might be necessary. I wrapped the `fluid_field-save()` call on line 447 with a conditional, so the first hook could optionally return an empty array, and if so then there isn't anything to save. I did this so I could save non-default language, otherwise they end up getting added to the native Fluid tables, which only hold the default language. For example:

```
if (ee()->extensions->active_hook('fluid_field_add_field') === true) {
            $values = ee()->extensions->call(
                'fluid_field_add_field',
                $fluid_field->ChannelField->getTableName(),
                $values,
                $fluid_field
            );
        }

        $field = ee('Model')->get('ChannelField', $field_id)->first();

        if (!empty($values)) {
            $query = ee('db');
            $query->set($values);
            $query->insert($field->getTableName());
            $id = $query->insert_id();

            $fluid_field->field_data_id = $id;
            $fluid_field->save();
        }

        if (ee()->extensions->active_hook('fluid_field_after_add_field') === true) {
            ee()->extensions->call(
                'fluid_field_after_add_field',
                $fluid_field,
                $fluid_field->ChannelField->getTableName(),
                $values,
                $id ?? 0
            );
        }
```

The conditional and the null check to default $id to 0 are the differences. I did something similar to the updateField method too:

```
if (ee()->extensions->active_hook('fluid_field_update_field') === true) {
            $values = ee()->extensions->call(
                'fluid_field_update_field',
                $fluid_field,
                $fluid_field->ChannelField->getTableName(),
                $values
            );
        }

        $fluid_field->field_group_id = array_key_exists('id', $group) ? $group['id'] : null;
        $fluid_field->group = array_key_exists('order', $group) ? $group['order'] : null;
        $fluid_field->order = $order;
        $fluid_field->save();

        if (!empty($values)) {
            $query = ee('db');
            $query->set($values);
            $query->where('id', $fluid_field->field_data_id);
            $query->update($fluid_field->ChannelField->getTableName());
        }

        if (ee()->extensions->active_hook('fluid_field_after_update_field') === true) {
            ee()->extensions->call(
                'fluid_field_after_update_field',
                $fluid_field,
                $fluid_field->ChannelField->getTableName(),
                $values
            );
        }
```

Even without the hook changes the `if (!empty()` check on those arrays could be a nice safety mechanism.

litzinger:
This looks good. There were a couple of other things that came up that we could take a look at now or I can submit a PR at a later date?

The first is this change to the display_field method. https://gist.github.com/litzinger/5f7f8f4abba565ff976b22868f95db99#file-ft-fluid_field-2-php-diff - This change seems more logical to me as it wants to output a collection of fields, but there isn't a loop? I tested this change with and without Publisher in place and it seems to work fine for both. When I was testing with Publisher, and I forget the details b/c it's been a few months, not having the loop there meant I couldn't use the hooks to modify the data and render a field properly.

The other thing... Did you take a look at the location of the `fluid_field_get_all_data` hook? It is located in the `FluidField->getFieldData()` method, but that method gets the contents of a single field value, not all the fields in a Fluid field. I think the hook name is misleading... with a name like that I'd expect it to be in `Fluid_field_ft->getFieldData()` where it fetches everything to render the entire field output. `fluid_field_get_field_data` is already called in in `FluidField->getFieldData()` around line 141

`$field_data = $this->setFieldData($this->fetchFieldData());`

So having the `fluid_field_get_all_data` data called just after that is kind of weird. The issue I had was `fluid_field_get_field_data` returned one data structure, and `fluid_field_get_all_data` returned a different one. I feel like putting `fluid_field_get_all_data` in it's current location was an oversight.

https://github.com/litzinger/ExpressionEngine/commit/614fdfb7779ddeb2c9708ca0cef4b069dd8ac138

I think it makes more sense to have it in `ft.fluid_field.php`:

```
private function getFieldData($fluid_field_id = '', $entry_id = '')
    {
        $fluid_field_id = ($fluid_field_id) ?: $this->field_id;
        $entry_id = ($entry_id) ?: $this->content_id;

        $cache_key = "FluidField/{$fluid_field_id}/{$entry_id}";

        if (($fluid_field_data = ee()->session->cache("FluidField", $cache_key, false)) === false) {
            $fluid_field_data = ee('Model')->get('fluid_field:FluidField')
                ->with('ChannelField')
                ->filter('fluid_field_id', $fluid_field_id)
                ->filter('entry_id', $entry_id)
                ->order('group')
                ->order('order')
                ->all();

            ee()->session->set_cache("FluidField", $cache_key, $fluid_field_data);
        }

        if (ee()->extensions->active_hook('fluid_field_get_all_data') === true) {
            $fluid_field_data = ee()->extensions->call(
                'fluid_field_get_all_data',
                $fluid_field_data,
                $fluid_field_id,
                $entry_id
            );
        }

        return $fluid_field_data;
    }
```

intoeetive:
@litzinger TBH, not all the stuff with the extension hooks in Fluid makes sense to me... I'm just assuming that the person who added those pursued specific goal and they knew what they were doing :)

I'd say why don't you open a PR against this `feature/7.x/fluid-field-hooks-update` branch and then we could just merge everything into a single PR? This might be faster and more efficient than me walking through your comments

litzinger:
I can do that. What timeline are you looking at for having this done?

intoeetive:
Oh, no timeline. We are likely not making this change in 7.4 (unless you insist:)

litzinger:
Sounds good, I'm not in a rush to get this sorted out. I still have a lot of integration to work on and things to figure out with Grid and Relationship fields inside of Fluid.

litzinger:
Created new PR for this work https://github.com/ExpressionEngine/ExpressionEngine/pull/4125

litzinger:
I don't see how to close this PR...

PR #3931, opened 2023-12-06
Fixed gmmktime bug in mod_channel_calendar.php where strings were being passed instead of ints

studiopatagon:
gmmktime expected ints and strings were being passed causing php errors


intoeetive:
@willowlight thank you for the submission - however I think `intval` needs to be all [lowercase](https://www.php.net/manual/en/function.intval.php)

Might be also better to just typecast using `(int) $variable`?

studiopatagon:
Typecasting makes sense to me. Do you want to take care of it or should I submit a new PR? Thanks!

intoeetive:
Maybe you could just push a commit to this same PR? Thanks!

studiopatagon:
@intoeetive done! Thanks.

intoeetive (dismissed):
Looks good, thanks!

bryannielsen (approved):

PR #3934, opened 2023-12-06
Allow automatic login after account activation; Allow redirect to custom page after account activation; #3581

intoeetive:
Allow automatic login after account activation; #3581
Allow redirect to custom page after account activation; #3581

Docs PR: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/835


bryannielsen:
Just noting that we also added a configuration option for disabling automatic login when the activation method is set to 'none'.

bryannielsen (dismissed):

intoeetive:
I have pushed an update to retain existing behaviour on upgraded sites. We set `registration_auto_login` to `y` on new installs, but we don't update or set it on upgrades, so the member will not be logged in.

We also need to update the documentation here

bryannielsen (approved):

PR #3935, opened 2023-12-06
Resolved issue where saving entry with Fluid field using groups could throw PHP error if field was removed from group

intoeetive:
The original issue:
- have a fluid using field group "news" with 2 fields
- create and entry with Fluid field, add "news" block
- remove second field from the group
- try to save entry
- get the error


robinsowell:
Replicated the bug- then applied fix- now now bug!  Tried it with an already saved one and a new one after patching, all good from here.  Note- tested on 74dp.

robinsowell (dismissed):

bryannielsen (approved):

PR #3944, opened 2023-12-11
Added Debug Channel Entries Utility; #3024

intoeetive:
Added Debug Channel Entries Utility; #3024


intoeetive:
Just had someone on Slack run into issue where this utility would be helpful, so I think we should aim to release it better sooner than later

matthewjohns0n (dismissed):
The code looks good on this. I did pull it down to test, but my entries were fine, so I just saw the success messages and didn't get to test the case where there are issues. 


I think this should go in the "Overview" section, like all the other debug tools:

Screenshot 2024-03-18 at 6 27 40 PM

matthewjohns0n (approved):
Looks good to me!

PR #3948, opened 2023-12-13
Added searchable Structure URI column to Entry Manager

intoeetive:
Added searchable Structure URI column to Entry Manager

Updated the Structure settings UI


matthewjohns0n:
@intoeetive Looks like this branch errors out on the setting page. Looks like some sort of namespace conflict. Any ideas there? 

Screenshot 2024-02-02 at 12 36 37 PM


(This is resolved now)

matthewjohns0n:
I'm having issues just loading entries listing on this branch. When I switch back to `7.dev` it loads quickly. 

Screenshot 2024-02-02 at 5 07 11 PM

Another thing to note - this still happens after I uninstall structure.

intoeetive:
@matthewjohns0n good catch!

It looks like while the `ColumnFactory` class is full of static methods, we were really not calling those in a way that would utilize the benefits. And that caused some sort of recursion. I have pushed the commit that is fixing those (not all related to Structure) - please review

matthewjohns0n:
I know this is not in the 7.4.0 release anymore, but I did pull and test, and can confirm the feature now works and no longer has memory issues.

PR #3960, opened 2023-12-18
Resolved #3959 where sort order of categories in Entry Manager filter could have been wrong

### LINKED ISSUE #3959: Entries - Category filter - list as per Category structure
### Discussed in https://github.com/ExpressionEngine/ExpressionEngine/discussions/3958

<div type='discussions-op-text'>

<sup>Originally posted by **ic360-com** December 16, 2023</sup>
### Suggest an idea for ExpressionEngine

When filtering listed Entries by Category... I think it would be really helpful if the Categories in the filter dropdown, were displayed in the same sequence (and indented) as per the Category setup.
Seems like they're in a random order at the moment.

Screenshot 2023-12-16 at 16 05 47
Screenshot 2023-12-16 at 16 06 05
</div>

intoeetive:
Resolved #3959 where sort order of categories in Entry Manager filter could have been wrong


bryannielsen:
@intoeetive could you take a look at my addition.  I was testing with an example where the category names were duplicated across groups and it was confusing to pick a filter without knowing the group.  Here are some before/after screenshots to illustrate the change

Before:
before

After:
after

intoeetive (commented):
@bryannielsen I have no objections, but one comment.

If there is just one category group - either globally or for selected channels - we need not to display the header (this is same behavior as we have in Categories tab for entry)

This is a nice addition, however it's not a bug fix as such, rather the feature on its own. And now that we need a change, it's preventing us from releasing a fix - so I'll move this to a later milestone.

bryannielsen:
These tests are actually passing https://cloud.cypress.io/projects/a8u2xo/runs/2821/overview there was some kind of issue with when the cypress job started and chrome versions being different on two of the workers.

bryannielsen (approved):

PR #3985, opened 2024-01-10
Resolved #3807 where EE Debug's Font Awesome CSS tag is too aggressive 

### LINKED ISSUE #3807: EE Debug's Font Awesome CSS tag is too aggressive
When EE debug is enabled EE loads a custom stylesheet to support this element of the output.
The CSS is loaded late in the build of the page, and it is not clear that it would be possible to load any other CSS after the debug CSS loads.
The EE6 debug CSS includes a rule that makes an aggressive selection of font awesome tags - selecting any tag with `.fa` or `.fas` classes, but also links these to the Font Awesome 5 font-family for the "Free" icon list.
On installations that use either earlier versions of Font Awesome in their setup, or are using icons that do not appear in the Font Awesome 5 free icon list, this aggressive selector effectively removes icons from the template output - by replacing the actual icon with an empty square box except in the cases where the icon used overlaps with a FA5 free icon.

**How To Reproduce**
As described

Debugging disabled
image

Debugging enabled
image

**Environment Details:**
 - Version: 6.4.12
 - PHP Version 8.1

**Possible Solution**
Limit the `.fa` and `.fas` selectors within the ee-debug css to only apply to elements within the ee-debug div.

Yulyaswan:
Resolved #3807  where EE Debug's Font Awesome CSS tag is too aggressive 

moved Font awesome files for debugger inside `ee-debugger` class to prevent rewriting custom styles when the debugger is on

EE6 version of #4002

### DIFF
--- a/cp-styles/app/styles/components/_debugger.scss
+++ b/cp-styles/app/styles/components/_debugger.scss
@@ -1,6 +1,8 @@
 
 // Has the ee prefix since it's a standalone component
 .ee-debugger {
+	@import 'vendor/fontawesome/scss/_fontawesome';
+	@import 'vendor/fontawesome/scss/_solid';
 	// Import the tab bar since these styles can be loaded on the front end
 	@import '_tab-bar';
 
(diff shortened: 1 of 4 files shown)


intoeetive:
@Yulyaswan do we need the same fix for EE7?

bryannielsen (approved):

PR #3986, opened 2024-01-10
Add Named Navs to Structure

matthewjohns0n:
Add the concept of "Named Navigations" to Structure. This allows you to specify a name to your navigation in your nav tags, and then hide certain entries from a single nav.

Instructions:

- Go to Module settings and scroll to "Named Navigations"
- Add new navigations. Example: Top, Side (You can add as many or as few as you need)
- Once these have been added, you can specifiy the name of a navigation in the templates, by using the "named_nav" tag parameter. This is how structure knows which tag actually is a "Side" nav or a "Top" nav. Your updated templates will look something like this:

```
{exp:structure:nav_advanced
  show_depth="1"
  start_from="/{segment_1}/"
  named_nav="Top"
}
```
- You can then hide entries from a particular navigation from the Structure tag in the Entry Create/Edit screen.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/addon.json
+++ b/system/ee/ExpressionEngine/Addons/structure/addon.json
@@ -1,7 +1,7 @@
 {
     "name": "Structure",
     "shortname": "structure",
-    "version": "6.1.1",
+    "version": "6.1.0",
     "description": "Create pages, generate navigation, manage content through a simple interface and build robust sites faster than ever.",
     "license": "ExpressionEngine",
     "namespace": "ExpressionEngine\\Structure",
--- a/system/ee/ExpressionEngine/Addons/structure/addon.json
+++ b/system/ee/ExpressionEngine/Addons/structure/addon.json
@@ -1,7 +1,7 @@
 {
     "name": "Structure",
     "shortname": "structure",
-    "version": "6.0.0",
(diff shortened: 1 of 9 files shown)


matthewjohns0n:
@Yulyaswan We still need to style the settings page of this one. Is that something you could work on?

system/ee/ExpressionEngine/Addons/structure/addon.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "Structure",
     "shortname": "structure",
-    "version": "6.0.0",
+    "version": "6.1.1",
```
intoeetive: Why is the version going to 6.1.1 and not 6.1.0? (the PR for  version with support for conditional access is setting 6.1.0)

Yulyaswan:
> @Yulyaswan We still need to style the settings page of this one. Is that something you could work on?

@matthewjohns0n I added styles
please fix lang problems ( added screenshot)
Screenshot 2024-01-11 at 19 42 55
Screenshot 2024-01-11 at 19 43 34

system/ee/ExpressionEngine/Addons/structure/addon.json
matthewjohns0n: I'll update that to 6.1.0.

PR #4005, opened 2024-01-16
Enabled specifying database for the Redis cache driver in config file; #3997

### LINKED ISSUE #3997: Add config for specifying Redis database
When creating a Redis connection it should be possible to choose a database through the configuration settings.  

Currently you can specify connection details in `config.php` 

```
$config['redis'] = array(
  'host' => '127.0.0.1',
  'password' => NULL,
  'port' => 6379,
  'timeout' => 0
);
```

It would be helpful if you could also specify a database to use:

```
$config['redis'] = array(
  'host' => '127.0.0.1',
  'password' => NULL,
  'port' => 6379,
  'timeout' => 0,
  'database' => 2
);
```

[The original post for this request can be found on the ExpressionEngine Forum](https://expressionengine.com/forums/topic/253818)

bryannielsen:
This change allows specifying a database for the Redis cache driver to use.

```
$config['redis'] = array(
  'host' => '127.0.0.1',
  'password' => NULL,
  'port' => 6379,
  'timeout' => 0,
  'database' => 2
);
```

Closes #3997

### DIFF
--- a/system/ee/legacy/libraries/Cache/drivers/Cache_redis.php
+++ b/system/ee/legacy/libraries/Cache/drivers/Cache_redis.php
@@ -179,7 +179,8 @@ class EE_Cache_redis extends CI_Driver
             'host' => '127.0.0.1',
             'password' => null,
             'port' => 6379,
-            'timeout' => 0
+            'timeout' => 0,
+            'database' => 0
         );
 
         if (($user_config = ee()->config->item('redis')) !== false) {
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -179,7 +179,8 @@ protected function _setup_redis()
             'host' => '127.0.0.1',
             'password' => null,
             'port' => 6379,
-            'timeout' => 0
+            'timeout' => 0,
+            'database' => 0
```
intoeetive, line 183: Is there a reason why the default is integer and not empty string or null?

intoeetive (approved):

system/ee/legacy/libraries/Cache/drivers/Cache_redis.php
bryannielsen: Good question, the default setup for Redis involves 16 databases that have IDs 0-15.  The default database value will be set to 0 to use the first one available.

PR #4007, opened 2024-01-16
Fix site_url issue in quick links

matthewjohns0n:


robinsowell:
So I don't trigger errors and the links seem to work, but it won't let me add an internal url with more than 1 get variable.  If I go to entries and filter by channel and add it, it's stripping the get. And if I manually add it, it objects.  Seems to not object with 1 variable, but two triggers a security validation:

Screenshot 2024-01-16 at 8 28 02 PM

Screenshot 2024-01-16 at 8 27 51 PM

Also, when I save some of them, it just sits on the page with no indication it saved, but if I go to the Shortlinks page, it did.

So I go to Quicklinks manager, add a new one- it's the frontend blog entry https://6link:8890/index.php/blog/entry/bacon-blog

After I hit save, I just reload:
Screenshot 2024-01-16 at 8 44 34 PM

robinsowell (changes requested):
I'd approve if it's acting as expected, but see comments.  Won't pass validation with multiple get values and doesn't always give an indication shortlink saved.

system/ee/legacy/libraries/Redirect.php
```diff
@@ -30,20 +30,17 @@
...
-        $protocol = "https://";
-    } else {
-        $protocol = "http://";
-    }
+    $protocol = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on') ? 'https://' : 'http://';
 
     // if the url starts with /, add the http host
     if (substr($_GET['URL'], 0, 1) === '/') {
-        $_GET['URL'] = $_SERVER['HTTP_HOST'] . $_GET['URL'];
+        $_GET['URL'] = ee()->config->site_url($_GET['URL']);
```
intoeetive, line 37: The site_url and base_url can be set to empty string in config file. So I'm not sure we can rely on those here (and also  we're processing HTTP_HOST later in this file anyway)

Additionally, this is kinda strange function because it's checking for `base_url` and not `site_url` so I wonder it we should completely deprecate it in favor of `ee()->function->create_uri()`

intoeetive (changes requested):

PR #4013, opened 2024-01-23
Added AJAX validation for folders in File manager; #3818

Yulyaswan:
Added validation to "New Folder" dialog in the Files area to prevent users from trying to create folders with unacceptable symbols in a file name
Resolved #3818


intoeetive:
@Yulyaswan can you resolve conflict here as well?

themes/ee/asset/javascript/src/common.js
```diff
@@ -1237,6 +1237,69 @@ $(document).ready(function(){
...
+				$.ajax({
+					url: action,
+					type: 'POST',
+					data: data+'&ee_fv_field='+that.attr('name'),
+					dataType: 'json',
+					error: function(err) {
+						console.log('err', err);
+					},
+					success: function(response) {
+						console.log(response);
```
intoeetive: I think this needs to be removed?

themes/ee/asset/javascript/src/common.js
```diff
@@ -1237,6 +1237,69 @@ $(document).ready(function(){
...
+						if(response['alert:inline:files-form']) {
+							that.parents('fieldset').addClass('fieldset-invalid');
+							that.css({'border':'1px solid var(--ee-error)', 'box-shadow':'none'});
```
intoeetive: can we apply class here instead of direct style?

intoeetive (changes requested):
Please see the comments in code.

Also, this seems to be very specific (used only on Files controller) and yet included into common.js

I think we need to either have it included on Files page only, or (and I think that would be better) rewrite it in the way so it could be applied to any form in modal.

intoeetive (changes requested):
@Yulyaswan sorry, I wasn't clear enough.

We don't need to call validation on every form on popup. We just need a way to be able to call it on the forms where we need that. For the time being, we only need validation for "new folder" and "rename folder" dialog. I think on channel entry form, the validation is triggered by the class ` class="ajax-validate"`, so perhaps we could do the same here

intoeetive (commented):
@Yulyaswan this works fine - however what I noticed is that the validation is only triggered when you blur out of the field.

This is same as with the "regular" forms, but because we have just 1 field here, this might never happen

So I wonder if we need to change the `init` for validation, so that of there's just one text input from `_textInputSelectors` it would trigger ajax not just on blur but also on like "when they make a pause in typing"

intoeetive (approved):

intoeetive (changes requested):
@Yulyaswan this now seems to be sending multiple requests each time I type

intoeetive:
@Yulyaswan just a reminder on this one

intoeetive (dismissed):

PR #4020, opened 2024-01-30
Improve Structure Tab Performance

bryannielsen:
Remove unnecessary and expensive query for data in structure tab

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/tab.structure.php
+++ b/system/ee/ExpressionEngine/Addons/structure/tab.structure.php
@@ -718,9 +718,14 @@ class Structure_tab
 
         // PARENT BUG
         // Update: Changed this to `$entry_id` which may cause some oddity with the Parent dropdown.
-        $data = $this->sql->get_data($entry_id);
+        // $data = $this->sql->get_data($entry_id);
 
         foreach ($data as $eid => $entry) {
+            // If we have an entry then this entry and its children
+            // cannot be an option to use as its own parent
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/structure/tab.structure.php
```diff
@@ -718,9 +718,14 @@ public function get_parent_fields($entry_id, $data)
 
         // PARENT BUG
         // Update: Changed this to `$entry_id` which may cause some oddity with the Parent dropdown.
-        $data = $this->sql->get_data($entry_id);
+        // $data = $this->sql->get_data($entry_id);
```
intoeetive, line 721: We are calling this function in just one place, where $data is present, so I think this is good removal of duplicate query

system/ee/ExpressionEngine/Addons/structure/tab.structure.php
```diff
@@ -718,9 +718,14 @@ public function get_parent_fields($entry_id, $data)
...
 
         foreach ($data as $eid => $entry) {
+            // If we have an entry then this entry and its children
+            // cannot be an option to use as its own parent
+            if ($entry_id && ($eid == $entry_id || $entry['parent_id'] == $entry_id)) {
+                continue;
+            }
```
intoeetive: This does not work for entries that are grand-children and grand-grand-children, and it causing weird formatting in the dropdown.
While some cleanup would be good idea, this particular approach is not working 100% and as such I don't think it should be in.

intoeetive (changes requested):

intoeetive (approved):

PR #4021, opened 2024-01-31
Resolved issue where module tabs where triggering saving during conditional field synchronization

intoeetive:


bryannielsen:
I think it makes a lot of sense to isolate some of these extra ChannelEntry model behaviors from the conditional field sync but I don't love the idea of adding more config variables to control this.  I'm wondering if we can approach this a little differently and just target the specific data we want to update.  Currently in the CLI command we're looping through entries and calling this: 

```
// Evaluate the conditions and save
$entry->evaluateConditionalFields();
$entry->save();
```

But the ChannelEntry model's onBeforeSave() is calling `evaluateConditionalFields()` again so we're effectively calling it twice.  If we don't need to do anything other than evaluate the conditional fields and save that relationship I think this might be better:

```
// Evaluate the conditions and save
$entry->evaluateConditionalFields();
$entry->HiddenFields->save();
```
I did a small test locally with 6 entries in a channel and one conditional field and the database load changes pretty significantly:

Before: `317 queries in 0.592300 seconds`
After: `91 queries in 0.046100 seconds`

This would also bypass the extra code being called in ChannelEntry's `onBeforeSave()` and `onAfterSave()` which handles storing autosave revisions, updating file usage, updating, entry statistics and saving the tab data.

I think the only thing we may want to add to the CLI command from `onAfterSave()` would be the cache clearing which I think could happen at the very end of the command instead of after each entry 'save'.

```
// clear caches
if (ee()->config->item('new_posts_clear_caches') == 'y') {
    ee()->functions->clear_caching('all');
} else {
    ee()->functions->clear_caching('sql');
}
```

intoeetive:
@bryannielsen would you mind opening a PR with your suggested changes?

bryannielsen:
@intoeetive here's a PR with what I was proposing - https://github.com/ExpressionEngine/ExpressionEngine/pull/4028

PR #4028, opened 2024-02-05
Improve performance in conditional field sync utility

bryannielsen:
Only save the HiddenFields relation after the logic is evaluated.  This reduces unnecessary side effects from saving the entire ChannelEntry model.


intoeetive (approved):
Worked well for me

matthewjohns0n (approved):

PR #4029, opened 2024-02-05
Added fallback for PHP `tmpfile()` function when it's not enabled

bryannielsen:
This PR aims to improve the user experience when ExpressionEngine is running in an environment where `tmpfile()` has been disabled.  We provide a polyfill that gives a similar experience to the native tmpfile() while also adding an entry to the developer log to indicate the fallback is being used.  It is still preferable to use PHP's native tmpfile function when possible.

You can test this behavior by updating your `disable_functions` line in `php.ini` to include `tmpfile` like this

```

disable_functions = 'tmpfile'

```

### DIFF
--- a/system/ee/ExpressionEngine/Boot/boot.common.php
+++ b/system/ee/ExpressionEngine/Boot/boot.common.php
@@ -660,6 +660,7 @@ if (!function_exists('array_key_first')) {
         foreach ($arr as $key => $unused) {
             return $key;
         }
+
         return null;
     }
 }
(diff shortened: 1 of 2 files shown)


intoeetive (commented):
@bryannielsen I did not test it, just looked at the code and here's what I see:
- there is `make()` function but it does not seem to be called
- there are PSR-12 issues shown here on GitHub 

Other than that, the code looks good

bryannielsen:
@intoeetive the PSR-12 issues should be resolved now, the `make()` method was added so the factory could be used directly by other code by directly calling `TempFileFactory::make()` instead of `tmpfile()`.  I haven't updated any existing code to this at the moment but thought it made sense to add now given the factory pattern and possible future use.

matthewjohns0n:
I also didnt test this, but the code looks good to me. I'll test it and then approve it as long as it's working fine.

intoeetive (approved):
worked well on my test

Issue #3741, opened 2023-09-01
New fields added to Field Groups not available to existing content within a Fluid Field


pyrobob83:
**Description of the problem**
If you have an existing Field Group, where that existing Field Group is already being used in a Fluid Field (such as a Page Builder) and there are existing entries that already have content utilizing the Field Group inside the Fluid Field (very typical site scenario)… and then you add a new field to the Field Group, all the existing instances of that Field Group inside the Fluid don’t see the new field - only new ones do. 

**How To Reproduce**
1. Create a Field Group with a set of fields
2. Create a Fluid Field (eg Page Builder) and assign the Field Group to it
3. Assign the Fluid Field to a channel
4. Edit an entry from that channel and add content to the Field Group within the Fluid Field
5. Go back to the Field Group and add a new field
6. Retry editing that entry and note you don't see the new fields for existing uses

I'm hoping this is an obvious bug and this wasn't intentional or some kind of technical limitation, but let me know either is true.

intoeetive:
This is a known limitation

Adding new Group (even if it's same group) would add new fields (with changed order, if it was changed) - but the existing Fluid components are saved in the state when they were added, so the Group contents/order might be different from what you have

pyrobob83:
@intoeetive @TomJaeger Wow, that's pretty huge.  If this is truly a restriction and we can't figure out a way to update existing entries that utilize Field Groups inside a Fluid to reveal new fields, I don't see how using Field Groups in Fluid Fields can be considered useful at all.  Who would want a site (or component) that once setup and content is added, can't evolve?

Brian mentioned in Slack that Bloqs does the same thing but saving the entry again would then introduce the new fields in existing bloqs... I don't even view that as an acceptable solution, since it adds responsibility to the user to know that's what they need to do, but maybe it'll give some ideas in terms of how to tackle this.

How feasible would some kind of sync utility be (CMS or CLI) that can be run manually by the DEV and essentially finds all the uses of a particular Field Group and injects fields that have been added since being saved?  Totally open to other solutions, but please, please let's find a way to fix this.

JimStrive:
Did anyone come up with a solution to this as an interim measure? I've a big site with tonnes of these components which need new fields pushing to them!

bryannielsen:
[...]

PR #4030, opened 2024-02-05
Fluid Field Group Improvements

bryannielsen:
This PR addresses several issues and improvements for handling field groups within a fluid field.

- Fix bug where adding new fields to field group in fluid would not allow existing entries to accept content for that field #3741
- Fix Live Preview bugs with Fluid field group #4388 
- Add new fixed_order and order parameters to {fields} tag within fluid field group
- Default order of fields within {fields} tag is controlled by field order within the Field Group
- Add ability to use a field group's field tags outside of the {fields} tag loop

Documentation PR: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/866


system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
```diff
@@ -287,9 +287,13 @@ public function post_save($data)
...
                 // Existing field
-                if (strpos($key, 'field_') === 0 && (!defined('CLONING_MODE') || CLONING_MODE !== true)) {
+                if (strpos($key, 'field_') === 0) {
                     $id = str_replace('field_', '', $key);
-                    if (isset($fluid_field_data[$id])) {
+                    // If we have cloned an existing field we need to populate the fluid_field_data with its data
+                    if(defined('CLONING_MODE') && CLONING_MODE === true && !isset($fluid_field_data[$id])) {
+                        $clonedField = ee('Model')->get('fluid_field:FluidField')->filter('id', $id)->first();
+                        $group_key = 'group_' . $clonedField->group;
+                    }else if (isset($fluid_field_data[$id])) {
```
intoeetive: This code is not PSR-12 compatible (needs spaces)

intoeetive: @bryannielsen also there are code conflicts to resolve

images/uploads/index.html
TomJaeger, line 1: Should this file have been removed? @bryannielsen

bryannielsen: Nope, I think I missed it in the rebase. I'll put it back 👍

pyrobob83:
@bryannielsen Is there anything I can do to help with this? I'm happy to test things in a vanilla build if that's helpful.

bryannielsen:
@pyrobob83 that would be much appreciated if you're willing to do some testing!  
I think we're mostly concerned with existing installations and if this change affects the order of how fluid field groups and fields are output in templates.

pyrobob83:
@bryannielsen FWIW, this seems to be working as I'd expect it without any adverse effects.  I made sure I had multiple field groups available to my fluid field, made sure I had several blocks setup with fully populated content in an entry and noted the presentation on the frontend.  Then I manually replaced the 3 core files in the PR - I just overwrote the entire files to makes sure I didn't screw up something:
- system/ee/ExpressionEngine/Addons/fluid_field/Service/Tag.php
- system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
- system/ee/legacy/libraries/Fluid_field_parser.php

Then re-checked the presentation and it was identical so proceeded to add a new field to my field groups and those did indeed show up and became usable in existing entries and I was able to reorder the fields inside the field groups and have the order update accordingly on the Publish screen.  As far as I can see, this is what we need, though appreciate there could be some edge-case scenarios I'm not testing that could surface issues.

matthewjohns0n (approved):
Pulled and tested those changes, no issues.

PR #4031, opened 2024-02-06
Improved CP performance issue with relationships in grids #3966

### LINKED ISSUE #3966: CP performance issue with relationships in grids
ExpressionEngine 7.3.12

It becomes painfully slow when there are more than a couple of dozens of grid rows with multiple relationships (grand total of over a hundred relationships).

I suspect, it's because of 15MB of this:

Screenshot 2023-12-23 000703

And some of this:

Screenshot 2023-12-22 235322

Yulyaswan:
Improved CP performance issue with relationships in grids
Resolved #3966
In this PR, we are loading JSON after clicking the button for the relationship React component if the 'Defer field initialization' option is used.
It works for single relationship field

@intoeetive could you please have a look for next items:

1. an error for the Relationship inside the Grid (screenshot added).
Screenshot 2024-02-06 at 14 45 16

2. Relationship inside the Fluid doesn't work
Screenshot 2024-02-06 at 14 46 56

3. Relationship inside the Grid inside the Fluid doesn't work
Screenshot 2024-02-06 at 14 49 07


intoeetive:
@Yulyaswan I made necessary updates to support Grid and Fluid fields.

What I noticed however is that if you have a Relationship in Grid (and also regular Relationship field) the click on "edit" in Grid (but not in Regular field) is processed twice, so the AJAX is performed twice. Could you please check why this would be happening

PR #4032, opened 2024-02-06
Feature/3892/response send output hook

eric-lamb62:
Fixes an issue where the `extensions` property isn't available through the Installer/Updater

### DIFF
--- a/system/ee/ExpressionEngine/Core/Response.php
+++ b/system/ee/ExpressionEngine/Core/Response.php
@@ -93,10 +93,6 @@ class Response
      */
     public function send()
     {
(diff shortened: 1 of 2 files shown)


eric-lamb62:
@intoeetive @TomJaeger a very timely issue arose on a client site where I had added this hook but then they went to run the installer. Turns out (and I didn't know this) the `extensions` property isn't available through the Updater/Installer. So a simple check for existence was added.

intoeetive:
@mithra62 thanks - we already figured that and are including the fix into release (that was also the reason why the tests were failing)

PR #4035, opened 2024-02-07
Resolved #4033 where Menu Sets item could not have been moved to the bottom

### LINKED ISSUE #4033: When editing Menu Sets, you can't move a menu item to the bottom of the list
When you're in the Menu Manager, re-ordering items in a Menu Set, you can't move an item to the bottom of the list.

**How To Reproduce**
1. Create Menu Set with several items. system/index.php?/cp/settings/menu-manager/edit-set/1
2. Grab any Menu Item by the grab handle and try to move it to the bottom spot.
3. It can only be put in the next-to-last spot
4. (You can then move the last item up, and that item will fall to the bottom, but that's not how it should work)

**Error Messages**
None.

**Environment Details:**
 - Version: 7.3.12
 - PHP Version 8.2
 - All browsers

**Additional context**
1) While you're in there, the various field labels could use tweaking:
Member role(s)? -- remove ?
Choose the role(s) to apply this menu to. -- remove period
Menu Items -- lower case items
Manage this menu sets contents -- Manage this menu set's contents

2) Is this re-ranking code used elsewhere with a similar problem? I couldn't find anything with the same issue.

Yulyaswan:
Fixed #4033 where you cannot move a menu item to the bottom of the list while editing Menu Sets

Also fixed some typo issues

### DIFF
--- a/system/ee/language/english/settings_lang.php
+++ b/system/ee/language/english/settings_lang.php
@@ -538,9 +538,9 @@ $lang = array(
 
     'menu_dropdown' => 'Dropdown',
 
-    'menu_items' => 'Menu Items',
+    'menu_items' => 'Menu items',
 
-    'menu_items_desc' => 'Manage this menu sets contents',
+    'menu_items_desc' => 'Manage this menu set\'s contents',
 
     'menu_label' => 'Name',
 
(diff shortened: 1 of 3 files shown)


intoeetive (commented):
@Yulyaswan this seems to be a big change affecting all fields using select_list component? What are the things that it can possible break and which pages / fields you think would make a good test to confirm it's not breaking anything?

Yulyaswan:
@intoeetive 
I don't expect that this will break anything. It simply changes the parent to which sortable items are connected.
the similar elements:
Channel-> statuses;
Entry->categories;

intoeetive (approved):
I tested and could not spot any issues, so approving

PR #4041, opened 2024-02-14
Monolog Logging

intoeetive:
This is a draft PR to implement Monolog as logging library, `ee('Logger')` service, `logger.php` config file, new Logging settings and new Log manager.

A lot of things really, I recorded a 15 min video for PacketTide team where I rather awkwardly explain what has been done. https://drive.google.com/file/d/1WMAg5t3C4U5Q9Oj0-_QAk_Ae7XMjyufd/view?usp=sharing

It needs to be reviews and possible some adjustments made before it can be released.

Also need to mention I went through several iterations on how the config is built, that included team discussions, and the current one is probably optimal, considering logging is really targeting tech people as primary audience

Closes #3995


intoeetive:
It's been nearly an year, I think we need to review this and include this into next release @TomJaeger @bryannielsen

robinsowell:
Installing this branch and after install I get a mysql error.

SQLSTATE[42000]: Syntax error or access violation: 1072 Key column 'role_id' doesn't exist in table:
CREATE TABLE `exp_log_manager_views` ( `view_id` int(10) unsigned NOT NULL AUTO_INCREMENT, `channel` varchar(45) DEFAULT NULL, `member_id` int(10) unsigned NOT NULL, `name` varchar(128) NOT NULL DEFAULT '', `columns` text NOT NULL, PRIMARY KEY (`view_id`), KEY `role_id_member_id` (`role_id`, `member_id`) );ENGINE=InnoDB DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci

Screenshot 2025-10-09 at 15-22-47 Exception - ExpressionEngine

Note- to get past it, I just removed role_id from the key.  Based on the model, I don't see that the column exists so...

robinsowell:
Another install issue- after removing the role_id from that key, tried again.

BadMethodCallException Caught
No such method onAfterInsert.

ee/ExpressionEngine/Library/Mixin/Manager.php:144

Screenshot 2025-10-09 at 16-20-23 BadMethodCallException - ExpressionEngine

After poking a bit, for now am just adding to UploadDestination model.  I'm betting we'll have some logging added there?  

```
	
    public function onafterInsert()
    {
		
	}	
	
    public function onafterDelete()
    {
		
	}	
```

robinsowell:
OK- after that, new install, another 'No such method onAfterInsert.'

Screenshot 2025-10-11 at 6 26 49 PM

So added to model/Category/CategoryGroup.php a 
``
	
    public function onAfterInsert()
    {
	}
``

Was called in service/channelset/set.php

```
                foreach ($channel_data->cat_groups as $cat_group) {
                    $this->category_groups[$cat_group]->on('afterInsert', $fn);
                }
```

intoeetive:
@robinsowell pushed the fixes for this

PR #4061, opened 2024-02-19
Resolved issue where only one category group was shown in Categories …

intoeetive:
Resolved issue where only one category group was shown in Categories filter in Entry manager if the channel is selected

Resolved issue where Categories filter was empty if no channel is selected in Entry Manager


bryannielsen (approved):

robinsowell (approved):
tested out fine if dropped on 7.4.1-dp

PR #4070, opened 2024-02-21
Resolved #4045 where Relationship field did not fully respect "sort by date" setting

### LINKED ISSUE #4045: Relationship field entry date sorting is wrong when used within a Grid that's then used in a Fluid
When using a Relationship field within a Fluid the returned entry listing's order is wrong. Instead of showing the available entries by entry date order as specified in the Relationsship field config (Order by > Entry Date > Desc) the entry appear in "some" order (neither date, title).

Steps to reproduce:

Create a Grid field, add a Relationship field to it and relate a channel. Specify the order to Entry Date and Desc order.
Assign that Grid field to a Fluid field.
Create a new entry in the channel the Fluid was assigned to.

Assuming the related Channel has entries on different dates the returned entry listing in the relationship selection does not follow the entry date order but some random order. 

When using the same Grid field outside a Fluid the returned entry listing's order is as expected. 

EE version 7.4 (reproducable on 7.3.15, too). 

We did report this to EE support and Robin noticed that the entries returned in the listing actually all have the same entry date so the underlying MySQL query somehow sorts by entry date and entry ID in this scenario, but assigns the same entry date to all returned entries?

intoeetive:
Resolved #4045 where Relationship field did not fully respect "sort by date" setting

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/libraries/EntryList.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/libraries/EntryList.php
@@ -57,8 +57,9 @@ class EntryList
 
         $entries = ee('Model')->get('ChannelEntry')
             ->with('Channel')
-            ->fields('Channel.channel_title', 'title', 'status')
-            ->order($order_field, $order_dir);
+            ->fields('Channel.channel_title', 'title', 'status', 'entry_date')
+            ->order($order_field, $order_dir)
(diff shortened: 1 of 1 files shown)


intoeetive:
The failing test is fixed in version/7.4.2 branch

bryannielsen (approved):

Issue #3708, opened 2023-08-18
MySQL error when cloning with missing status


robinsowell:
Bit of an odd one- you can throw an error when cloning with a 'closed' status- but it is not universal.  Happened on one 7.3.11 install but not another:

![Screenshot 2023-08-17 at 14-29-41 Exception - ExpressionEngine](https://github.com/ExpressionEngine/ExpressionEngine/assets/1181219/7d4a6704-4eb2-49d4-8aa1-caa398461a58)

One was an upgrade at some point, one wasn't.  I suspect it has to do with how the 'closed' status is 'assigned' to the entry, and in some cases, it's just not and it's defaulting.  But when that happens, status drops out of the clone data and you get this.

Note- also reported in 7.3.9.

intoeetive:
@robinsowell if you go to `exp_statuses` table on that site - is "closed" even there? If it's not, then it hasn't been upgraded smootly and there are other things that can potentially go wrong

chunkitdesign:
<img width="1225" alt="Screenshot 2024-02-15 at 12 34 25 PM" src="https://github.com/ExpressionEngine/ExpressionEngine/assets/3195497/89f30f0a-d154-4eef-907d-daa0e2995b14">
@robinsowell, the issue still persists on ee 7.3.15

intoeetive:
@chunkitdesign might be not in changelog, but the fix for this has been included with 7.4.0 release

PR #4073, opened 2024-02-22
Resolved issue where cloning entry with Closed status was causing an error

intoeetive:
https://www.dropbox.com/scl/fi/4we8xrspi75jblu75mezr/Clone-issue.mp4?rlkey=6fpntx5ji9er8xbalrzz232hx&dl=0


chunkitdesign:
do you think this is the same bug as I reported earlier?
https://github.com/ExpressionEngine/ExpressionEngine/issues/3708

intoeetive:
Is probably the same. We had two issues with the same error, but of different origin. One has been fixed in 7.4, the other one is getting fixed here

Yulyaswan (approved):
Works for me

PR #4110, opened 2024-03-04
Added ability to send http status code with ajax response, #4108

dougblackjr:
Addresses #4108 

Adds ability to send status code, instead of a boolean value, for `send_ajax_response()` function.

Example:

`ee()->output->send_ajax_response(['success' => false, 'message' => 'not allowed'], 401)`

This will return the ajax response with a 401 ( Unauthorized) status code.

This is backwards compatible, so:

`ee()->output->send_ajax_response(['success' => false, 'message' => 'not allowed'])` will still return 200.

`ee()->output->send_ajax_response(['success' => false, 'message' => 'not allowed'], true)` will return 500

Let me know what other info is needed

### DIFF
--- a/system/ee/legacy/core/Output.php
+++ b/system/ee/legacy/core/Output.php
@@ -688,15 +688,19 @@ class EE_Output
      *
      * @access  public
      * @param   string
-     * @param   bool    whether or not the response is an error
+     * @param   bool|int    HTTP status code. If boolean, status code to send.
      * @return  void
      */
-    public function send_ajax_response($msg, $error = false)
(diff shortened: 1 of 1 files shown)


intoeetive:
@dougblackjr awesome, thank you very much!

Would you mind also updating the User Guide reflecting this change?

dougblackjr:
@intoeetive  Done and done! https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/819

intoeetive (approved):
thank you!

PR #4125, opened 2024-03-07
Added new Fluid hooks; #1643

### LINKED ISSUE #1643: Issues with the new Fluid field hooks

litzinger:
resolves #1643

  - Added fluid_field_after_update_field hook
  - Added fluid_field_after_add_field hook
  - Added additional parameters to fluid_field_get_field_data and fluid_field_add_field hooks
  - Move fluid_field_get_all_data hook, it was in the wrong place
  - Only attempt to save Fluid data if savable data exists

These are changes and additions made to hopefully allow Publisher to support Fluid field.

The addition of the conditionals checking if $values is empty or not lets the first hook return no data. In Publisher if saving a non-default language, we don't want the native Fluid field to be updated with non-default data.

The change in the foreach loop when displaying fields is hard to explain at this point because it's been so long since I originally tested it, but it seems like a logical change, even outside the context of these hook changes. I tested it with and without Publisher installed and it still saves and displays Fluid field data just fine.

I'm pretty certain that the fluid_field_get_all_data hook was just in the wrong file. There are two getFieldData methods in different files and I think it was some confusion when the hook was originally added.

With these changes I'm able to successfully save basic text fields in Fluid with drafts and translations. Default Fluid functionality remains unchanged. Grid and Relationship fields inside of Fluid are a different beast that I need to figure out, but I'm pretty certain those need to be managed outside the context of any hooks added to the Fluid field.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
@@ -177,6 +177,15 @@ class Fluid_field_ft extends EE_Fieldtype
 
         $fluid_field_data = $this->getFieldData()->indexBy('id');
 
+        if (ee()->extensions->active_hook('fluid_field_before_save') === true) {
+            $data = ee()->extensions->call(
+                'fluid_field_before_save',
+                $this,
+                $fluid_field_data,
+                $data,
+            );
+        }
+
         if (empty($fluid_field_data)) {
             return '';
         }
@@ -261,6 +270,13 @@ class Fluid_field_ft extends EE_Fieldtype
         $previous_field_group_id = null;
         $previous_group_key = null;
 
+        $collection = [
+            'added' => [], // Only fields added
+            'updated' => [], // Only fields updated
+            'deleted' => [], // Only fields deleted
+            'saved' => [], // All fields, added and/or updated, when entry is saved
+        ];
+
         $i = 1;
         $g = 0;
         $total_fields = count($data['fields']);
@@ -327,11 +343,34 @@ class Fluid_field_ft extends EE_Fieldtype
                     $thisFieldValue = array_filter($value, function ($k) use ($field_id) {
                         return strrpos($k, '_' . $field_id) === strlen($k) - strlen('_' . $field_id);
                     }, ARRAY_FILTER_USE_KEY);
-                    $this->addField($i, $group, $field_id, $thisFieldValue);
+                    $insertId = $this->addField($i, $group, $field_id, $thisFieldValue);
+
+                    $rowData = [
+                        'order' => $i,
+                        'group' => $group,
+                        'fieldId' => $field_id,
+                        'value' => $thisFieldValue,
+                        'dataId' => $insertId,
+                    ];
+
+                    $collection['added'][] = $rowData;
                 } else {
                     $this->updateField($fluid_field_data[$id], $i, $group, $value);
+
+                    $rowData = [
+                        'order' => $i,
+                        'group' => $group,
+                        'fieldId' => $fluid_field_data[$id]->field_id ?? 0,
+                        'value' => $value,
+                        'dataId' => $fluid_field_data[$id]->field_data_id ?? 0,
+                    ];
+
+                    $collection['updated'][] = $rowData;
+
                     unset($fluid_field_data[$id]);
                 }
+
+                $collection['saved'][] = $rowData;
 
                 $i++;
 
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
```diff
@@ -410,10 +410,21 @@ private function updateField($fluid_field, $order, $group, array $values)
         $fluid_field->order = $order;
         $fluid_field->save();
 
-        $query = ee('db');
-        $query->set($values);
-        $query->where('id', $fluid_field->field_data_id);
-        $query->update($fluid_field->ChannelField->getTableName());
+        if (!empty($values)) {
```
intoeetive, line 473: Worried about this check, will it not block removing all fields from Fluid to save it as empty?

Also similar one below

system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
```diff
@@ -593,17 +620,22 @@ public function display_field($data)
                             }, $field_data),
                             'field_name' => $field_group->short_name,
                         ]);
+
+                        $fields .= ee('View')->make($view)->render($viewData);
                     } else {
-                        $field = $field_data[0]->getField();
+                        foreach ($field_data as $field_datum) {
+                           $field = $field_datum->getField();
```
intoeetive: Please see PSR-12 notices on files tab in the PR, looks like you're missing some spaces

intoeetive (changes requested):
@litzinger thank you for submitting this!

Please see a couple of comments I posted on the code. 

Am I getting it right that including this PR is making #3927 unnecessary and merging it will close #1643 ?

Also asking @bryannielsen to review

system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
litzinger: No b/c that's handled by removeField(). I just did a test without Publisher installed just to verify, I was able to remove all fields from a Fluid field, saved, and it reloaded with an empty Fluid field.

system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
litzinger: Whoops.

litzinger:
> Am I getting it right that including this PR is making #3927 unnecessary and merging it will close #1643 ?

That is correct.

litzinger:
Hold off on merging this... I think I might need to add a hook to the very end of `Fluid_field_ft->save()` or `post_save()`

bryannielsen:
@litzinger I'm going to try and get some time into reviewing this soon since it's related to work I've been doing on #4030.  Are there more changes you're wanting to make before I take a look?

litzinger:
> @litzinger I'm going to try and get some time into reviewing this soon since it's related to work I've been doing on #4030. Are there more changes you're wanting to make before I take a look?

At this point no, I got it working with Relationship fields without further changes. 🤞 the Grid implementation goes smoothly.

litzinger:
I got Grid and Relationship field saving correctly, then moved onto the sync'ing feature that Publisher has (saving as published will update the draft content to match the published data), and when saving a draft as published, all heck breaks loose because it is not properly removing existing published data to replace it with the draft data. The very specific hooks don't give me a broader picture of what is happening when a save is occurring, so I'm exploring a new option, which is a before and after save hook. The Grid field does something similar by building a collection of added, removed, and updated rows. Maybe that'll work with Fluid as well. It'll take me a few days to get a clearer picture if this route will work instead of the previously suggested hooks.

https://github.com/ExpressionEngine/ExpressionEngine/pull/4125/commits/5413d8d21015c539318a179339fae3c50fba6c05

PR #4126, opened 2024-03-11
Enabled testing of email configuration from email settings page

intoeetive:
Enabled testing of email configuration from email settings page

https://github.com/ExpressionEngine/ExpressionEngine/assets/752126/f4caeafe-f373-464e-984c-6af24bc6cca4

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Settings/Email.php
+++ b/system/ee/ExpressionEngine/Controller/Settings/Email.php
@@ -222,15 +222,36 @@ class Email extends Settings
             ee()->view->set_message('issue', lang('settings_save_error'), lang('settings_save_error_desc'));
         }
 
+        ee()->javascript->set_global([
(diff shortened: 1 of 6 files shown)


matthewjohns0n (commented):
One potential issue I see with this is that it ignores the [email_send](https://docs.expressionengine.com/latest/development/extension-hooks/global/email.html) hook. That hook can set/change some data that might affect the outcome of the email, so I think it should be fired off with this test email.

intoeetive:
@matthewjohns0n what makes you think it's ignoring that? It's calling `ee()->email->send()` which is triggering the hook in turn. The code is the same that is on other places where we send email, but with enforced debugging.

matthewjohns0n:
Oh gotcha - I read that wrong. Looks good then!

matthewjohns0n (approved):

PR #4144, opened 2024-03-15
Resolved #4129 where CSRF token not getting refreshed when CP session type is "session only"

### LINKED ISSUE #4129: CSRF token not getting refreshed when CP session type is "session only"
When CP session type is set to session only, session identificator is appended to URL like this: `&S=a5644d41749e3c4e69befff96a662376`

When idle popup is shown and password is entered, this code gets executed:
```
$.getJSON(EE.BASE + '/login/refresh_csrf_token', function(result) {
```
which results in invalid URL, so token is never updated, causing logout on form submission

Yulyaswan:
Resolved #4129 where CSRF token not getting refreshed when CP session type is "session only"


intoeetive (approved):
tested, the requests are getting session ID appended correctly now

PR #4150, opened 2024-03-18
Resolved #4147 where label of dropdown select field for member was not parsed correctly

### LINKED ISSUE #4147: Member custom fields- select fieldtype label output broken - shows value instead of label
User ran into it upgrading from EE 7.3.x to EE 7.4.4.  Make a custom member field- make it a select with label/value.  Mine is 

Label -- Value

Value 1 -- val1
Value 2 -- val2
Value 3 -- val3

I select 'Value 2' for my profile- go view in a template:

```
{exp:member:custom_profile_data}
{m_select:label} = {m_select:value} = {m_select}
{/exp:member:custom_profile_data}
```

Output is:

val2 = val2 = val2

Should be:

Value 2 = val2 = and really not sure what the default for the single tag should be- same as regular select tag.

Per docs: Tip: Checkboxes, Multiselect and Selectable Buttons all default to showing the label. Radio Buttons and Selects default to showing the value. 

So a regular select, {m_select} should show the value, so that one's correct.  It's just the label jamming.  UNLESS the old member select field worked differently, which is not impossible.

intoeetive:
Resolved #4147 where label of dropdown select field for member was not parsed correctly

### DIFF
--- a/system/ee/legacy/libraries/api/Api_channel_fields.php
+++ b/system/ee/legacy/libraries/api/Api_channel_fields.php
@@ -222,7 +222,7 @@ class Api_channel_fields extends Api
     */
     public function fetch_custom_member_fields()
     {
-        ee()->db->select('m_field_id, m_field_name, m_field_fmt, m_legacy_field_data, m_field_type');
+        ee()->db->select('m_field_id, m_field_name, m_field_fmt, m_legacy_field_data, m_field_type, m_field_settings');
         $query = ee()->db->get('member_fields');
 
         $mfields = array();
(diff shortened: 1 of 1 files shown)


robinsowell:
The fix tests out for me, and the behavior is the same as with the channel select field.  I also checked the multi-select and it checks out and is the same as the channel field.

I will make one note on multi-selects, which I'm not sure were affected.

Per the docs: 

> Tip: Checkboxes, Multiselect and Selectable Buttons all default to showing the label. Radio Buttons and Selects default to showing the value. 

This is obviously not optimum, but it is legacy.

For both the select and the multi-select, if I use only the single tag, I get the label (in the case of the multi-select, a comma delimited version.  [Multi-select docs example of tag pair:](https://docs.expressionengine.com/latest/fieldtypes/multiselect.html):

Screenshot 2024-03-18 at 9 10 52 AM

I get:
Screenshot 2024-03-18 at 9 13 21 AM

Code:

```
{exp:member:custom_profile_data member_id="15"}
{m_select:label} = {m_select:value} = {m_select}<br>
Multiple:  {multi_select}<br><br>
{multi_select}
    Label/item: {item}<br> {!-- 1 --}
    Value: {item:value}<br> {!-- 1 --}
    Label: {item:label}<br> {!-- One --}
{/multi_select}
{/exp:member:custom_profile_data}
```
Channel entry parses the same.  Not sure if this is a recent change or not for multi-select, but it seems inconsistent with docs example.

intoeetive:
@robinsowell The documentation appears to be wrong here (and I will submit a PR)

`item` will ALWAYS return value as defined in https://github.com/ExpressionEngine/ExpressionEngine/blob/7.dev/system/ee/legacy/fieldtypes/OptionFieldtype.php#L509
`field_name` however will be rendered as string of either values or labels, depending on the actual fieldtype

intoeetive:
Docs update: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/828

robinsowell:
Screenshot 2024-03-18 at 11 01 48 AM

For my reference, same basic test in v3

robinsowell (approved):
Agreed- compared to v3 output is identical and fixed the member weirdness.

PR #4154, opened 2024-03-18
Resolved #4098 where Date field by default was not setting time to current

### LINKED ISSUE #4098: Date field- the time should default to 'now' rather than noon
Not sure if it's a bug or a feature request- but before, the time defaulted to 'now'.  Now, the time, if included, defaults to noon.

And users are requesting it work the way it used to as far as default.

Yulyaswan:
Resolved #4098 where Date field by default was not setting time to current

### DIFF
--- a/themes/ee/asset/javascript/src/cp/date_picker.js
+++ b/themes/ee/asset/javascript/src/cp/date_picker.js
@@ -135,11 +135,17 @@ EE.cp.datePicker = {
 			var selected = null,
 			    year     = null,
 				month    = null;
+				hour     = null;
+				min      = null;
+				sec      = null;
 
 			this.element = element;
 			this.calendars = [];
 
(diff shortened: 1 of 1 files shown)


robinsowell (dismissed):
Boom- worked for me!

themes/ee/asset/javascript/src/cp/date_picker.js
```diff
@@ -683,7 +697,7 @@ EE.cp.datePicker = {
 			return
 		}
 
-		elements.on('focus', function() {
+		elements.off('focus').on('focus', function() {
```
intoeetive: @Yulyaswan I'm slightly concerned with this, as it removes ALL `focus` listeners. I think we need to namespace the event here

intoeetive (approved):

PR #4161, opened 2024-03-21
Fix PHP warnings from string functions

bryannielsen:


system/ee/legacy/core/Input.php
```diff
@@ -988,6 +988,8 @@ public function _clean_input_data($str)
             return $new_array;
         }
 
+        $str = (string) $str;
```
intoeetive: Should we check if it's NULL or number instead, and if yes just return here and stop following execution? Similar to what you did in the other file here

bryannielsen: I suppose we could check for a `empty($str)` and return early if that's the case instead.  I just had warnings on an install the other day and made this fix quickly

intoeetive (approved):

Issue #4037, opened 2024-02-10
Synchronising Files Directory removes EXIF data from images


jcogs-design:
**Description of the problem**
If you either add an image to the Files System (e.g. by drag-drop to files directory etc) or put an image into a Files directory and then synchronise it, if the image has EXIF data attached this is removed.
This removal of EXIF data is not described in documentation and provides no utility to the user.

- EE does not currently have the ability to add EXIF data to an image, and removes it whenever it undertakes an image processing operation, so this removal of EXIF data is easy for a user to trigger if required.

There may be circumstances where EXIF data retention *and* use of the EE File system is required (for example if the user plans to process images using some other mechanism than EE's built-in file manipulations).

**Environment Details:**
 - Version: 7.3.15
 - PHP Version 8.2

**Possible Solution**
When adding files to a File Upload directory EE performs an unnecessary GD file save operation, which is both inefficient and removes EXIF data. Removing this unnecessary file save operation would preserve the EXIF data.

**Additional context**
EXIF data is used by some image capture devices to record information about the context of the image capture - for example mobile phone images may include information about location and exposure, and detail results of some image processing operations (such as rotation or flips applied to the image when captured).  Some devices can also include information about areas of the image identified as having key features, and what it thinks those features are (e.g. faces).  Access to this information where available might be of utility to users.

bryannielsen:
@jcogs-design does your upload destination have a height/width/size limit placed on it?  I just tested with an upload destination that had no limits and it doesn't appear to be removing the EXIF data.  

It looks like there is a special flag ([--enable-exif](https://www.php.net/manual/en/exif.installation.php)) to build PHP with if you want to use GD image for manipulations and keep EXIF data.  I haven't tested with using image magic as the manipulation library but I wonder if that preserves EXIF data by default.

jcogs-design:
No, no limits. 
Yes you could compel people to enable the `--enable-exif` flag but in general getting people to run custom builds of php is tricky when many use commercial hosting with (relatively) inflexible php build policies.
[...]

PR #4162, opened 2024-03-21
Fix issue #4037 where uploading a file could remove EXIF data

bryannielsen:
When an image is uploaded to an Upload Destination using a local filesystem adapter we will move the file in a way that preserves any EXIF data present.  Note that if any transformations are performed upon upload the EXIF data may be removed if the image processing library does not support EXIF.

### DIFF
--- a/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
+++ b/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
@@ -1131,13 +1131,33 @@ class Filesystem
     }
 
     /**
+     * Get the adapter for the filesystem
+     *
+     * @return Flysystem\AdapterInterface
+     */
+    public function getAdapter()
+    {
+        return $this->flysystem->getAdapter();
+    }
+
+    /**
+     * Determine whether or not the Filesystem is using a CachedAdapter
+     *
+     * @return boolean
+     */
+    public function hasCachedAdapter()
+    {
+        return ($this->getAdapter() instanceof Flysystem\Cached\CachedAdapter);
+    }
+
+    /**
      * Get the base adapter for the filesystem
      *
-     * @return void
+     * @return Flysystem\AdapterInterface
      */
     public function getBaseAdapter()
     {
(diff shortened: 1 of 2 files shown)


intoeetive (dismissed):
Did not test through different environments, but the code looks good to me

bryannielsen:
@intoeetive tests are passing now, I'm also hoping that @jcogs-design can chime in when he has a chance and see if this resolves the original issue.

jcogs-design:
Apologies for delay in responding to this.
All seems to work OK here.
One observation though. The edits to Upload.php include this comment

>          * Move the file to the final destination
>          *
>          * If we have an UploadDestination and it is not a local filesystem
>          * we will write the stream of data to that remote filesystem
>          *
>          * Otherwise we will act locally to avoid potentially losing EXIF data
>          * To deal with different server configurations we'll attempt to use
>          * copy() first.  If that fails we'll use move_uploaded_file().
>          * One of the two should reliably work in most environments
>          */
From what testing I've done it looks like the Flysystem file utilities preserve EXIF data (so if you upload an EXIF bearing file to an EE7 Cloud File S3 folder and then download it again, the downloaded file still has EXIF data attached). So possibly if this is true also for the Flysystem Local Adapter (and presumably it will be) then you maybe can avoid having the split local vs. remote code, and simply use Flysystem throughout.

HTH
🐾

bryannielsen:
@jcogs-design thanks for taking a look!  

That's interesting about what you're observing with remote filesystems and EXIF data.  The [current code](https://github.com/ExpressionEngine/ExpressionEngine/pull/4162/files#diff-1197c0cb37abd54c39e83a6e966b62359e9845362365d97ce1db534c3408fdedL354) is using Flysystem for the local filesystem as well as remote ones but that seems to be what causes the loss of EXIF data locally.  Did your environment that you were testing on have EXIF support compiled into PHP?

jcogs-design:
I'm not sure why basic file operations in php would remove EXIF regardless of whether php is built with the EXIF library installed.

PR #4169, opened 2024-03-25
Fix #4159 where on-the-fly image manipulations were not cleared

### LINKED ISSUE #4159: On-the-Fly Image Manipulations Don't Update When An Image Is Replaced
It appears that #740 is happening again - noticed with a client using  7.3.4 and tested in 7.4.4 and it's happening again in both.  See #740 for full details - I couldn't find/see a way to simply reopen that existing issue.

bryannielsen:
When uploading a new image with the same name as an existing image choosing to "replace" the existing image not properly clear dynamically generated image manipulations.

### DIFF
--- a/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
+++ b/system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
@@ -303,6 +303,46 @@ class Filesystem
     }
 
     /**
+     * Get a list of files within the $path that share the same prefix as the path's filename
+     *
+     * @param string $path
+     * @return array
+     */
+    public function filesMatchingPrefix($path)
+    {
+        $path = $this->normalize($path);
+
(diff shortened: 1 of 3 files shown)


intoeetive:
@bryannielsen I believe we have code somewhere that already does that, but it's probably not working well. Should that be removed?

bryannielsen:
This was the fix that was made on the original issue https://github.com/ExpressionEngine/ExpressionEngine/commit/ef16875db13412197723e479621aa1c5d54594e1 but I think that code was refactored/removed and never really accounted for these dynamic image manipulations.  We were correctly handling the pre-defined image manipulations when a file was replaced but not these "on-the-fly" manipulations.  I didn't see any other code trying to handle this but if you find anything I think it can be safely removed.

intoeetive:
Oh I see. I'm fine with the code then, although I did not test it

intoeetive (approved):

PR #4172, opened 2024-03-25
Resolved #4167 where filter dropdown sometimes requires more space

### LINKED ISSUE #4167: Member column dropdown display for French (foreign) language packs
Hi everyone,

I use the French language pack (v.7.4.5) in EE's control panel. Foreign translation sometimes require more space (width) as words might get longer than the native English ones. In the new Member area (that includes great new features), there is the column icon dropdown menu that allows admins to chose whatever column they wish to display in the Member page setup. Its width is hard coded and looks to narrow. Names are partially hidden and non fully readable. Since there is room to enlarge the dropdown menu width, can you manage to provide a correction?

AU

example

Yulyaswan:
Resolved #4167 where filter dropdown sometimes requires more space

### DIFF
--- a/cp-styles/app/styles/components/_filter-bar.scss
+++ b/cp-styles/app/styles/components/_filter-bar.scss
@@ -306,7 +306,7 @@
       }
 
       .dropdown {
-      	width: fit-content;
+      	width: 100%;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


intoeetive (changes requested):
@Yulyaswan I don't think this is the right fix, because it's breaking other dropdowns in filters, which are now less than 100%

As I mentioned, this is already working well in channel manager, so we probably just need to apply the correct class/style in member manager

intoeetive (approved):
This one works well, thank you

PR #4195, opened 2024-04-03
Resolved #4185 where the page scrolled up when the file field modal window was opened

### LINKED ISSUE #4185: File field UI issue
EE 7.4.5 testing in Firefox

With a file field, if you choose 'Select existing', it opens the modal but behind the modal, the page with the form 'snaps' to the top.  It's not always very obvious.  But, if you 'x' out of the modal, you're left at the top of the page rather than the field where you were adding the image.

I can also duplicate the behavior if I 'choose existing' and pick an image.  When I do that the first time, I'm still right at the field where I chose the image.  But if I 'x' out that image and go choose another image, it inserts it but I'm left at the top of the page.

Some kind of funky js going on there, but it's definitely disconcerting when it happens.

Yulyaswan:
Resolved #4185 where the page scrolled up when the file field modal window was opened


bryannielsen:
I think this fixes the behavior of bringing the scroll back to the field you were editing but when I try to choose an image I can no longer scroll within the modal for selecting an image further down in my upload location.  If I switch back to 7.dev it appears to work again.

bryannielsen (changes requested):
Now I'm having issues scrolling back to the top of the file modal.  I can't see the top of my files list or the search/filter bars in the file modal and I can't scroll at all.

My setup is a file field down at the bottom of the channel entry publish screen and a long list of files in my upload location.  I'm also making my window fairly narrow to make the scroll area bigger.

Yulyaswan:
I can't reproduce this

robinsowell (approved):
This definitely fixed it form me in a test- though granted, quirky one to test.  But I could duplicate before dropping in the fix, so I think it looks good.

bryannielsen (approved):

PR #4200, opened 2024-04-03
Resolved #4193 where Select dropdown could show incorrect values as selected

### LINKED ISSUE #4193: Select dropdown double selecting if -- is first option EE 7.4.6
Super weird- make a select field, value/lable.  For your options:

Screenshot 2024-04-02 at 3 24 58 PM

If you go to the entry, select either 0 or 1 and it grays out both and worse, if you save it, even if you pick 1 it goes in as 0.  It only happens if you start out with not 0 at the start.  If I remove -- row from the top option, it works like normal.  If I change -- to None string... it still breaks!

Screenshot 2024-04-02 at 3 26 00 PM

Yulyaswan:
Resolved #4193 Where Select dropdown double selecting if -- is first option

### DIFF
--- a/themes/ee/asset/javascript/src/components/select_list.es6
+++ b/themes/ee/asset/javascript/src/components/select_list.es6
@@ -66,7 +66,7 @@ class SelectList extends React.Component {
         // array of values for multi select
         var value = (multi) ? items[key] : key
         var newItem = {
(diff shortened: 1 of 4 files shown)


robinsowell (approved):
User tested- worked fine.

PR #4202, opened 2024-04-03
Resolved #4069 where incorrect fieldtype options were shown on load

### LINKED ISSUE #4069: Creating a new field shows irrelevant field options
When creating a new field, selecting a different field type shows only the relevant options for that field type. But on page load, the default field (Text Input) also shows options related to Relationship fields. It's not clear what the side effects might be of saving the new Field with these extra options.

image

**How To Reproduce**
1. Go to Fields
2. Click "New Field"
3. The selected field type should be "Text Input", but scrolling down shows fields for Relationship fields (and maybe others).
4. Change the field type to anything else, and then change it back to "Text Input". Scrolling down now only shows fields relevant to "Text Input"

**Environment Details:**
 - Version: EE v7.4.1

bryannielsen:
Resolved #4069 where incorrect fieldtype options were shown on load

EE7 Version of #4206


robinsowell (approved):
Fixed it!

PR #4203, opened 2024-04-04
Adding target="_blank" to field info

dougblackjr:
When creating a Date field, there is a small line on localization:

```
Choose how the field data should be localized. [More info](https://docs.expressionengine.com/v7/fieldtypes/date.html)
```

If you click this link, it opens in the current window, which loses all of the field group you set up, or the current field info.

This just adds `target="_blank"` so that doesn't happen.


bryannielsen (approved):

bryannielsen:
Looks good, thanks @dougblackjr !

PR #4211, opened 2024-04-08
Improved #4103 where made the add column button searchable for entry table filters

Yulyaswan:
Improved #4103 where made the add column button searchable for entry table filters


robinsowell (approved):
Yay!  Thanks for this one!

PR #4212, opened 2024-04-09
Fix comparison of absolute path with filesystem using relative prefix

bryannielsen:


system/ee/ExpressionEngine/Library/Filesystem/Filesystem.php
```diff
@@ -1142,6 +1142,11 @@ protected function removePathPrefix($path)
     {
         $prefix = $this->getPathPrefix();
 
+        // If path is absolute but prefix is not shift the prefix
```
intoeetive, line 1145: IDK what the initial issue was, so please forgive me if I'm getting it wrong.

I don't remember what the prefix can be - can it be like `C:\` on Windows? If yes, this code might be problematic.

Otherwise, looks good to me

PR #4221, opened 2024-04-11
Fixed an issue where reordering items in Menu Sets did not work

Yulyaswan:
Fixed an issue where reordering menu manager items would not work.


intoeetive (approved):
Tested, it's fixing reordering for me

PR #4222, opened 2024-04-12
Enable cloning of Menu Sets; #4214

intoeetive:
Enable cloning of Menu Sets; #4214

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Settings/MenuManager.php
+++ b/system/ee/ExpressionEngine/Controller/Settings/MenuManager.php
@@ -14,6 +14,7 @@ use CP_Controller;
 use ExpressionEngine\Service\Model\Query\Builder;
 use ExpressionEngine\Library\CP\Table;
 use ExpressionEngine\Model\Menu\MenuSet;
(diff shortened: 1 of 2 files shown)


robinsowell (approved):
works for me, though reminded me I wanted to make a feature request. which I shall go do elsewhere.

bakin1999:
When testing I had the following. When cloning menu set that contains a dropdown.  The dropdown itself will clone correctly but the rows in the dropdown will not transfer over:


empty_dropdown

system/ee/ExpressionEngine/Controller/Settings/MenuManager.php
```diff
@@ -218,6 +230,19 @@ private function form($set_id = null)
 
             if ($result->isValid()) {
                 $set->save();
+
+                if (defined('CLONING_MODE') && CLONING_MODE === true && $kids->count() == 0) {
+                    $originalKids = ee('Model')->get('MenuItem')->filter('set_id', $originalSetId)->all();
```
bakin1999, line 235: this part drops dropdown kids because their set id is set to null and their parent id is what needs to be referenced

PR #4241, opened 2024-04-15
Resolved #4227 where `expire_session_on_browser_close` config override was not always respected

### LINKED ISSUE #4227: Session expire_browser_on_session_close doesn't work with cookie consent
Configuration override should cause session cookies to expire on close, and I couldn't figure out why it wasn't working.

Looks like in Input class

```
        // check the cookie setting for expiration date
        if (is_numeric($data['expire'])) {
            $loadedCookieSettings = ee('CookieRegistry')->getCookieSettings($name);
            if (!empty($loadedCookieSettings)) {
			
			log_message('error', 'input first: ' . ' - ' . $data['expire']);
				
                if ($loadedCookieSettings['cookie_lifetime'] === null) {
                    $cookieSettings = ee('Model')->get('CookieSetting')->filter('cookie_name', $name)->first();
                    $cookieSettings->cookie_lifetime = $data['expire'];
                    $cookieSettings->save();
                    ee('CookieRegistry')->registerCookieSettings($cookieSettings);
                } else {
                    if (is_numeric($loadedCookieSettings['cookie_lifetime'])) {
                        $data['expire'] = ($loadedCookieSettings['cookie_enforced_lifetime'] !== null) ? $loadedCookieSettings['cookie_enforced_lifetime'] : $loadedCookieSettings['cookie_lifetime'];
                    }
                }
				
				log_message('error', 'input second: ' . ' - ' . $data['expire'] . ' _ ' . $data['name']);
            }
        }

```
That's overriding it with whatever is set in the cookie register settings.

Since you can set the duration in the cookie settings, seems like we should probably just remove this on? Possibly logging it as deprecated so they'll know what's happening if it was working prior to an upgrade and now doesn't.

intoeetive:
Resolved #4227 where `expire_session_on_browser_close` config override was not always respected

EE7 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/4242

### DIFF
--- a/system/ee/legacy/core/Input.php
+++ b/system/ee/legacy/core/Input.php
@@ -143,8 +143,8 @@ class EE_Input
         // Clean up the value.
         $data['value'] = stripslashes($data['value']);
 
-        // check the cookie setting for expiration date, unless config override is forcing the lowest value
-        if (is_numeric($data['expire']) && ! bool_config_item('expire_session_on_browser_close')) {
+        // check the cookie setting for expiration date
+        if (is_numeric($data['expire'])) {
             $loadedCookieSettings = ee('CookieRegistry')->getCookieSettings($name);
             if (!empty($loadedCookieSettings)) {
                 if ($loadedCookieSettings['cookie_lifetime'] === null) {
@@ -158,6 +158,11 @@ class EE_Input
                     }
                 }
             }
+        }
+
+        // let config override set session cookies lifetime
+        if (bool_config_item('expire_session_on_browser_close') && in_array($name, ['sessionid', 'last_activity', 'last_visit'])) {
+            $data['expire'] = 0;
         }
 
         // Handle expiration dates.
--- a/system/ee/legacy/core/Input.php
+++ b/system/ee/legacy/core/Input.php
@@ -143,8 +143,8 @@ class EE_Input
         // Clean up the value.
         $data['value'] = stripslashes($data['value']);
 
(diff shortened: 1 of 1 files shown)


robinsowell (commented):
So it's weird, but it doesn't seem to do anything with regard to the session expiration.  Also.... wouldn't this skip the cookie registry check for all cookies, not just sessions.  I'm pretty sure it's only that session cookie that's supposed to be affected by the setting.

robinsowell:
Also... the more I think on it, the more I start to lean toward removing it entirely.  We can manage the session cookie length in the cookie settings now, which I believe would keep it in line with the cookie consents.  Which it... probably should be? Though granted, the setting can only make it more private, not less.

Weird that didn't fix it, though.  But it definitely didn't for me.

intoeetive:
Right, you could just set the value to 0 and that will do it.

intoeetive:
@robinsowell upon further thinking, I think we still need to make override work for sake of upgraded sites.

Please check the new version I just pushed

robinsowell:
I'm not sure why, but it still doesn't seem to be working.

Screenshot 2024-04-22 at 1 26 55 PM

```$config['expire_browser_on_session_close'] = 'y';```

I tried logging into the backend, logging into the frontend- both running cookies only for session setting.  Running EE 7.4.8, I even grabbed a fresh system/ee folder, then dropped the new file into that.

PR #4242, opened 2024-04-15
Resolved #4227 where `expire_session_on_browser_close` config override was not always respected

### LINKED ISSUE #4227: Session expire_browser_on_session_close doesn't work with cookie consent

intoeetive:
…
EE 6 version of https://github.com/ExpressionEngine/ExpressionEngine/pull/4241

### DIFF
--- a/system/ee/legacy/core/Input.php
+++ b/system/ee/legacy/core/Input.php
@@ -144,8 +144,8 @@ class EE_Input
         // Clean up the value.
         $data['value'] = stripslashes($data['value']);
 
-        // check the cookie setting for expiration date
-        if (is_numeric($data['expire'])) {
+        // check the cookie setting for expiration date, unless config override is forcing the lowest value
+        if (is_numeric($data['expire']) && ! bool_config_item('expire_session_on_browser_close')) {
             $loadedCookieSettings = ee('CookieRegistry')->getCookieSettings($name);
             if (!empty($loadedCookieSettings)) {
                 if ($loadedCookieSettings['cookie_lifetime'] === null) {


TomJaeger:
It looks like we have an open PR for this for the EE 7 version of this mentioned above. In testing that it had some issues, which is why this was never merged... Closing this for now since the EE 6 Legacy is starting at the end of this month.

PR #4250, opened 2024-04-17
Added the focus on search input upon creating a new field type.

Yulyaswan:
Added the focus on search input upon creating a new field type.


bryannielsen (commented):
@Yulyaswan This looks great to me!  The one thing I'm noticing is that if you `tab` off of the input the select doesn't close.  But I'm not sure if that's really related to this code or the select component in general.  (Edit) I just tested the select in other places and it happens there too so not related to this PR

bryannielsen (approved):

PR #4253, opened 2024-04-18
Resolved issue where an entry without structure parent fails validation

bryannielsen:


robinsowell:
So I 100% do not think this is related, but just trying to break stuff.  I had a saved entry open.  Structure tree open.  I move the entry 3 levels deep in the tree.  Refreshed, tab dropdown fine.  I then move it back to top of tree.  Refreshed, dropdown showed it 1 level deep.  But if I open to edit in new tab, it's showing 'none' as it should.  If I go to the entry directly from structure tree, it shows none.  It's not browser caching, but I suspect it's some mysq/server caching.  And totally not related to this.  But I saw it, so I'll mention it.

Screenshot 2024-04-18 at 11 58 55 AM

robinsowell:
Also, dragging things around in the tree, if I log errors to file, I get a ton of

```
ERROR - 2024-04-18 12:06:31 --> Severity: E_NOTICE  --> Undefined offset: 1 /Users/robinsowell/Sites/74/system/ee/ExpressionEngine/Addons/structure/mcp.structure.php 348
ERROR - 2024-04-18 12:06:31 --> Severity: E_NOTICE  --> Undefined offset: 2 /Users/robinsowell/Sites/74/system/ee/ExpressionEngine/Addons/structure/mcp.structure.php 348
ERROR - 2024-04-18 12:06:31 --> Severity: E_NOTICE  --> Undefined offset: 45 /Users/robinsowell/Sites/74/system/ee/ExpressionEngine/Addons/structure/mcp.structure.php 348
ERROR - 2024-04-18 12:06:31 --> Severity: E_NOTICE  --> Undefined offset: 3
```
Which I tend to think is unrelated as well.

matthewjohns0n:
Just tested this and I dont think it works completely. If you set a structure parent entry, save, and then set the parent to NONE and save, the parent stays as the one that was set initially. I think we're going to have to let it save if parent_id is 0.

robinsowell (approved):

PR #4273, opened 2024-04-24
Enable custom fields to be populated with Bulk Edit; #4240

intoeetive:
Enable custom fields to be populated with Bulk Edit; #4240

Bulk Edit UI is essentially Fluid Field UI, so adding all fields that can be used in Fluid.

At the same time, the javascript does not seem to be initialized when you add a new field using Bulk Edit - @Yulyaswan is that something you can look into?

After we get javascript working, this needs to be tested, especially with complex fields like Relationships or Grid


ic360-com:
Looking forward to seeing this update.

PR #4301, opened 2024-05-08
Resolved #4291 where the move action was missing from the all files listing

### LINKED ISSUE #4291: File Upload - No 'move' option available in 'All Files'
In File Manager, when listing All Files... there is no 'move' option available in the dropdown menu.

EE 7.4.9
PHP 8.1

bryannielsen:
Resolved #4291 where the move action was missing from the all files listing


robinsowell (approved):
Tested dropped into my 7.4.9 just fine.

PR #4314, opened 2024-05-14
Resolved #4298 error backing up table with set type column

### LINKED ISSUE #4298: Error backing up database if field is set() type
User was trying to backup the database using the Utilities tool, and gets a 

```
Cannot Make Backup
Non-existant column requested: exp_m_accounts.account_active
```
I dig- field does exist.  I describe:

```
account_active 	set('YES','NO')
```

Turns out, when figuring out the mysql table field types, we do (service/database/backup/query.php):

```
    protected function getDataType($column_type)
    {
        $type = strtolower($column_type);

        if (strpos($type, 'binary') !== false or
            strpos($type, 'blob') !== false) {
            return self::BINARY_TYPE;
        } elseif (strpos($type, 'char') !== false or
            strpos($type, 'text') !== false or
            strpos($type, 'date') !== false or
            strpos($type, 'time') !== false or
            strpos($type, 'enum') !== false) {
            return self::STRING_TYPE;
        } elseif (strpos($type, 'int') !== false or
            strpos($type, 'float') !== false or
            strpos($type, 'double') !== false or
            strpos($type, 'decimal') !== false) {
            return self::NUMBER_TYPE;
        }
    }
```

And set is never included.  I think it's an easy fix, but I couldn't figure out whether it should be a string or number type or... idk.  So I'm dropping it here.  Rare case, but... it happens.

robinsowell:
Creating as draft pr until I hear back from end user

### DIFF
--- a/system/ee/ExpressionEngine/Service/Database/Backup/Query.php
+++ b/system/ee/ExpressionEngine/Service/Database/Backup/Query.php
@@ -449,6 +449,7 @@ class Query
             strpos($type, 'text') !== false or
             strpos($type, 'date') !== false or
             strpos($type, 'time') !== false or
+            strpos($type, 'set') !== false or
             strpos($type, 'enum') !== false) {
             return self::STRING_TYPE;
         } elseif (strpos($type, 'int') !== false or


robinsowell:
@bryannielsen or whoever gets the joy of reviewing, user got back to me and this did work for them.  Yay!

bryannielsen (approved):

PR #4322, opened 2024-05-16
Resolved #4317 where replacing an image would not create manipulations

### LINKED ISSUE #4317: 7.4.9 File upload replace existing file doesn't create image manips.
User explains it perfectly- and I replicated:

I upload an image with image manipulations (Constrain or Crop) in small and medium sizes, and then I get the following files:

image.png
_small/image.png
_medium/image.png

Then when I replace the image file (upload again) I will get this:

image.png
file deleted
file deleted

the replacement deletes the cropped files and does not create new ones.

So it replaces the main image and there's still a system thumbnail (I didn't double check if it's a new one) BUT any file manipulations are just gone.

bryannielsen:
Resolved #4317 where replacing an image would not create manipulations


robinsowell (approved):
WOOT!  Worked for me.

PR #4334, opened 2024-05-22
Resolved #4324 where add-on name with caps was snaked cased between caps

### LINKED ISSUE #4324: Stop eecli.php messing up capitalised words when generating addons ...
Say you want to create an add-on called "JCOGS Cats"
When the CLI thing generates the add-on the addon name becomes ... "j_c_o_g_s_cats"
The module class becomes "J_c_o_g_s_cats"
Not sure why this is.
It is tedious to have to go through and remove all the _ characters ... 

**Environment Details:**
 - Version: EE 7.4.9
 - PHP Version 8.3.6

**Possible Solution**
Don't insert the unnecessary _ characters ... ?

TomJaeger:
Resolved #4324 where add-on name with caps was snaked cased between caps

### DIFF
--- a/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
+++ b/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
@@ -45,7 +45,8 @@ class AddonGenerator
         $this->str = $str;
 
         $this->name = $data['name'];
-        $this->slug = $this->str->snakecase($data['name']);
+        $this->slug = str_replace(' ', '_', $data['name']);
+        $this->slug = strtolower($this->slug);
         $this->slug_uc = ucfirst($this->slug);
 
         // Setup the generator data
--- a/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
+++ b/system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
@@ -46,6 +46,7 @@ class AddonGenerator
 
         $this->name = $data['name'];
         $this->slug = str_replace(' ', '_', $data['name']);
+        $this->slug = str_replace('-', '_', $this->slug);
         $this->slug = strtolower($this->slug);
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Service/Generator/AddonGenerator.php
```diff
@@ -45,7 +45,8 @@ public function __construct(Filesystem $filesystem, Str $str, array $data)
         $this->str = $str;
 
         $this->name = $data['name'];
-        $this->slug = $this->str->snakecase($data['name']);
+        $this->slug = str_replace(' ', '_', $data['name']);
```
matthewjohns0n, line 48: I'm wondering if this will handle things like `-` or possibly other names passed in as punctuation. Can we just run `$this->str->snakecase($data['name']);` after processing the slug initially?

TomJaeger: So I think it will be fine. Since on input it goes through the following
`// Lets filter the name to only allow alphanumerics, "-", "_" and spaces
        $name = preg_replace("/[^A-Za-z0-9 \-_]/", '', $name);`

TomJaeger: https://github.com/ExpressionEngine/ExpressionEngine/blob/7.dev/system/ee/ExpressionEngine/Cli/Commands/CommandMakeAddon.php#L125

matthewjohns0n: Yes, but then we are still allowing `-` in the name. Maybe we want to add something like `$this->slug = str_replace('-', '_', $data['name']);`?

TomJaeger: Added

matthewjohns0n (approved):

PR #4351, opened 2024-06-10
Select specific fields from POST to add to the file data

matthewjohns0n:

### DIFF
--- a/system/ee/ExpressionEngine/Service/File/Upload.php
+++ b/system/ee/ExpressionEngine/Service/File/Upload.php
@@ -519,7 +519,21 @@ class Upload
 
         $action = ($file->isNew()) ? 'upload_filedata' : 'edit_file_metadata';
 
-        $file->set($_POST);
+        $fields = [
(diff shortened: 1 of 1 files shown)


bryannielsen (commented):
@matthewjohns0n this looks fine to me, but I think you could shorten it to something like this on Line 528 and drop the loop and conditional set.

```
$file->set(array_intersect_key($_POST, $fields));
```

or even more simply just do it all in the same line since there's no reason to keep $fields in it's own variable.

```
$file->set(array_intersect_key($_POST, [
    'title', 'description', 'credit', 'location', 'categories',
    'crop_width', 'crop_height', 'crop_x', 'crop_y',
    'rotate', 'resize_width', 'resize_height'
]));
```

matthewjohns0n:
@bryannielsen I like the `array_intersect_key` better, but I was thinking we might end up using the fields again, in which case we'd pull that variable out somewhere else. For now I'll just go with the one-liner though.

bryannielsen (dismissed):

bryannielsen (approved):

PR #4401, opened 2024-07-19
Resolved #4400 where addon control panel routes did not allow multiple segments

### LINKED ISSUE #4400: Addon Control Panel Routes do not work with multiple segments
When an addon attempts to use a multi-segment route in the control panel (example:  /item/edit) the $id parameter is not preserved.  It is re-written to the `action` parameter and the `id` parameter is therefore lost.

ee-forge:
Changes the `buildObject` method to assume the new Control Panel/Routes and fallback to the legacy Mcp/ folder.  This is the inversion of the original approach.

Fixes #4400


matthewjohns0n:
From what I can tell, this issue (#4400) was closed as not being a problem with the usage of `ee()->uri->segment(3)`. As such, I'm closing this PR.

PR #4418, opened 2024-08-08
Fix PHP warning when resetting password.

robinsowell:
PHP 8, when you reset the password on the frontend, you can get a 

Severity: E_WARNING  --> Undefined array key "username" /Users/robinsowell/Sites/6.1.5/system/ee/ExpressionEngine/Service/Validation/Rule/ValidPassword.php 41

May/may not show depending on redirects, but user saw the error and I see it in my logs.  Username isn't set when submitting the password change form.

Alternative fix would be to go get the username when missing.


intoeetive:
@robinsowell did you mean to submit a PR for this as well?

TomJaeger (approved):

PR #4435, opened 2024-08-24
Resolved #4429 where setting Number input to decimal did not allow decimals if step was not set

### LINKED ISSUE #4429: HTML5 validation not playing nicely with decimal numbers
So- create a number field, set the type to decimal.  Make a decimal.  html 5 validation will object.

If the type = number? It's just not liking it.

Screenshot 2024-08-19 at 6 00 27 PM

Weirdly- a text field set to either number OR decimal does just fine with a decimal, but that's because the type in those cases is still text.

Screenshot 2024-08-19 at 5 39 35 PM

And I think we need to clarify in the docs the difference between a number field and a decimal field- for both text and decimal.  But in a text field, it saves decimals ok, but drops the trailing zero if there is one.

Also note- if you switch back and forth between text and number types, the db doesn't always change.  So don't test that way.

intoeetive:
Resolved #4429 where setting Number input to decimal did not allow decimals if step was not set

### DIFF
--- a/system/ee/ExpressionEngine/Addons/number/ft.number.php
+++ b/system/ee/ExpressionEngine/Addons/number/ft.number.php
@@ -54,6 +54,10 @@ class Number_ft extends Text_ft
             $field['step'] = $this->settings['field_step'];
         }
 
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):

TomJaeger:
@robinsowell can you run quick test on this one to confirm it solves your issue above as well?

PR #4453, opened 2024-09-07
Resolved #4451 where "Resend activation email" link was not working correctly

### LINKED ISSUE #4451: CP Pending members don't show 'Resend email' option when it should.
In the CP member list, pending member toolbar should show a 'Resend email' option for member types that require an email of some sort:

Screenshot 2024-09-05 at 9 50 17 AM

But the conditional appears to be wrong.

cp/membermanager/columns/Manage.php:

```
        if ($member->role_id == Member::PENDING) {
            $toolbar['approve'] = array(
                'href' => ee('CP/URL')->make('members/approve/' . $member->getId()),
                'title' => lang('approve'),
            );
            if (ee()->config->item('req_mbr_activation') !== 'email' && ee('Permission')->can('edit_members')) {
                $toolbar['resend'] = array(
                    'href' => '#',
                    'title' => lang('resend'),
                    'rel' => 'modal-confirm-rename-file',
                    'class' => 'm-link',
                    'data-file-id' => $member->getId(),
                );
            }
            if (ee('Permission')->has('can_delete_members')) {
                $toolbar['decline'] = array(
                    'href' => '',
                    'class' => 'm-link with-divider',
                    'rel' => 'modal-confirm-decline',
                    'data-file-id' => $member->getId(),
                    'title' => lang('decline'),
                );
            }
        } else {
```
This bit:

```
 if (ee()->config->item('req_mbr_activation') !== 'email' && ee('Permission')->can('edit_members')) {
```

There are 3 options for member activation- 'none' they are just automatically in and pending really shouldn't matter, email where they get an email with a verification link in it, and moderator only, where the moderator manually approves and an email is sent out informing them (I think).

We definitely want a 'resend' for the 'email' option, which is the one we're saying DON'T show it for.  I can imagine wanting to resend the 'Moderated' one as well, but I don't think it's ever worked that way and haven't checked to see whether it would go out on a resend or not.  But at the lease, we want to resend the email verification.

I didn't just fix it because we really do need to double check what email(s) are going out when you click 'resend' and make sure all the logic works.

intoeetive:
Resolved #4451 where "Resend activation email" link was not working correctly

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Members/Members.php
+++ b/system/ee/ExpressionEngine/Controller/Members/Members.php
@@ -204,7 +204,7 @@ class Members extends CP_Controller
      * @param array $ids The ID(s) of the member(s) being approved
      * @return void
      */
-    public function resend(array $ids)
+    public function resend($ids)
     {
         if (! ee('Permission')->can('edit_members') || ee()->config->item('req_mbr_activation') !== 'email') {
             show_error(lang('unauthorized_access'), 403);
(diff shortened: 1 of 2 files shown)


robinsowell:
Note- user tried the patch and it worked fine for them.

bryannielsen (approved):

PR #4475, opened 2024-09-14
Enabled use of all file field variables within modifier tag pair; #4335

intoeetive:
Enabled use of all file field variables within modifier tag pair; #4335

e.g. {title}, {file_name}, {description} within {image:resize} tag pair

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -656,11 +656,11 @@ JSC;
             return ($return_as_path ? $destination_path : $destination_url);
         } else {
             // tag pair
-            $vars = [
+            $vars = array_merge($data, [
                 'url' => $destination_url,
                 'width' => $props['width'],
                 'height' => $props['height']
-            ];
+            ]);
 
             return ee()->TMPL->parse_variables($tagdata, [$vars]);
         }


jaredchesebro:
@intoeetive Any chance this is in the next release? This is something that I need frequently.

TomJaeger:
I think there's a good chance... any chance you would be willing to test the addition and confirm it's working as intended for you?

jaredchesebro:
> I think there's a good change... any chance you would be willing to test the addition and confirm it's working as intended for you?

@TomJaeger Sure I'll test it right now.

jaredchesebro:
@TomJaeger Worked as expected for native file field, file grid, and within a Bloqs field.

Only caveat is that it displays the file name, file size (in bytes), upload directory, and mime type of the original. I don't see any issues with that though for my use cases.

matthewjohns0n (approved):

TomJaeger:
As per internal conversation, we believe we should prefix at least the title so it doesn't stop over the main channel entries title. We will aim to release this in the next patch release.

jaredchesebro:
@TomJaeger I know there may have been some issues with this. For example some conflicting tags like mime type, and file dimensions. And some issues with the image and manipulation tag pairs not properly escaping eachother. Any chance that was resolved? This came up in conversation again today in EE Slack.

Issue #456, opened 2020-07-02
Reloading CP page after session timed out throws PHP warning about undefined constant BASE


jcogs-design:
**Description of the problem**
If you reload a page that was accessing a control panel page (e.g. with URL such as `https://domain.com/admin.php?/cp/fields&group_id=5`) a PHP warning is reported relating to an undefined contant (BASE) - vis:

> Warning
> Use of undefined constant BASE - assumed 'BASE' (this will throw an Error in a future version of PHP)
> 
> ee/legacy/core/URI.php, line 333
> 
>     Severity: E_WARNING

**How To Reproduce**
This can happen for example when returning to a page in a browser that was connected / logged in to an EE site CP and then was left for session to time out / expire.  The error appears on first refresh by browser, and then disappears when on next refresh the browser is directed to the CP log in screen.

**Error Messages**
> Warning
> Use of undefined constant BASE - assumed 'BASE' (this will throw an Error in a future version of PHP)
> 
> ee/legacy/core/URI.php, line 333
> 
>     Severity: E_WARNING

**Screenshots / Videos / Template Code**
![image](https://user-images.githubusercontent.com/13821249/86355976-6a559800-bc63-11ea-8e7a-dbbf489ac684.png)

**Environment Details:**
 - Version: 5.3.2
 - PHP Version 7.2.31
 - MySQL Version 5.7.30
 - OS: macOS 10.14.6
 - Web Server: Litespeed 7.6

**Possible Solution**
It appears that BASE is defined in /legacy/Core.php via private function `somebody_set_us_up_the_base()` and conjecture is that in refresh this library is not called, and so BASE is not set: the error links to a function `reformat` in legacy/URI.php.  A solution would be to trap for BASE not being set in the `reformat` function and if it is not set, trigger the setting of BASE to a value at that point.  The most direct way to do this would be to recast `somebody_set_us_up_the_base()` as a public function so that it could be accessed directly by the URI.php `reformat` function.

**Additional context**

intoeetive:
We are planning to address this in the next EE6 beta release

robinsowell:
Fixed in [beta 2](https://docs.expressionengine.com/v6/installation/changelog.html#beta-2-release-october-22-2020)

Issue #23, opened 2018-11-27
Control Panel "My Profile" page doesn't allow "https" URL scheme


h3xl3r:
**Description of the problem**
In the Control Panel navigating to "My Profile" from the top-left current user drop-down, trying to update the "Custom Field > URL" with anything starting with "https://" results in a validation error 
`Your URL must begin with a valid scheme:`

**How To Reproduce**
- Click on top left username drop down
- Select "My Profile"
- Scroll down to "URL" and enter an URL with https:// scheme
- Try to submit/savesave

**Error Messages**
<pre><code>
Your URL must begin with a valid scheme: 
</code></pre>

**Screenshots / Videos / Template Code**
![screen shot 2018-11-27 at 10 28 47](https://user-images.githubusercontent.com/2123/49052664-2b15c880-f230-11e8-8e81-c1ab4b5a3aa5.png)

**Environment Details:**
 - Version: 5.0.1
 - PHP Version 5.6.38

jmathias:
That is a custom member field, when setting it up you should see an option to allow `https://` schemes as well as several others.

I'm unable to replicate your bug, unless I turn off that scheme for my custom URL field.

You can edit your custom member fields under Members > Member Fields

h3xl3r:
OK, I see. Thing is, I didn't set up this field. It was just magically there ever since I installed and since upgraded from Expression Engine 2.x and associated forum module.

jmathias:
Are you able to get it working by editing the field under Members > Member Fields?

h3xl3r:
Actually, no. "https://" was already checked there under "Allowed URL Schemes"

![screen shot 2018-11-27 at 10 52 49](https://user-images.githubusercontent.com/2123/49053240-b98b4980-f232-11e8-816b-840f68d86b18.png)

h3xl3r:
Un-checking it here, saving, then re-checking it and saving again did the trick though.

jmathias:
Great! I was just going to recommend doing that.

h3xl3r:
Well, ok. It is fixed for **me** now.

Have to admit though, if my code didn't recognize an option that is clearly begin displayed as enabled in the UI, I'd still call it an "issue". And not an _invalid_ one at that.

jmathias:
Unfortunately without steps to reproduce we don't have anything to take action on. 

I cannot reproduce this issue on a fresh install, nor an install that has been upgraded over time from version 2. 

[...]

PR #4476, opened 2024-09-14
Added ability to search using file ID in File Manager; #4471

intoeetive:
Added ability to search using file ID in File Manager; #4471

### DIFF
--- a/system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
+++ b/system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
@@ -126,7 +126,11 @@ trait FileManagerTrait
         $search_terms = ee()->input->get_post('filter_by_keyword');
 
         if ($search_terms) {
-            $files->search(['title', 'file_name', 'mime_type'], $search_terms);
+            if (is_numeric($search_terms) && strlen($search_terms) < 3) {
+                $files->filter('file_id', $search_terms);
+            } else {
+                $files->search(['title', 'file_name', 'mime_type', 'file_id'], $search_terms);
+            }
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):

system/ee/ExpressionEngine/Library/CP/FileManager/Traits/FileManagerTrait.php
```diff
@@ -126,7 +126,11 @@ protected function listingsPage($uploadLocation = null, $view_type = 'list', $fi
         $search_terms = ee()->input->get_post('filter_by_keyword');
 
         if ($search_terms) {
-            $files->search(['title', 'file_name', 'mime_type'], $search_terms);
+            if (is_numeric($search_terms) && strlen($search_terms) < 3) {
```
litzinger, line 129: Does this mean that a file_id of 3 or more digits can't be searched? E.g. I can search for file `#23`, but not `#456`?

litzinger: Yeah I just did a test for file `17` locally, and it returned file_id=17, then I tried `119`, and it returned 3 files, all had `119` in the file name (they're really long random file names), but it did not return the file with the actual file_id=119.

bryannielsen: That's interesting, the `search()` on line 132 should be looking in the file_id field as well now.  But I just did a similar test and got the same result as you @litzinger.  This needs another look.

I think I would also prefer doing a more performant `filter()` when the search term looks like `id:12345`.  But regardless we should be searching inside the `file_id` along with the other fields as this original change intended.

PR #4477, opened 2024-09-14
Resolved #924 where template routing did not work properly with optional segments

### LINKED ISSUE #924: Long-standing issue with template route matching where final segment is optional
This report concerns a long-standing issue with template route matching. In at least some cases where the last segment of the route is an optional variable segment, the route fails to match when the optional segment is not present.

As a real use-case, I created the following route:

`/about-us/global/photos/{setid:integer}`

which is intended to match both when the final segment is present, and when it is absent. The "Require all variables?" option is turned off. In practice, the route matches only when the optional variable is present, and fails to match when it is absent. So:

`/about-us/global/photos/123456` (matches)
`/about-us/global/photos` (doesn't match, but should)

An informal workaround for this situation has been in circulation for some years, and involves using a REGEX to define the previous segment, like this:

`/about-us/global/{seg3:regex[(photos)]}/{setid:integer}`

This modified route does match the URLs both with and without the optional last segment.

See the following discussion from several years ago on the official EE forums, which describes the REGEX hack, and suggests that it was provided by Ellis Labs as a workaround:

https://expressionengine.com/forums/topic/250610/problem-with-pagination-and-template-routes

The issue seems to persist in EE5.4.1. It's possible that it's fixed in EE6, but I haven't seen anything in the change logs which would be relevant. It would be nice if this glitch could finally be resolved.

intoeetive:
Resolved #924 where template routing did not work properly with optional segments

### DIFF
--- a/system/ee/legacy/libraries/Template_router.php
+++ b/system/ee/legacy/libraries/Template_router.php
@@ -33,7 +33,8 @@ class EE_Template_Router extends CI_Router
      */
     public function match($uri)
     {
-        $request = $uri->uri_string;
+        // Make sure we have a trailing slash so segments parse correctly
+        $request = trim($uri->uri_string, '/ ') . '/';
 
         // First check if we have a bare match
         if (! empty($this->end_points[$request])) {
--- a/system/ee/legacy/libraries/template_router/Route.php
+++ b/system/ee/legacy/libraries/template_router/Route.php
@@ -100,7 +100,7 @@ class EE_Route
         $index = 0;
 
         foreach ($this->segments as $segment) {
-            $regex = $segment->regex();
+            $regex = $segment->validator();
 
             $add_question_mark = (! $this->required);
 
@@ -108,15 +108,18 @@ class EE_Route
                 $add_question_mark = false;
             }
 
-            if ($index < count($this->segments) - 1) {
-                $regex .= '\/';
-            }
-
             if ($add_question_mark) {
                 $regex .= '?';
             }
 
-            $url[] = $regex;
+            if ($index < count($this->segments) - 1) {
+                $regex .= '\/';
+                if ($add_question_mark) {
+                    $regex .= '?';
+                }
+            }
+
+            $url[] = '(' . $regex . ')';
             $index++;
         }
 


bryannielsen (commented):
@intoeetive I haven't tested this but looking at the code would it be better to do `rtrim` and just focus on the trailing slash like your comment says?

adster101 (approved):
Tested locally on 7.4.11 and seems to work. My template route is:

`/(uk|london|manchester|bristol|brighton)/search/{page:pagination}`

When I apply the patch urls with optional segments are working as expected:

http://localhost:3000/london/search
http://localhost:3000/london/search/P20

whereas without the patch http://localhost:3000/london/search resolves to a 404.

adster101:
On further testing it seems as though this fix doesn't fully work. It seems to work okay as outlined above but with a more complex route it seems not to work. 

For example I have this template route:

/(uk|london|manchester|bristol|brighton)/search/{segment_3:alpha_dash}/{page:pagination}

With the fix applied the following urls still give a 404 not found error

/london/search
/manchester/search
/bristol/search

intoeetive:
@adster101 thank you for reporting this. Pushed the change that would hopefully fix the issue (you would need to re-save the routes to make it work)

@bryannielsen need a good review here as well as some testing. The regex wasn't build correctly to consider multiple optional segments - previously, the slash was marked as optional, but the segment stayed required. I also tried to make it avoid extra unnecessary group nesting.

adster101:
Further testing confirms the patch posted by @intoeetive DOES NOT resolve this issue. 

I've updated my site to 7.5.22 and patched Route.php with pull request 4477 but I still get a 404 not found error using the following template route with segments required off and after resaving the template routes:

/(uk|london|manchester|bristol|brighton)/search/{page:pagination}

The following urls give 404 errors:

/london/search
/brighton/search
...

The following urls work:

/london/search/P0
/brighton/search/P0
...

adster101:
Resorted to using the template route:

/(uk|london|manchester|bristol|brighton)/{segment_2:regex[(search)]}/{page:pagination}

which works around the issue.

PR #4480, opened 2024-09-18
Improved #4403; Added setting for default value for the Entry Manager > 'Search Titles Only' checkbox

Yulyaswan:
Improved #4403; Added setting for default value for the Entry Manager > 'Search Titles Only' checkbox
Fixed #4403


themes/ee/asset/javascript/src/cp/publish/entry-list.js
```diff
@@ -143,13 +143,29 @@ $(document).ready(function () {
 		}, 1000)
 	});
 
+
+	// set coolie to remember last user choose about
```
nep: typo: cookie

PR #4484, opened 2024-09-21
Ability to exclude offline sites from `exp:channel:entries` output; #2895

intoeetive:
Ability to exclude offline sites from `exp:channel:entries` output; #2895

User Guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/889


robinsowell (approved):
Worked for me.  

I will note, in general, setting it offline means superadmin still see content, so I'm thinking a setting/param that does that would make sense.  But since nobody has requested it, I don't inherently have an issue with adding a param that locks it down entirely.

PR #4487, opened 2024-09-22
Add buttons for template routes, variables and partials at the top of Template Manager

intoeetive:
I find it very inconvenient that every time I need to access template variables or routes I need to scroll all way down (I have a lot of template groups)

This PR is adding shortcut buttons for those at the top
image


robinsowell (dismissed):
Well this makes me incredibly happy!  Worked perfect, looked good.  Think it's breaking a test, but I suspect it's due to the display change.

Approving, but one of the guys will need to look at the test.

TomJaeger:
We need to consider the UI / UX on this further before we merge.

PR #4488, opened 2024-09-23
Improved #4292; Saved to the localStorage the path to the previously used folder  when clicking on 'Choosing Existing' button for file upload field

Yulyaswan:
Fixed #4292;


Yulyaswan:
before 

https://github.com/user-attachments/assets/dd168860-97bf-4ef6-8656-b554bc642515

after

https://github.com/user-attachments/assets/9127ae72-fe06-4216-850a-2835b1a44093

PR #4497, opened 2024-09-27
Added Anti-XSS as ExpressionEngine security library

intoeetive:
This PR is adding the Anti-XSS library that was originating from CodeIgniter back into EE as composer dependency.

Because from our current Security/XSS library we use 2 public methods, I added wrappers for those, so we don't need to change existing code.

Because we had an addition to strip `console.log` and Anti-XSS wasn't flexible enough to configure that, I had to make a fork and add a method in Anti-XSS, so currently we're including our fork. Once the PR for that change will be approved by maintainer , we can switch back to including original repo


intoeetive:
It looks like AntiXSS is missing the functionality of our `remove_invisible_characters` plus a couple of other things. Although it does other things. We probably need to look into combining the functionality

PR #4510, opened 2024-10-06
Resolved #4504 where POSTed Channel Form values were sometimes not populated correctly in case of validation error

### LINKED ISSUE #4504: Using url_title parameter with Channel Form does not persist form data with inline error handling
When using Channel Form, I've encountered an issue when using `error_handling="inline"` and `url_title=""` where after form submission with a field validation error, the posted data is not returned back to the form as and old value.

If I use `entry_id=""` instead, the form submits correctly, the validation error appears, and the old value remains in the input field.

**How To Reproduce**
Steps to reproduce the behaviour:

1. Create a channel with a custom text field
2. In a template, use the template tag below and set the rules only so that the form validation is triggered and the form doesn't actually succeed and submit the changes.
3. Complete the form to force the validation error (in my example, I tested it using the required and numeric validations and then submitted text only)
4. Click submit
5. The form error shows, but the field value is blank instead of the text entered

**Using url_title _does not_ work**
```
{exp:channel:form
    url_title="playground"
    channel="sandbox"
    error_handling="inline"
    rules:basic_input="required|numeric"}
        
    <input type="text" name="basic_input" value="{basic_input}" />
        
    {if error:basic_input}
        {error:basic_input}
    {/if}
        
    <button type="submit">Submit</button>
{/exp:channel:form}
```

**Using entry_id _does_ work**
```
{exp:channel:form
    entry_id="1"
    channel="sandbox"
    error_handling="inline"
    rules:basic_input="required|numeric"}
        
    <input type="text" name="basic_input" value="{basic_input}" />
        
    {if error:basic_input}
        {error:basic_input}
    {/if}
        
    <button type="submit">Submit</button>
{/exp:channel:form}
```

**Environment Details:**
 - EE Version 7.4.11
 - PHP Version 8.2.15
 - MySQL Version 8.0.27
 - OS: Mac OS Sonoma
 - Web Server: nginx

intoeetive:
Resolved #4504 where POSTed Channel Form values were sometimes not populated correctly in case of validation error

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -497,9 +497,16 @@ class Channel_form_lib
                         $checkbox_fields[] = $key;
                         $this->parse_variables[$key] = ($this->entry($name) == 'y') ? 'checked="checked"' : '';
                     } elseif (property_exists($this->entry, $name) or $this->entry->hasCustomField($name)) {
-                        $this->parse_variables[$key] = $this->encode_ee_tags(
-                            form_prep($this->entry($name), $name)
-                        );
+                        // override with POST, if there was validation error
+                        if (isset($_POST[$name])) {
+                            $this->parse_variables[$key] = $this->encode_ee_tags(
+                                form_prep(ee()->input->post($name, true), $name)
+                            );
+                        } else {
+                            $this->parse_variables[$key] = $this->encode_ee_tags(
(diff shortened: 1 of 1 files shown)


nerdgency:
@intoeetive this solution works! I was wondering for a shorter version if you would consider the following:

```php
$this->parse_variables[$key] = $this->encode_ee_tags(
    form_prep(
        ee('Request')->post($key, $this->entry($name)),
        $name
    )
);

TomJaeger (approved):

matthewjohns0n (approved):

PR #4520, opened 2024-10-10
Fix htmlentities(): Handle array case in Filter.php

### LINKED ISSUE #4525: Selecting specific date in filter shows error when paginating
Issue:

When navigating the Developer Logs section, if a specific date is selected from the date filter dropdown and a pagination link is clicked, the following error occurs:

```
TypeError Caught
htmlentities(): Argument #1 ($string) must be of type string, array given
```

Location:
ee/ExpressionEngine/Service/Filter/Filter.php:103

Cause:
The value() method in Filter.php is passing an array to htmlentities(), which expects a string.

karlosb:
Fixes https://github.com/ExpressionEngine/ExpressionEngine/issues/4525

Issue:

When navigating the Developer Logs section, if a specific date is selected from the date filter dropdown and a pagination link is clicked, the following error occurs:

```
TypeError Caught
htmlentities(): Argument #1 ($string) must be of type string, array given
```

Location:
ee/ExpressionEngine/Service/Filter/Filter.php:103

Cause:
The value() method in Filter.php is passing an array to htmlentities(), which expects a string.

Fix:
Updated the value() method in Filter.php to handle cases where the value is an array by converting it to a string before passing it to htmlentities().

### DIFF
--- a/system/ee/ExpressionEngine/Service/Filter/Filter.php
+++ b/system/ee/ExpressionEngine/Service/Filter/Filter.php
@@ -93,14 +93,25 @@ abstract class Filter
         if (isset($this->selected_value)) {
             return $this->selected_value;
         }
-
+    
         $value = $this->derivedValue();
-
+    
         if (! $this->has_custom_value) {
             $value = $this->isValid() ? $value : null;
         }
-
-        return is_null($value) ? null : htmlentities($value, ENT_NOQUOTES, 'UTF-8');
+    
+        // Check if $value is an array, and handle it appropriately
+        if (is_null($value)) {
+            return null;
+        }
+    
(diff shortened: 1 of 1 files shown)


intoeetive:
@karlosb thank you for the contribution! We might use another solution for this, so creating a separate issue: https://github.com/ExpressionEngine/ExpressionEngine/issues/4525

karlosb:
Hi @intoeetive many thanks for the feedback on this. I must admit I ran further tests around the fix but admittedly I'm not 100% sure it's working as initially thought. So look forward to seeing your alternative solution on this. 🙂👍

intoeetive:
https://github.com/ExpressionEngine/ExpressionEngine/pull/4526

PR #4521, opened 2024-10-10
Resolve #4512 where Issue with multiple MiniGridInput

### LINKED ISSUE #4512: Issue with multiple MiniGridInput
We currently use MiniGridInput  ($grid = ee('CP/MiniGridInput', array('field_name' => 'options'));) in several spots natively and I just used it on an add-on.  It outputs a handy display.  You can set a min/max number of rows, similar to a full grid.

However, if you have more than 1 on a page, the min/max setting only works once.  In other words, it's global.  @matthewjohns0n figured out what the problem is and suggested @Yulyaswan could take a look.

Yulyaswan:
Resolve #4512 where Issue with multiple MiniGridInput


robinsowell (approved):
Boom!  Totally fixed it in my test and didn't see any issues.

Thanks!

PR #4527, opened 2024-10-13
Support PHP 8.4

intoeetive:
Adding PHP 8.4 support, and some initial 8.5 support based on https://php.watch/versions/8.5

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
           - name: Setup PHP
             uses: shivammathur/setup-php@v2
             with:
-                php-version: '8.1'
+                php-version: '8.2'
                 extensions: dom, curl, sqlite, libxml, mbstring, zip, pcntl, pdo, mysql, bcmath, soap, intl, gd, exif, iconv, imagick
                 coverage: none
                 ini-values: error_log=/home/runner/php_errors.log, memory_limit=128M
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -101,9 +101,9 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [7.4, 8.0, 8.1, 8.2, 8.3]
+                php: [7.4, 8.0, 8.4]
                 os: [ubuntu-latest]
-                containers: [0, 1, 2, 3, 4, 5]
+                containers: [0, 1, 2, 3, 4, 5, 6, 7, 8]
 
         name: Cypress Tests, PHP${{ matrix.php }} - ${{ matrix.os }} ${{ matrix.containers }}
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,7 @@ on:
       - '*.dev'
       - 'release/**'
       - '!release/**-alpha'
+  workflow_dispatch:
 
 jobs:
 
--- a/system/ee/ExpressionEngine/Boot/boot.php
+++ b/system/ee/ExpressionEngine/Boot/boot.php
@@ -61,7 +61,7 @@
  */
     try {
         if (file_exists(SYSPATH .'../.env.php')) {
-            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createImmutable(SYSPATH .'../', '.env.php');
+            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createUnsafeImmutable(SYSPATH .'../', '.env.php');
             $dotenv->load();
             // force the installer/updater?
             defined('INSTALL_MODE') || define('INSTALL_MODE', getenv('EE_INSTALL_MODE') === 'TRUE');
--- a/system/ee/ExpressionEngine/Boot/boot.php
+++ b/system/ee/ExpressionEngine/Boot/boot.php
@@ -61,10 +61,10 @@
  */
     try {
         if (file_exists(SYSPATH .'../.env.php')) {
-            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createUnsafeImmutable(SYSPATH .'../', '.env.php');
+            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createImmutable(SYSPATH .'../', '.env.php');
             $dotenv->load();
             // force the installer/updater?
-            defined('INSTALL_MODE') || define('INSTALL_MODE', getenv('EE_INSTALL_MODE') === 'TRUE');
+            defined('INSTALL_MODE') || define('INSTALL_MODE', ($_ENV['EE_INSTALL_MODE'] ?? false) === 'TRUE');
         }
     } catch (\Exception $e) {
 
--- a/system/ee/legacy/core/Exceptions.php
+++ b/system/ee/legacy/core/Exceptions.php
@@ -165,7 +165,7 @@ class EE_Exceptions
         }
 
         // if this is CP request and they are logged in, throw special kind of Exception
-        if (defined('REQ') && constant('REQ') == 'CP' && isset(ee()->session) && ee()->session->userdata('admin_sess') != 0) {
+        if (defined('REQ') && constant('REQ') == 'CP' && ee() && isset(ee()->session) && ee()->session->userdata('admin_sess') != 0) {
             throw new \ExpressionEngine\Error\CPException($message, $status_code);
         }
 
--- a/system/ee/legacy/core/Exceptions.php
+++ b/system/ee/legacy/core/Exceptions.php
@@ -350,6 +350,9 @@ class EE_Exceptions
      */
     private function lookupSeverity($severity)
     {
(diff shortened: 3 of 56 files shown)


system/ee/legacy/core/Exceptions.php
```diff
@@ -165,7 +165,7 @@ public function show_error($heading, $message, $template = 'error_general', $sta
         }
 
         // if this is CP request and they are logged in, throw special kind of Exception
-        if (defined('REQ') && constant('REQ') == 'CP' && isset(ee()->session) && ee()->session->userdata('admin_sess') != 0) {
+        if (defined('REQ') && constant('REQ') == 'CP' && ee() && isset(ee()->session) && ee()->session->userdata('admin_sess') != 0) {
```
TomJaeger, line 168: Wondering why this was needed? @bryannielsen

tests/cypress/support/sql/database_7.0.0.sql
```diff
@@ -623,7 +623,7 @@ INSERT INTO `exp_global_variables` (`variable_id`, `site_id`, `variable_name`, `
 	(6, 1, 'html_close', '<p id="elapsed_time">Elapsed time: <em>{elapsed_time}</em></p><p id="memory_usage">Memory usage: <em>{memory_usage}</em></p></body>\n</html>', 0),
 	(7, 1, 'html_head', '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">\n<html xmlns="http://www.w3.org/1999/xhtml">\n<head>\n<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />\n', 0),
 	(8, 1, 'html_head_end', '</head>\n', 0),
-	(9, 1, 'js', '<!-- JS -->\n<script src="http://code.jquery.com/jquery-1.12.1.min.js" type="text/javascript"></script>\n<script src="{site_url}themes/site/default/js/onload.js" type="text/javascript"></script>', 0),
+	(9, 1, 'js', '<!-- JS -->\n<script src="http://code.jquery.com/jquery-1.12.1.min.js" type="text/javascript"></script>', 0),
```
TomJaeger, line 626: was this dropped on purpose? @bryannielsen

TomJaeger (commented):
Few questions in this for you @bryannielsen

tests/cypress/support/sql/database_7.0.0.sql
bryannielsen: Yes that's just the texting database fixture and that `js/onload.js` file isn't in the theme so it was causing cypress errors

system/ee/legacy/core/Exceptions.php
bryannielsen: In testing I believe there were certain circumstances where `ee()` was returning null, it may have been during the install routine before EE was fully setup

.github/workflows/tests.yml
```diff
@@ -100,9 +101,9 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [7.4, 8.0, 8.1, 8.2, 8.3]
```
matthewjohns0n, line 103: Are we intentionally not testing 8.1-8.3?

system/ee/ExpressionEngine/Boot/boot.php
```diff
@@ -61,7 +61,7 @@
  */
     try {
         if (file_exists(SYSPATH .'../.env.php')) {
-            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createImmutable(SYSPATH .'../', '.env.php');
+            $dotenv = ExpressionEngine\Dependency\Dotenv\Dotenv::createUnsafeImmutable(SYSPATH .'../', '.env.php');
```
matthewjohns0n: Can someone explain why we are making this change?

system/ee/legacy/libraries/Xmlrpc.php
```diff
@@ -620,10 +620,16 @@ public function parseResponse($fp)
...
         $this->xh[$parser_name]['isf_reason'] = 0;
 
-        xml_set_object($parser, $this);
         xml_parser_set_option($parser, XML_OPTION_CASE_FOLDING, true);
-        xml_set_element_handler($parser, 'open_tag', 'closing_tag');
-        xml_set_character_data_handler($parser, 'character_data');
+        xml_set_character_data_handler($parser, function($parser, $cdata) {
+            return $this->character_data($parser, $cdata);
+        });
+        xml_set_element_handler($parser, function($parser, $tag, $attributes) {
```
matthewjohns0n, line 627: This will require PHP 8.0 to work; I dont think we can use closures here.

system/ee/legacy/libraries/Xmlrpcs.php
```diff
@@ -159,10 +159,15 @@ public function parseRequest($data = '')
...
         $parser_object->xh[$parser_name]['method'] = '';
 
-        xml_set_object($parser, $parser_object);
         xml_parser_set_option($parser, XML_OPTION_CASE_FOLDING, true);
-        xml_set_element_handler($parser, 'open_tag', 'closing_tag');
-        xml_set_character_data_handler($parser, 'character_data');
+        xml_set_character_data_handler($parser, function($parser, $cdata) {
+            return $this->character_data($parser, $cdata);
+        });
+        xml_set_element_handler($parser, function($parser, $tag, $attributes) {
```
matthewjohns0n, line 166: This will require PHP 8.0 to work; I dont think we can use closures here.

system/ee/legacy/libraries/simplepie/SimplePie/Parser.php
```diff
@@ -155,9 +155,14 @@ public function parse(&$data, $encoding, $url = '')
             $xml = xml_parser_create_ns($this->encoding, $this->separator);
             xml_parser_set_option($xml, XML_OPTION_SKIP_WHITE, 1);
             xml_parser_set_option($xml, XML_OPTION_CASE_FOLDING, 0);
-            xml_set_object($xml, $this);
-            xml_set_character_data_handler($xml, 'cdata');
-            xml_set_element_handler($xml, 'tag_open', 'tag_close');
+            xml_set_character_data_handler($xml, function($parser, $cdata) {
+                return $this->cdata($parser, $cdata);
+            });
+            xml_set_element_handler($xml, function($parser, $tag, $attributes) {
```
matthewjohns0n, line 161: Same here - this code is PHP 8.0 and up.

system/ee/legacy/libraries/Xmlrpc.php
bryannielsen: Where are you seeing a PHP 8.0 requirement?  The anonymous function shouldn't be a problem, I think support was added in PHP 5 and we use them all over.  It also seems like xml_set_element_handler has always accepted a callable as the second parameter so we should be set there too.  Are you seeing an actual error on PHP < 8.0 or a lint error?

.github/workflows/tests.yml
bryannielsen: Yea we're figuring 7.4, 8.0, and 8.4 will be enough coverage and the reduced time to complete a full suite would be a benefit.  We have the ability on the minimal-tests now to run them on demand and specify a version though if we have reason to believe a specific 8.x minor version is having a problem

system/ee/ExpressionEngine/Boot/boot.php
bryannielsen: Good catch, it came from this commit (https://github.com/ExpressionEngine/ExpressionEngine/commit/2f4df0be4384e5b7809b8a3aa101216066ffee9e) but I don't know what issue Yuri was trying to fix.  It sounds like `createUnsafeImmutable` makes ENV vars available for native `getenv()` and `putenv()` calls.  So maybe reverting this and changing the line below from `getenv('EE_INSTALL_MODE')` to `$_ENV['EE_INSTALL_MODE'] ?? false` would be a good solution.

All the other `getenv()` calls I see in the core are for PHP env vars not user-defined variables

.github/workflows/tests.yml
matthewjohns0n: Sounds good - that makes sense

system/ee/legacy/libraries/Xmlrpc.php
matthewjohns0n: This was a lint error. If we're confident this is fine in PHP 7, I'll mark as resolved.

system/ee/ExpressionEngine/Boot/boot.php
matthewjohns0n: Yeah, it just seemed like a big change for PHP 8.4 compatibility. Whatever option you think is better @bryannielsen

TomJaeger: I believe this has been reverted and we can mark this as resolved?  @bryannielsen and @matthewjohns0n

matthewjohns0n (approved):

TomJaeger (approved):

PR #4537, opened 2024-10-17
Fixed #4535 where  if  have an odd number of toggle fields in a channel form, they become unresponsive

### LINKED ISSUE #4535: EE 7.5 if I have more than 24 toggle fields in a channel form, they become unresponsive
SUPER WEIRD.

Have a channel, have 25 toggle fields in the channel, no other fields.  Works fine in the cp.  If I put them in a channel form- in my test, I was using the custom field loop- they become untoggleable.  

I go to the form, I click a toggle, it does nothing.  There is no error, it just doesn't toggle.  If I remove 1 of the toggle fields so now there are 24 assigned to the channel, they toggle.

So weird, think it has to be JS.  I'll get you a channel set so you don't have to make 25 toggle fields because that is ZERO FUN.

Yulyaswan:
Fixed #4535 where  if  have an odd number of toggle fields in a channel form, they become unresponsive


robinsowell (approved):
Fixed it!

PR #4548, opened 2024-10-22
Add additional filter to file upload

matthewjohns0n:


robinsowell (dismissed):
So I couldn't test the part where I embed php in an image, but everything else I did tested out fine.

robinsowell (approved):
All good in my tests

PR #4557, opened 2024-10-24
Set the socket to non-blocking before checking the socket response

matthewjohns0n:
Fixes a bug with certain socket stream responses in moblog

### DIFF
--- a/system/ee/ExpressionEngine/Addons/moblog/mod.moblog.php
+++ b/system/ee/ExpressionEngine/Addons/moblog/mod.moblog.php
@@ -223,7 +223,7 @@ class Moblog
         stream_set_blocking($this->fp, false);
 
         // Set a timeout for the socket
-        stream_set_timeout($this->fp, 10); // 5 seconds timeout
+        stream_set_timeout($this->fp, 10); // 10 seconds timeout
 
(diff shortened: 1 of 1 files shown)


system/ee/ExpressionEngine/Addons/moblog/mod.moblog.php
```diff
@@ -219,7 +219,18 @@ public function check_pop_moblog()
             return false;
         }
 
-        if (strncasecmp(fgets($this->fp, 1024), '+OK', 3) != 0) {
+        // Set the socket to non-blocking mode
+        stream_set_blocking($this->fp, false);
+
+        // Set a timeout for the socket
+        stream_set_timeout($this->fp, 10); // 5 seconds timeout
```
bryannielsen: Should this be 10 seconds or 5 seconds.  The code and comment don't line up

matthewjohns0n: I think I set it to 5 seconds and then changed it to 10 seconds. I don't think it matters a huge amount, but I'm updating the comment to say 10 seconds.

bryannielsen (approved):

PR #4570, opened 2024-11-01
Resolved #4569 where older versions of PHP might throw an error during Pro access checks

### LINKED ISSUE #4569: PHP error in pro access service when checking license management permissions
After successful upgrade to EE 7.5.2, there was a parse error in the cp upon login about extra comma in ee/ExpressionEngine/Addons/pro/Service/Access/Access.php (line 203).  Looks like the trailing comma in the hasany check may cause issues.

I couldn't replicate, despite my best efforts.  But... it definitely happened.

ETA- happened on php 7.2, which is probably why I failed to replicate.

bryannielsen:
Resolved #4569 where older versions of PHP might throw an error during Pro access checks

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro/Service/Access/Access.php
+++ b/system/ee/ExpressionEngine/Addons/pro/Service/Access/Access.php
@@ -200,7 +200,7 @@ class Access
                 'delete_channel_fields',
                 'create_members',
                 'edit_members',
-                'delete_members',
+                'delete_members'
             );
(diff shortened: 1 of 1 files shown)


matthewjohns0n (approved):

matthewjohns0n:
Beautiful. You have a way with code Bryan.

bryannielsen:
Thank you @matthewjohns0n, it feels good to have your approval!

PR #4591, opened 2024-11-13
Resolved #4589 where there was an error removing a field from a Fluid field that was also used in a field group

### LINKED ISSUE #4589: Removing a field from a fluid field group deletes the content if that same field is used in another field group assign to the the same fluid field
Removing a 'Custom Field' from a fluid field deletes the content if that same field is used in another 'Custom field groups' assigned to the the same fluid field.

bryannielsen:


robinsowell (approved):
Fixed it for me.  I did have to make a new fluid field to test with, since testing the original borked my db.  But new one should have been a good approximation of the initial test, and all was fine.

PR #4604, opened 2024-11-19
Fix copy field with cp session type as session only

### LINKED ISSUE #4588: Copy field code not working when session id is in the cp url
EE 7.5.3

On the field manager and in the entry edit, if you're using sessions in the cp, the copy field doesn't work.

I tried it- it triggers a url like https://75gen:8890/admin.php?/cp/design/copy/fields&S=1053ee30af84cfc809af3cdaea3f78ce/5 which 404's.

URL should be https://75gen:8890/admin.php?/cp/design/copy/fields/5&S=1053ee30af84cfc809af3cdaea3f78ce/ - basically we just need to switch it up and it should be fine.

matthewjohns0n:
Fixes #4588


robinsowell (approved):
Worked for me!

PR #4610, opened 2024-11-20
Gets stable chrome version from release schedule in cypress test

matthewjohns0n:
This will get a Stable Chrome version that is 2 releases back from latest, and install it 
It gets the version from the Official Chromium Dash API
Get a couple releases back ensures we do not run into the issue where not all runners have the latest version


bryannielsen:
@matthewjohns0n looks like all of the tests are failing here

bryannielsen (approved):

PR #4611, opened 2024-11-21
Improve member tag support for multiple template engines

bryannielsen:


robinsowell (approved):
Tested the form and the profiles and pagination- well, I paged something.  Couldn't spot an issue.

PR #4615, opened 2024-11-22
Resolved #4566 where file_name was treated as binary when performing SQL queries

### LINKED ISSUE #4566: Performance issue with the exp_files table doing a "BINARY" query.
This is a very large site with over 600,000 rows in the exp_files table.. So, these performance things probably aren't noticed by most people/most sites.. But, at this scale, just opening an entry to edit in the CP or saving it if there are any file fileds takes much longer than needed.

The issues seems to be related to the (long join) query that EE does with the exp_files.file_name in the where clause..  EE does a BINARY type on that query. Example:

`WHERE
    (
        `File_files`.`model_type` = 'File' AND BINARY `File_files`.`file_name` = 'Day_two_-_1066.jpg' AND `File_files`.`upload_location_id` = '6' AND `File_files`.`site_id` = 1
    )
LIMIT 1`

That "BINARY" in the SQL where clause basically sets the comparison on the file_name to be case-sensitive.    We did some testing, and ulimately, we changed the exp_files.file_name column from varchar 255 to varbinary 2040. (db string type fields: varchar, text, etc.) have a storage size in characters.. varbinary uses bytes.. So, I converted 255 characters to 2040 bytes, to ensure no existing data was truncated). 

We also indexed that column since EE is using that field in "where" clauses of queries.  Then we edited the following file for this:
/admin/ee/ExpressionEngine/Model/File/FileSystemEntity.php
commented out line 38 which sets the "file_name" column for the binary_comparison to remove the addition of the "BINARY" in there and we are seeing SUBSTANTIAL performance improvement.

A channel entry with a single file first party EE file grid field with 10 images..
Before changing:
opening to edit took about 6 seconds
saving took about 10 seconds.
After changes:
opening to edit to about 2 seconds
saving took about 3 seconds.

The BINARY sql statement forces a case-sensitive comparison.. BUT, for a file name column in the DB, there should never be a case-insensitive comparison needed to be made. AND since (at least on unix/linux) the file system is case-sensitive,  it makes sense to just make that db column a varbinary..  then index it.. Then you can do just a standard SQL query/comparison, without forcing the BINARY type... since the DB already takes care of that..

To follow up with this.. After testing through our dev site.. We implemented these changes on our live site.  A very robust server with 16 processors and 32 GB RAM..

A single channel entry with only 2 native EE grid file fields in it.. 
1 has 2 images, the other as 36..  

Before these changes it took 24 seconds to just load the entry in the CP.. Now loads in 1.2 seconds..  It took 36 seconds to save the entry..  now saves in less than 3 seconds..

I dont' beleive the performance issues to be related to file size.. we tested with even small images do the same..  It's related to the exp_files table size, and how long the query takes to execute.

All this time, I thought it was just Ansel fields that were the bog-down in the Control Panel.. BUt, we've removed Ansel.. So, these are all native EE file grid fields and still saw the performance problems... which led us to this deep-dive testing and, seemingly, straight-forward solution.

intoeetive:
Resolved #4566 where file_name was treated as binary when performing SQL queries

### DIFF
--- a/system/ee/ExpressionEngine/Model/File/FileSystemEntity.php
+++ b/system/ee/ExpressionEngine/Model/File/FileSystemEntity.php
@@ -35,7 +35,7 @@ class FileSystemEntity extends ContentModel
         'beforeSave'
     );
     protected static $_binary_comparisons = array(
-        'file_name'
+        //'file_name'
     );
 
     protected static $_hook_id = 'file';


TomJaeger:
After discussion on this one.  It needs further testing to confirm we're not introducing a degradation for case sensitive file systems.  We believe this could be the case currently.

intoeetive:
@TomJaeger that's a valid point, however I think the NEW behaviour is the desired one. If I have files named "OneFirst" and "oneSecond", I want both to be found if I search for "one" (so ignoring the case)

Furthermore, if we really need to make the search strict (which I highly doubt), a lot more efficient way to do that would be to use one of "_cs" (case sensitive) collations on MySQL table cell

bryannielsen:
I don't think we can disregard the case-sensitivity across the board.  If you search the codebase for `filter('file_name'` you'll see areas where it's being used in conjunction with `->first()` to find a specific file.  So this change could lead to unexpected results if two files have the same case-insensitive representation.

Changing the `file_name` column collation to be case-sensitive would be the best move I think but that seems like a MySQL 8+ only solution (using `utf8mb4_0900_as_cs`).  MySQL 5.7 has limited support for case-sensitive collations and `latin1_general_cs` wouldn't support utf8 characters which would be a problem.

I think our second best solution is to use a `_bin` collation which would be supported by MySQL 5.7+.
I did a little testing with a local database that was using the utf8 character set but this approach would work fine with `utf8mb4` and `utf8mbf_bin` as well.

```
ALTER TABLE exp_files
MODIFY COLUMN file_name varchar(255) CHARACTER SET utf8 COLLATE utf8_bin
```

As long as that sounds reasonable I think we should do this in an update migration and drop the binary comparison like @intoeetive has done in this PR.  And perhaps at the same time we should add an index on `file_name` which appears to be missing and would improve performance on large data sets.

intoeetive:
I'll let you guys decide and handle that, however it's kinda strange to me to have binary collation on column that's clearly text.  Unless we only do it as temporary solution for old systems that don't support _cs

bryannielsen:
After looking into this more I think the binary collation wouldn't be a great solution because sorting by byte values would be unnatural i.e. `A,B,a,b` instead of `A,a,B,b`.  This sorting issue would also be present with the solution proposed in the initial issue #4566 where the column type was changed to VARBINARY.

So I think the best solution is to use a case sensitive `_cs` collation in MySQL 8+ and drop the binary cast.

A solution for supporting MySQL 5.7+ would be to create an additional column for `file_name` (maybe `_file_name_bin`) that would use the binary collation or a binary column type and would be strictly for filtering data where the regular character collation field would be used for sorting.  I really don't like this idea but I'm not sure how else we can improve this.

intoeetive:
What if we use `_cs` collation for the MySQL versions that support it, and for older versions just use case-insensitive?

(I also still think this needs to be case-insensitive in call cases though)

bryannielsen:
We're going to go with #5051 and take the original approach from this PR but make it possible to opt-in to the old behavior of using the BINARY cast to do a case-sensitive comparison.

PR #4621, opened 2024-11-27
Resolved an issue where the default template engine was not used when copying field codes

bryannielsen:


robinsowell (dismissed):
Tested in straight ee- copy group, field and channel.  All worked.

matthewjohns0n (approved):

PR #4635, opened 2024-12-17
Improved #4581 where filter drop-down controls should show as much of the filter value as possible

Yulyaswan:
Improved #4581 where filter drop-down controls should show as much of the filter value as possible


Yulyaswan:
before
Screenshot 2025-02-05 at 10 51 51

after 
Screenshot 2025-02-05 at 10 52 47

robinsowell (approved):
Nice- I hadn't run into the problem before, but this is a vast improvement.

Screenshot 2025-03-03 at 3 25 11 PM

PR #4638, opened 2024-12-19
Make parseLayoutVariables public

litzinger:
Making this public will let me greatly improve how Speedy parses cached templates that use layout variables.

$this->layout_vars is already public, along with most other methods to parse globals, vars etc, and even some methods that are prefixed with an underscore, which is usually indicative of a private method. Making this public doesn't feel like it would have adverse side effects?

### DIFF
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -691,10 +691,11 @@ class EE_Template
      * @param  array $layout_vars Layout variables to parser, 'variable_name' => 'content'
      * @return string The parsed template/string
      */
-    private function parseLayoutVariables($str, $layout_vars)
+    public function parseLayoutVariables($str, $layout_vars)
     {
         $this->log_item("Layout Variables:", $layout_vars);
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/Template.php
```diff
@@ -691,10 +691,11 @@ public function parse(&$str, $is_embed = false, $site_id = '', $is_layout = fals
      * @param  array $layout_vars Layout variables to parser, 'variable_name' => 'content'
      * @return string The parsed template/string
      */
-    private function parseLayoutVariables($str, $layout_vars)
+    public function parseLayoutVariables($str, $layout_vars)
     {
         $this->log_item("Layout Variables:", $layout_vars);
         $this->layout_conditionals = [];
+        $layout_conditionals = [];
```
litzinger, line 698: Line 790 attempts to assign a variable to the class property, but it's not created until line 725, or in my case not at all. Add this so it doesn't throw an error.

intoeetive (approved):

PR #4663, opened 2025-01-02
Resolved #4634 where Member Template Generator was missing some templates

### LINKED ISSUE #4634: Member generator doesn't create reset_password_form template
If you reset your password with the Member generatorated templates, it send you an email with a link to a reset-password-template, but it doesn't work because the template doesn't exist.  And it's not a matter of the link being wrong, we just don't have a template with the correct tag- which is ```{exp:member:reset_password_form}```.

Just need to add that template with that tag and it's good.

intoeetive:
Resolved #4634 where Member Template Generator was missing some templates


bryannielsen (approved):

robinsowell (approved):
We even had a couple I didn't know were missing!  Looks good.

PR #4664, opened 2025-01-03
Resolved #4658 where it was not possible to deselect entry category if multiple selection is not allowed

### LINKED ISSUE #4658: Radio Buttons in Categories cannot be de-selected
EE 7.4.11

When a Category is set to not allow multiple selections, and thus uses a radio button, you cannot unselect that button if you decide you want to leave that category blank for that entry.

Current workaround is to edit that Channel's Category to Allow Multiple Selections, return to entry and un-check checkbox, then return to category settings and toggle Allow Multiple Selections back to the off position.

intoeetive:
Resolved #4658 where it was not possible to deselect entry category if multiple selection is not allowed

Resolved the issue where all categories were selected for entry initially


robinsowell (approved):
Works for me- tested new and edit, with/without a cat required.  NICE!

PR #4665, opened 2025-01-03
Resolved #4576 where replace in Pro Search did not work with shared Grid fields

### LINKED ISSUE #4576: Pro search find and replace has an issue with grid fields
If you have a grid field that's searchable, you do a find/replace with pro search.  It replaces the content in the field id for the grid field, but it's not replacing it in the actual content stored in the individual grid columns.  So effectively, it doesn't replace it.

intoeetive:
Resolved #4576 where replace in Pro Search did not work with shared Grid fields


robinsowell (approved):
The fix worked for me, but I did find another bug in testing.

https://github.com/ExpressionEngine/ExpressionEngine/issues/4685

I'm treating them as separate.

PR #4668, opened 2025-01-05
Added `orderby` parameter to `exp:channel:categories` tag; #4340

intoeetive:
Added `orderby` parameter to `exp:channel:categories` tag; #4340

User Guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/919


matthewjohns0n (dismissed):

TomJaeger:
@robinsowell can you run a quick manual test on this one?

robinsowell (approved):
I could not break it.  Tried sorting on title and a custom select field.  Added in a show_empty param and specifying which cats.  Everything worked, no errors logged.

I did try adding an order param and it does work and isn't documented.  I did expect it to work- so yay.  I'll go stick it in the doc pr.  

But yep- totally worked.

PR #4671, opened 2025-01-05
Resolved #4657 where automatic URL titles were not generated for entries with non-ASCII titles

### LINKED ISSUE #4657: Cannot create an entry when `Enforce automatic URL title` is enabled
EE 7.5.6

When "Enforce automatic URL title" is enabled, users are unable to create entries with titles containing only non-Latin characters.

### Steps to Reproduce:

1. Enable `Enforce automatic URL title` in the ExpressionEngine settings.
2. Create a new entry with a title that uses only non-Latin characters (e.g., `ექსფრესსიონენგინე`).
3. Attempt to publish the entry.

`URL title` field doesn't populate automatically. It's required. It's readonly.
  
image

intoeetive:
Resolved #4657 where automatic URL titles were not generated for entries with non-ASCII titles

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
@@ -397,8 +397,12 @@ abstract class AbstractPublish extends CP_Controller
             $entry->markAsDirty();
         } else {
             if ($entry->isNew() && $entry->Channel->enforce_auto_url_title) {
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
Worked for me- I still kind of expected it to fill in something before I submitted.  But it passes the validation checks with nothing in the url title field and saved it and then had a working url and as Yuri said- if you want it to have an actual translation, add the config.

PR #4673, opened 2025-01-06
Resolved #4622 where applying `webp` modifier to images with same name but different extensions would produce same result

### LINKED ISSUE #4622: `:webp` modifier outputs same image for files with same name but different extensions
When using the `:webp` modifier in a grid, files with the same name but different extensions (e.g., `picture.jpg` and `picture.png`) are processed into the same output image. Despite the images being different, ExpressionEngine renders them as identical when the `:webp` modifier is applied.

## Steps to Reproduce:
- Upload two different images with the same filename but different extensions (e.g., testpic.jpg and testpic.png) in a grid.
- Apply the `:webp` modifier to both files.
- Observe that the output is the same for both images.

## Original pictures:
Screenshot 2024-11-28 122658

## Output:
Screenshot 2024-11-28 122706

intoeetive:
Resolved #4622 where applying `webp` modifier to images with same name but different extensions would produce same result

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -532,6 +532,7 @@ JSC;
         ee()->load->library('image_lib');
         $filename = ee()->image_lib->explode_name($data['fs_filename']);
         if ($function == 'webp') {
+            $filename['name'] = $filename['name'] . '_' . $filename['ext'];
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):

robinsowell (approved):
Fixed!  Though I didn't realize you couldn't slap an image modifier on the url variable inside an image tag pair.  But that's an FR for another day.

PR #4691, opened 2025-01-24
Improved #4094: In a Grid field, the handle and trash icons can now be placed as the first column instead of the last column.

Yulyaswan:
Improved #4094: In a Grid field, the handle and trash icons can now be placed as the first column instead of the last column.


Yulyaswan:
I am closing this PR because a new, improved solution to the request has been created #5203

PR #4694, opened 2025-01-27
Add before_version_delete hook

swierczek:
This allows devs to adjust the logic of which versions get deleted. For example, when using Publisher, it allows us to maintain x languages _per language_ per entry, instead of just x languages per entry.

User Guide Pull Request: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/935


bakin1999:
Thanks for this idea, looks good! Thanks for doing the docs part as well.

bakin1999 (approved):

PR #4702, opened 2025-02-06
Enhancement - Add a get_keys() function to SmartForge class

jcogs-design:
Smartforge allows for creation and deletion of keys, but only where you know the name of the key to be deleted.

This function adds the ability to obtain the output from the MySQL SHOW INDEX FROM command and presents the results as an associative array.

### DIFF
--- a/system/ee/legacy/libraries/Smartforge.php
+++ b/system/ee/legacy/libraries/Smartforge.php
@@ -386,6 +386,44 @@ class Smartforge
 
         return false;
     }
+
+    /**
+     * List Keys
+     *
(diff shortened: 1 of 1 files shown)


TomJaeger:
Thanks so much for your contribution! After reviewing it, we think it would be best to place this in mysqli_connection.php instead of smartforge. We'd also like to rename get_keys to get_indexes for clarity.

Would you be open to making these changes? If not, no worries—we're happy to take care of them on our end. Just let us know what works best for you!

jcogs-design:
Happy for you to make the changes you have suggested!

PR #4703, opened 2025-02-06
Decouple site query from log query when viewing CP logs

swierczek:
TL;DR - when using MySQL specifically (not MariaDB), viewing the Control Panel logs takes 1-2 minutes to load because this query takes so long and returns a lot of data. This PR drastically improves page speed because the `site_pages` data isn't returned for each row of the query. Additionally, the page queries now only fetch `site_pages` once instead of for every row of the logs.

---

When viewing the Control Panel logs (`/admin.php?/cp/logs/cp`), there is a query that is joining the `cp_log` table with the `sites` table, and then selecting all columns. As seen in the CMS debugger, it takes a long time and it returns an unnecessary amount of data:

image

```
SELECT
    CpLog_cp_log.id as CpLog__id,
    CpLog_cp_log.site_id as CpLog__site_id,
    CpLog_cp_log.member_id as CpLog__member_id,
    CpLog_cp_log.username as CpLog__username,
    CpLog_cp_log.ip_address as CpLog__ip_address,
    CpLog_cp_log.act_date as CpLog__act_date,
    CpLog_cp_log.action as CpLog__action,
    Site_sites.site_id as Site__site_id,
    Site_sites.site_label as Site__site_label,
    Site_sites.site_name as Site__site_name,
    Site_sites.site_description as Site__site_description,
    Site_sites.site_color as Site__site_color,
    Site_sites.site_bootstrap_checksums as Site__site_bootstrap_checksums,
    Site_sites.site_pages as Site__site_pages
FROM
    (`exp_cp_log` as CpLog_cp_log)
    LEFT JOIN `exp_sites` AS Site_sites ON `Site_sites`.`site_id` = `CpLog_cp_log`.`site_id`
ORDER BY `CpLog_cp_log`.`act_date` desc
LIMIT 25;
```

It took over a minute to load this page with ~15k rows in `cp_log`, and the `sites.site_pages` field was 300 KB. That meant that for every row returned, it was also returning 300 KB of data, which was the same data for every row.

image

Decoupling the `cp_log` query from the `sites` query (because we don't need to know that data per row, we just need all of the `sites` rows once) makes this page load nearly instantly, and only returns a small amount of data (from 14 MB to 42 KB).

image

image

The returned results are the same and still include the `site_label`.

image

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Logs/Cp.php
+++ b/system/ee/ExpressionEngine/Controller/Logs/Cp.php
@@ -42,7 +42,6 @@ class Cp extends Logs
 
         if ($search = ee()->input->get_post('filter_by_keyword')) {
             $logs->search(['action', 'username', 'ip_address'], $search);
-            $sites->search(['Site.site_label'], $search);
         }
 
         $filters = ee('CP/Filter')
--- a/system/ee/ExpressionEngine/Controller/Logs/Cp.php
+++ b/system/ee/ExpressionEngine/Controller/Logs/Cp.php
@@ -96,8 +96,6 @@ class Cp extends Logs
             ->offset($offset)
             ->all();
 
-        $sites = $sites->order('site_id')->all();
-
         $pagination = ee('CP/Pagination', $count)
             ->perPage($this->params['perpage'])
             ->currentPage($page)
@@ -105,7 +103,7 @@ class Cp extends Logs
 
         $vars = array(
             'logs' => $logs,
-            'sites' => $sites,
+            'sites' => $sites->all(),
             'pagination' => $pagination,
             'form_url' => $this->base_url->compile(),
         );
--- a/system/ee/ExpressionEngine/Controller/Logs/Cp.php
+++ b/system/ee/ExpressionEngine/Controller/Logs/Cp.php
@@ -37,10 +37,12 @@ class Cp extends Logs
         $this->base_url->path = 'logs/cp';
         ee()->view->cp_page_title = lang('view_cp_log');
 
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Controller/Logs/Cp.php
```diff
@@ -37,10 +37,11 @@ public function index()
         $this->base_url->path = 'logs/cp';
         ee()->view->cp_page_title = lang('view_cp_log');
 
-        $logs = ee('Model')->get('CpLog')->with('Site');
+        $sites = ee('Model')->get('Site');
+        $logs = ee('Model')->get('CpLog');
 
         if ($search = ee()->input->get_post('filter_by_keyword')) {
-            $logs->search(['action', 'username', 'ip_address', 'Site.site_label'], $search);
```
swierczek, line 43: IMO there's no benefit to keyword searching `Site.site_label` because it's already a separate dropdown field with MSM enabled:

image

And with MSM disabled, there's no benefit to searching on site_label because all results would be connected to the same site_id.

swierczek, line 43: There's no benefit to keyword searching `Site.site_label` because it's already a separate dropdown field with MSM enabled:

image

And with MSM disabled, there's no benefit to searching on site_label because all results would be connected to the same site_id.

TomJaeger:
Thank you for this contribution!

robinsowell (approved):
Nice!  Worked perfectly for me.

bakin1999 (approved):
Tested, Worked well for me.

PR #4709, opened 2025-02-13
Resolved an issue where numeric channel shortnames caused a js error in Pro Search settings

bryannielsen:


jaredchesebro:
I may be the only one affected but thank you!

robinsowell (approved):
I had the js error- dropped in new file- now there is no js error.

bakin1999:
nice, yes I saw the error before and this is fixing that.

bakin1999 (approved):

PR #4710, opened 2025-02-13
Add function to list indexes of nominated table

jcogs-design:
Tools exist within EE db drivers to add and remove indexes from tables, but only work if you know the name of the index(es) you want to work on.

This function makes it possible to get an array of the names of the index(es) already defined for a table, making manipulation of index(es) easier.

### DIFF
--- a/system/ee/legacy/database/drivers/mysqli/mysqli_connection.php
+++ b/system/ee/legacy/database/drivers/mysqli/mysqli_connection.php
@@ -226,6 +226,44 @@ class CI_DB_mysqli_connection
     {
         return isset($this->connection);
     }
+    
+    /**
+     * List Indexes
+     *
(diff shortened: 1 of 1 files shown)


bakin1999:
PHP might have warnings around $key being returned and not set here.

Depending, we might also need to have log be loaded before the fail conditions here.
ee()->load->library('logger');

Besides those, this looks good.
Thanks for the idea jcogs!

jcogs-design:
Thanks @bakin1999 - the $keys is a typo. Originally this used "keys" to be consistent with terminology used in SmartForge functions ... but @TomJaeger suggested it should be based on Indexes ... I changed the variable name accordingly so it would not be confusing at a later date ... but I guess I missed one 🤦🏻‍♂️

TomJaeger (approved):

PR #4715, opened 2025-02-17
Resolved #4713 where required fields in a field group were not checked by fluid on entry saves

### LINKED ISSUE #4713: Required fields are not enforced in fluid field group
ExpressionEngine 7.5.7

Required fields inside a Fluid Field Group are not marked as mandatory and are not validated. Entries can be saved with empty required fields, bypassing validation.

Image

bakin1999:
What's in this pull request?

The title of the pull request should look like the change log line, with reference to the corresponding issue number if available. For example:
  - Resolved #4713 where fluids using field groups were not checking against a required 

In the pull request body, give a good description of what the nature of the change is, and the reasoning behind it. Provide links to external references/discussions if appropriate.

If this pull request resolves a project issue, provide a link with a closing keyword. For example:
  Resolves https://github.com/ExpressionEngine/ExpressionEngine/issues/4713

Assign this PR the appropriate label depending on what it does, e.g. 'Bug:Accepted', or 'enhancement'

### DIFF
--- a/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
+++ b/system/ee/ExpressionEngine/Addons/fluid_field/ft.fluid_field.php
@@ -131,16 +131,23 @@ class Fluid_field_ft extends EE_Fieldtype
                     $f->setName($field_name);
                     $f = $this->setupFieldInstance($f, $datum, !is_null($fluid_field_data_id) ? $fluid_field_data_id : $key);
 
+                    // Is Field Required And Empty?
+                    if ($f->isRequired() && empty($f->getData())) {
+                        $this->errors->addFailed('field_id_'. $field_id, ' This field is required');
+                    }
+
                     $validator = ee('Validation')->make();
                     $validator->defineRule('validateField', function ($key, $value, $parameters, $rule) use ($f) {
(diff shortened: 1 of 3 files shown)


robinsowell (changes requested):
It's working for me now running the validation, but I think it needs a few design tweaks.

1. The 'required' indicator doesn't show up.  Here it is in a fluid vs standalone:
Screenshot 2025-02-18 at 9 44 48 AM

I think we need the required indicator on there.

2. The required error message gets a little funky when used in fluid- this may be tricky to fix.  But see:

Screenshot 2025-02-18 at 9 46 21 AM

That bit is showing at the bottom of the fluid field.  It probably should be either at the top of the fluid field as a whole, or better yet, at the top of the group.

This bit- at the top of the fluid field, does show where you'd expect:

Screenshot 2025-02-18 at 9 46 53 AM

robinsowell:
After discussion, decided not to include in this milestone as it's a bit of a breaking change.

intoeetive:
This is a breaking change, I'm highly opposed to including it at all. 

Instead, we should provide a separate setting. Please see the discussion here: https://github.com/ExpressionEngine/ExpressionEngine/discussions/2954

PR #4733, opened 2025-03-06
Match function's return type

swierczek:
The function signature in `system/ee/ExpressionEngine/Service/TemplateGenerator/Factory.php` is `public function getFieldVariables($fieldInfo, $channelContext = null): array`, but in the conditional it's doing `return null` which throws an exception in some versions of PHP because the return type doesn't match the function signature.

This PR updates it to return an empty array so the return type matches, as well as updates the places it's used to account for an empty array instead of `null`.


bryannielsen:
@swierczek thanks for your PR!  I think this was probably an error on the function signature and the smaller change would be to make it nullable by prepending a question mark like this:

```
public function getFieldVariables($fieldInfo, $channelContext = null): ?array
```

If you agree and don't mind making that change I'd love to have you get the credit for working on this!

bryannielsen (approved):

matthewjohns0n (approved):

PR #4738, opened 2025-03-18
Improved #4693; show row count on grid fields

Yulyaswan:
Fixed #4693 where show row count on grid fields


robinsowell (changes requested):
Looks good- we just need to get docs in before release.

I do wonder if there's a better way to achieve what they're doing.  And I am still not clear on what they're doing. But as long as docs are clear it's a count rather than a row id, I think it's good.

Yulyaswan:
@robinsowell , I added a description.

robinsowell (approved):
Looks good- I tweaked the docs a wee bit. It's easy to confuse row id with count.

PR #4742, opened 2025-03-22
Resolved #4682 where cloning entry with hidden RTE field was giving PHP error

### LINKED ISSUE #4682: Clone an entry with a hidden rte field throws php error
Set up a channel with an rte field in it.  Create a layout, hide the field.

Go make an entry.  Then clone the entry.  You may or may not see this error on the landing page, but if you don't see it with your eyeballs, it should still be in the EE error logs:

```
ERROR - 2025-01-20 18:48:38 --> Severity: E_DEPRECATED  --> preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated /Users/robinsowell/Sites/75gen/system/ee/ExpressionEngine/Addons/rte/ft.rte.php 326
```

Might happen with other fields- no idea why RTE is a special snowflake.  I didn't SEE a referrence to any other fields, and I hid a few of them.  But something is clearly 'different' in the data when cloning.

EE 7.5.6 php 8.3

intoeetive:
Resolved #4682 where cloning entry with hidden RTE field was giving PHP error

### DIFF
--- a/system/ee/ExpressionEngine/Addons/rte/ft.rte.php
+++ b/system/ee/ExpressionEngine/Addons/rte/ft.rte.php
@@ -323,7 +323,7 @@ class Rte_ft extends EE_Fieldtype
     public function save($data)
     {
         // Trim out any whitespace/empty tags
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):

robinsowell (approved):
No more error!

PR #4743, opened 2025-03-22
Resolved #4697 where replacing the file did not update dimensions info

### LINKED ISSUE #4697: Replacing a file in the File Manager doesn't update meta-data
Uploading a file to the File Manager as a replacement for an existing file doesn't cause the file meta-data to be updated.

Example case: if I upload a very large image, and then later upload a smaller version of the image to replace this — same upload directory, same filename — having cropped/resized, the file size and dimension information in the File Manager remains that of the original file. This is misleading and incorrect.

[EE 7.5.7]

intoeetive:
Resolved #4697 where replacing the file did not update dimensions info

### DIFF
--- a/system/ee/ExpressionEngine/Service/File/Upload.php
+++ b/system/ee/ExpressionEngine/Service/File/Upload.php
@@ -359,6 +359,12 @@ class Upload
         );
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
Works and makes sense.

PR #4744, opened 2025-03-22
Resolved #4689 where adding category from entry page could show wrong UI

### LINKED ISSUE #4689: Oddness when managing cats in an entry with only single category selections allowed
Somewhat related to https://github.com/ExpressionEngine/ExpressionEngine/pull/4664 so bug is for EE 7.5.7.

If you have your channel set such that category selects are limited to 1 category, it all works fine.  BUT if have your one category selected you flip on 'manage categories' when on the entry page and you add a new category, it gets a little weird.  This is what I have after I did that and flipped manage cats back off:

Image

So the old category remains selected, the new category is also selected, there's an extra category selected that I'm not sure why.  And the display is off.  If you save the entry, you don't get any validation errors, but you end up with only one of the selections saved.  

Short story- with 'disallow multiple categories' on, if you manage categories in your entry and add a new one, things can get a bit weird display wise and you may end up with extra categories selected and yet not saved.

intoeetive:
Resolved #4689 where adding category from entry page could show wrong UI

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Categories/Categories.php
+++ b/system/ee/ExpressionEngine/Controller/Categories/Categories.php
@@ -606,7 +606,7 @@ class Categories extends AbstractCategoriesController
                     $field->setItem('field_required', true);
                 }
                 if (!$categoryGroupSettings->cat_allow_multiple) {
-                    $field->setItem('field_type', 'radio');
+                    //$field->setItem('field_type', 'radio');
                 }
             }
         }


robinsowell:
Tested and... it's better though not 100% what I expected.

If I have it limited to one, go and select one category then manage the categories in the entry to add another one, I still end up with the original and the newly added category checked and no warning on the page that I have a validation error.

Screenshot 2025-03-25 at 5 24 26 PM

The save button is still submitable as well.

BUT when I submit, it gives me a validation error rather than saving it.  So big improvement, the user can't think 2 categories are going in, they have to go fix their error.

But I did expect - at least if they turn off cat editing- that they'd get a red warning on the page at that point and the submit wouldn't be active.

I'd say release it as an improvement and the ui can still be tweaked when it's ready.

robinsowell (approved):
Approving with note we might could tweak ui more, but it does throw the validation error if both are picked.

PR #4747, opened 2025-03-23
Resolved #4708 where superadmins were not able to edit their own secondary roles

### LINKED ISSUE #4708: Superadmin can't remove a secondary role from themselves, but ought to be able
As a Superadmin, I should be able to change what secondary roles are assigned to my own account.
Instead, you get an error message when you try.
I understand that you should not be able to remove your own superadmin status. But other roles should be modifiable.

**How To Reproduce**
Go to the Member page. admin.php?/cp/members
For your own superadmin account, go to edit your "Roles"
You'll get an error

**Error Messages**
<pre><code>
500: Internal Server Error
Super Admins cannot change their role assignment. Please have another Super Admin reassign you.
</code></pre>

**Environment Details:**
 - Version: 7.5.7
 - PHP Version doesn't matter
 - MySQL Version doesn't matter
 - OS: doesn't matter
 - Web Server: doesn't matter

**Possible Solution**
it's just a logic check.

intoeetive:
Resolved #4708 where superadmins were not able to edit their own secondary roles

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Members/Profile/Profile.php
+++ b/system/ee/ExpressionEngine/Controller/Members/Profile/Profile.php
@@ -177,12 +177,7 @@ class Profile extends CP_Controller
 
             $list->addItem(lang('blocked_members'), ee('CP/URL')->make('members/profile/ignore', $this->query_string));
 
-            $sa_editing_self = ($this->member->isSuperAdmin() && $this->member->member_id == ee()->session->userdata['member_id']);
-            $group_locked = (! ee('Permission')->isSuperAdmin() && $this->member->PrimaryRole->is_locked);
-
-            if (! $sa_editing_self && ! $group_locked) {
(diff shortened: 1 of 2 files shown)


robinsowell (dismissed):
Hrm- something is going really wrong.  The roles are accessible and the superadmin main role is disabled, so all that is good.

But when I add a new 'regular' role and save, it logs me out and I can't get back in and if I look in the db in exp_members, my role_id is now 0.  Switch that back to 1 and I'm ok.

The secondary roles 'stick' and are checked when I can log back in, but it's just setting the primary role id in exp_members to 0

TomJaeger:
@robinsowell can you test this one and confirm fix?

Yulyaswan:
It looks correct to me, but I think it would be best if Robin could confirm as well.

robinsowell (approved):
Checks out for me- thanks!

PR #4766, opened 2025-04-01
Added update url button

Yulyaswan:
Added a button to update the Entry page and Structure URL based on the title after the Entry has been saved.

I am changing the PR status to “Draft” until the client responds to the questions.


themes/ee/asset/javascript/src/cp/publish/publish.js
```diff
@@ -54,6 +54,21 @@ $(document).ready(function () {
 		});
 	}
 
+	if (EE.publish.which == 'edit') {
```
TomJaeger, line 57: I think we need to add a title to this for accessibility

TomJaeger (commented):
see comments

PR #4774, opened 2025-04-05
Resolved #4762 where in some cases it was not possible to change site 404 template

### LINKED ISSUE #4762: Setting a default 404 page fails if there are many templates
On the Template settings screen, you can set a default 404 page template.

The initial list of templates is limited in number for performance reasons. But you can search all the templates to find the one you want to use.

However, if you try to choose a template that was NOT in the initial list, you are given a validation error and you cannot save your choice.

Therefore you can only currently set the 4040 template to something in the initial list of values, which is limiting.

There is a workaround: set it in config.php.  But this is not a good longterm solution.
`$config['site_404'] = 'search/404-page';`

You can also change the name of the template to something alphabetically prioritized.  Again, not a good longterm solution.

**How To Reproduce**
1. Start with a site that has many templates.
2. Go to the Template Settings screen.
3. Use the search box on the 404 setting to choose a template that is NOT in the initial list
4. When you select that template, you get the error message: "The option you selected is not valid." and you cannot save the settings
5. See screenshot below.

**Error Messages**

<pre><code>
The option you selected is not valid.
</code></pre>

Image

**Environment Details:**
 - Version: 7.5.8
 - PHP Version 
 - MySQL Version 
 - OS: 
 - Web Server: 

**Possible Solution**
The validation options appear to be set at load time. They should refresh when the search returns values. 

**Additional context**
I would be curious if to see if this bug happens on any other places in the UI where an initial validation list of values is loaded.

intoeetive:
Resolved #4762 where in some cases it was not possible to change site 404 template

Note: I'm not fan how the validation was / is currently working for ajax selects (essentially it might get skipped), but altering that would be rather a global change and new feature request. This PR is focusing on pushing the fix for the reported issue

### DIFF
--- a/system/ee/legacy/libraries/Form_validation.php
+++ b/system/ee/legacy/libraries/Form_validation.php
@@ -163,7 +163,8 @@ class EE_Form_validation
             if (is_array($setting)) {
                 foreach ($setting['fields'] as $field_name => $field) {
                     // For ajaxified fields with options not currently showing, we skip
-                    if (isset($field['filter_url']) && isset($field['choices']) && count($field['choices']) == 100) {
+                    $choicesLimit = isset($field['limit']) ? $field['limit'] : 100;
+                    if (isset($field['filter_url']) && isset($field['choices']) && count($field['choices']) >= $choicesLimit) {
                         continue;
                     }
 


robinsowell (approved):
Oh- this one was tricky.  I kept not being able to replicate.

It wasn't throwing the validation error the first time I picked an option that wasn't showing in the initial list.  But if I saved that, went back and picked ANOTHER option that wasn't in the list, I got the validation error.  

This fixed it- and everything else kept chugging away.

PR #4776, opened 2025-04-05
Resolved #4772 where MFA dialog could still be shown if set required for member but disabled globally

### LINKED ISSUE #4772: MFA is disabled in settings but if it was already 'on' for the user it's still kinda on
So... this one is weird and difficult to replicate.

User had the site offline.  Non-logged in users get the offline message.  I login - no mfa required.  I go to Settings and MFA is disabled there.  Cool- I assume it is totally out of the picture.

When I go to visit the frontend the first page loads but when I navigate to another page it either shows me the offline message (treating me as logged out) OR it gives me the MFA page.  Which is disabled so I don't know why that would ever happen.

They had some cookie issues which I resolved but would still get that mfa popping up on the frontend.

I finally went in and set all of the individual settings to 'n' via query and everything then worked normal:

```
Update exp_members set enable_mfa = 'n'
```

But... I think even when disabled site wide, at least on the frontend there's still some MFA check cropping up if it's set to 'y' for the individual user.

intoeetive:
Resolved #4772 where MFA dialog could still be shown if set required for member but disabled globally

### DIFF
--- a/system/ee/legacy/libraries/Core.php
+++ b/system/ee/legacy/libraries/Core.php
@@ -418,7 +418,7 @@ class EE_Core
         }
 
         // Is MFA required?
-        if (REQ == 'PAGE' && ee()->session->userdata('mfa_flag') != 'skip') {
+        if (REQ == 'PAGE' && ee()->config->item('enable_mfa') === 'y' && ee()->session->userdata('mfa_flag') != 'skip') {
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
zero fun to test, but this fixed it for me.

TomJaeger:
Looks like tests are failing.. not sure if they need to be updated.. or if something not intended is happening.

PR #4780, opened 2025-04-06
Resolved #4632 where test email was not sent using some SMTP servers

### LINKED ISSUE #4632: Outbound Email Test throwing false failure with smtp
This might happen in other circumstances, but it has definitely happened with office365 and gmail smtp.  If you do a test email, it says it failed, despite the fact if you send an email from Communicate it goes through.

To use gmail smtp, here's the setup.

First- the gmail account you're using to send the emails requires 2 factor authorization be on.
 
Second- you need to go create an 'app' and get the authorization code for that.  Go to https://myaccount.google.com/apppasswords create the app, name it whatever, copy down the code.  You'll use that instead of your gmail account password.
 
-- info on app passwords https://support.google.com/accounts/answer/185833?hl=en
 
And here are settings that worked- possibly not the ONLY settings, but I stopped at 'working'.
 
Screenshot 2024-12-10 at 5 39 53 PM

Go test sending an email from Communicate.  This works solid for me.  But the email test says:

Screenshot 2024-12-10 at 5 41 10 PM

intoeetive:
Resolved #4632 where test email was not sent using some SMTP servers

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Settings/Email.php
+++ b/system/ee/ExpressionEngine/Controller/Settings/Email.php
@@ -270,7 +270,7 @@ class Email extends Settings
         $config = [
             'wordwrap' => ee('Security/XSS')->clean(ee('Request')->post('word_wrap')),
             'mailtype' => ee('Security/XSS')->clean(ee('Request')->post('mail_format')),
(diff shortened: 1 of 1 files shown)


bryannielsen (approved):

robinsowell:
Whee- and Jared C tested it and confirms it worked for him.  Since... I could not test!

PR #4781, opened 2025-04-06
Resolved #4004 where CLRF setting was not properly set to handle quoted-printable emails

### LINKED ISSUE #4004: Formatting issue with HTML emails and 365 smtp adding = in middle of word.
We send our emails through office365 via smtp. Somehow every 70-80 characters a character gets replaced by an equal sign such as below:

Dear C=ris Thank =ou for contacting us. We app=eciate that you’ve taken the time to write. If you=eed a response, we’ll get back to you as quickly as we can.

I''ve found other users are experiencing this too:
https://expressionengine.com/forums/topic/253810/formatting-issue-with-html-emails-and-365-smtp-adding-in-middle-of-word

And i've found a possible solution here:
https://stackoverflow.com/questions/13065695/codeigniter-smtp-email-message-characters-replaced-with-equal-signs
"once I put in the "$this->email->set_crlf( "\r\n" );" it worked immediately without any issues!!!"

I have seen this issue on this site both when it was on EE5 a few months ago, but also now that it's running on EE7.3.14

I changed line 169 in /ee/legacy/libaries/email.php to public $crlf = "\r\n"; which seems to solve this

There is a sense of urgency to this one sine it's affecting orders from an ecommerce site

intoeetive:
Resolved #4004 where CLRF setting was not properly set to handle quoted-printable emails

### DIFF
--- a/system/ee/ExpressionEngine/Model/Config/Config.php
+++ b/system/ee/ExpressionEngine/Model/Config/Config.php
@@ -45,8 +45,8 @@ class Config extends Model
 
     public function set__value($value)
     {
-        // exception for email_newline, which uses backslashes, and is not a path variable
-        if ($this->key != 'email_newline') {
+        // exception for email_newline and email_crlf, which uses backslashes, and is not a path variable
+        if (!in_array($this->key, ['email_newline', 'email_crlf'])) {
             $value = str_replace('\\', '/', (string) $value);
         }
 
--- a/system/ee/legacy/core/Config.php
+++ b/system/ee/legacy/core/Config.php
@@ -353,6 +353,7 @@ class EE_Config
 
         $config['email_newline_form_safe'] = $config['email_newline'];
         $config['email_newline'] = $this->setEmailNewline($config['email_newline']);
+        $config['email_crlf'] = $this->setEmailNewline($config['email_crlf'] ?? '');
 
         if ($mutating) {
             $this->config = $config;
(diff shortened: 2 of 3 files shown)


robinsowell:
So I know there IS an issue and I'm sure this solves it.  But I never could replicate.  If I didn't have a config override set in config.php, crlf was 'n' per the default and if I set it it was whatever I set.

I do see it wasn't running through set_crlf but... it was already set to \n.

In any case- I don't understand why it breaks and I couldn't replicate quote printable not running if I had an override. And the fix didn't break anything either.

But I do think we either need to add a comment where we made the fix, because it looks nonsensical/like a typo.

Or- could we instead do:

```

        if (ee()->config->item('email_newline') !== false) {
            $config['newline'] = ee()->config->item('email_newline');
            //$config['crlf'] = ee()->config->item('email_newline');
        }
		
		$config['crlf'] = (ee()->config->item('email_crlf') !== false) ? ee()->config->item('email_crlf') : $this->crlf;

        //if (ee()->config->item('email_crlf') !== false) {
        //    $config['crlf'] = ee()->config->item('email_crlf');
        //}
```
Unless I'm missing something.

bryannielsen:
I think the reason that `$config['email_crlf']` wasn't always working for people in the original issue is because it was not handled the same way as `email_newline`.  

Setting `$config['email_crlf'] = '\r\n';` would not work properly but `$config['email_crlf'] = "\r\n";` would.  This was not in line with how `email_newline` was handled though and could obviously lead to confusion.  I pushed a commit that fixes this so the original issue should be addressable by using the `email_crlf` setting.  

However, I think it is still reasonable to use the `email_newline` value as a default for `email_crlf`.  It is also more convenient since `email_newline` is manageable through the Control Panel.  It is also still possible to set `email_crlf` to a different value if needed.

bryannielsen (dismissed):

bryannielsen (approved):

PR #4782, opened 2025-04-08
Add support for linkable internal tabs and fields that display correctly on page load using #anchors; #4736

Yulyaswan:
Added setting 'Add anchor links?' to make labels clickable and use them like anchors
1. if toggle is 'Off', then site url works like before, labels are not clickable, tab rel doesn't showing in url
2.  if toggle is 'On', user can use labels as an anchor

https://github.com/user-attachments/assets/dad9b967-09df-42f1-8bde-8c79a4bac0e2


intoeetive:
Loving it! I think the text could be improved, but leaving that to native speakers :)

PR #4792, opened 2025-04-20
URLFactory: cp_url was always ignored

Houdini:
Even if cp_url variable has been setup, http header location always returned to SELF value, as well as all links in CP.
This made hiding admin.php not as efficient as it was designed

### DIFF
--- a/system/ee/ExpressionEngine/Service/URL/URLFactory.php
+++ b/system/ee/ExpressionEngine/Service/URL/URLFactory.php
@@ -85,7 +85,9 @@ class URLFactory
             $session_id = $this->session_id;
         }
 
-        $cp_url = ($cp_url) ?: $this->default_cp_url;
+        if (empty($cp_url)) {
+            $cp_url = ($this->cp_url) ?: $this->default_cp_url;
+        }
 
         return new URL($path, $session_id, $qs, $cp_url, $this->uri_string, $this->encrypt_delegate);
     }
--- a/tests/cypress/cypress/elements/pages/channel/FieldGroups.js
+++ b/tests/cypress/cypress/elements/pages/channel/FieldGroups.js
@@ -10,7 +10,7 @@ class FieldGroups extends ControlPanel {
 
         this.selectors = Object.assign(this.selectors, {
             //"create_new": '.sidebar a.button--action.left',
-            "create_new": 'a[href="admin.php?/cp/fields/groups/create"]', //added /groups/
+            "create_new": 'a[href $= "admin.php?/cp/fields/groups/create"]', //added /groups/
 
             "field_groups": '.folder-list > div',
             //"field_groups_edit": '.folder-list div li.edit a',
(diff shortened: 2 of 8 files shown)


Houdini:
@bryannielsen hi! Could you please check this PR

bryannielsen:
@Houdini thanks for submitting this PR!  I'm going to approve for the tests to run but I'm going to need to take a closer look at how this is setup but I think your code is ignoring the local `$cp_url` variable in the `make()` method now which doesn't appear to be used too often but I did find this occurrence https://github.com/ExpressionEngine/ExpressionEngine/blob/7.dev/system/ee/ExpressionEngine/Addons/comment/Service/Notifications.php#L103 so that behavior should still be supported. 

Perhaps checking for `$cp_url` and then `$this->cp_url` and then `$this->default_cp_url` would be a good approach.

Houdini:
@bryannielsen Thanks for your response, let me research this. I'll try to prepare fix soon

Houdini:
@bryannielsen Could you please run tests once again?
I did the fix.

I could say, that variable shadowing is absolute evil :)

Houdini:
@bryannielsen hi! Could you please check this updated PR

bryannielsen:
@Houdini sorry for the delay and thanks for making the change!  
I agree the number of ways for that variable too be set is excessive.  Does your latest change still resolve your issue with customizing the urls?

Houdini:
@bryannielsen Yes, last change fix my issue.

I could setup test environment. It was not easy, as I never write php code in my life.
So what I can say, it's a test issue.

For example:
When access new Field group cypress is looking for `a[href="admin.php?/cp/fields/groupscreate"]`, but with my fix, this button it now has full url link:
`http://expressionengine.test/admin.php?/cp/fields/groups/create` 
where `expressionengine.test` is my localhost, and I have this line in my config:
`$config['base_url'] = 'http://expressionengine.test/';`

In the last change I also fixed tests, but I could not completly accomplish all tests passed at my localhost, like this one.
Test `tests/cypress/cypress/support/commands.js:145`

image

I have a lot of deprecated warnings, but not errors. Have no idea what to do with all of these.
What do you think? Maybe just ignore deprecated warnings?

Could you please run tests one again?

Houdini:
@bryannielsen hi, could you please run tests once again? And could you please check my last comment? Probably I'll need one more iteration after this one

Houdini:
@bryannielsen hi, eventually I remove test that test `on line` because it false positive and catch just deprecated warnings.
Could you please run tests once again. Should work now

Houdini:
@bryannielsen hello, please I need your review. I'd like to merge it, this fix my issue

bryannielsen:
@Houdini can you show me an example of what you're doing with the `cp_url` to hide `admin.php`?  I just want to make sure I understand the original issue and how this addresses it.

Houdini:
@bryannielsen I made a video overview for my problem, please check it: https://youtu.be/cQxVv7qVvOo

Houdini:
@bryannielsen friendly reminder

bryannielsen:
Thanks @Houdini the video was very helpful.  I see we have one failing test still and I need to take a closer look at the test you removed but if everything checks out I think we can move forward with this

Houdini:
@bryannielsen let me then fix the test. I'm glad that essentially we can forward

Houdini:
@bryannielsen could you please run tests once again

PR #4797, opened 2025-04-27
Enabled setting default value for some custom fields; #3632

intoeetive:
Enabled setting default value for some custom fields; #3632


intoeetive:
@Yulyaswan can you do some additional work on this one -  for the fieldtypes that support multiple values (such as checkboxes) adding default values does not work when it is used inside Grid (pressing "add new" button is submitting the form). I think there needs to be some extra JS added to `OptionFieldtype.php`

Once done, please give a quick test to everything and mark the PR as "ready for review" as I'll have very limited time next few weeks

Yulyaswan:
I fixed js code, but it still has a PHP warning (added screenshot and video)

https://github.com/user-attachments/assets/020453dc-25ee-43cd-8db2-e568c0d858ea

Screenshot 2025-04-28 at 14 33 38

+ I think we should add this default value so it can be used in the entry

PR #4805, opened 2025-04-30
4 new hooks for email and forgotten passwords

robinsowell:
Also, additional developer logging. Docs https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/1021

Proposed hooks:
member_auth_send_reset_token_start
cp_login_send_reset_token_start
email_from_address
email_to_address

Finalized hooks:
member_auth_send_reset_token_start
cp_member_send_reset_token_start
email_from_address
email_to_address

### DIFF
--- a/system/ee/legacy/libraries/Email.php
+++ b/system/ee/legacy/libraries/Email.php
@@ -481,9 +481,14 @@ class EE_Email
      */
     public function from($from, $name = '', $return_path = null)
     {
-        // email_from_address allows overriding the from address	
+        // email_from_address allows overriding the from/name
         if (ee()->extensions->active_hook('email_from_address')) {
-            $from = ee()->extensions->call('email_from_address', $from, $name);
+            $processed_address = ee()->extensions->call('email_from_address', $from, $name);
+            $from = $processed_address['from'] ?? $from;
+            $name = $processed_address['name'] ?? $name;
+            if (ee()->extensions->end_script === true) {
+                return;
+            }
         }
         
         if (preg_match('/\<(.*)\>/', $from, $match)) {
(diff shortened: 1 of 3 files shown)


robinsowell (commented):
I'm leaning strongly toward removing the logging.  I tend to think it should be in the extension rather than the core.

Also some concerns about how much this could fill up logs if they're sending emails to member groups and such.

system/ee/legacy/libraries/Email.php
```diff
@@ -481,6 +481,11 @@ public function clear($clear_attachments = false)
      */
     public function from($from, $name = '', $return_path = null)
     {
+        // email_from_address allows overriding the from address	
+        if (ee()->extensions->active_hook('email_from_address')) {
+            $from = ee()->extensions->call('email_from_address', $from, $name);
```
matthewjohns0n: Since we pass `$name`, do we want the ability to set it as well?

robinsowell: I see no reason not to.  I'm also not 100% sure why we pass it there, but giving them the extra flexibility seems reasonable.  I'll make it so.

matthewjohns0n: Yeah, my concern was that if they end up changing the email and the name, but then the name is the original name, and the email is the new email, it might send like that.

robinsowell:
@matthewjohns0n this one is ready for final (hopefully) review.

The main changes that are worth a look/thought are - in email library, what do we want to default to if the extension doesn't set something. And- do we want to do end_script on all of the new hooks.  I actually thought end_script was uniform across all hooks.  As in- if it's a hook, there's an end_script.  This does not look to be the case so, I figure it's worth an eyeball.

Once approved, I'll tweak docs to match the final product.

matthewjohns0n (approved):

PR #4808, opened 2025-05-01
Renamed hook and now low reorder extension not working....

robinsowell:
Low reorder has an extension calling 'low_search_post_search' that can be called to alter pro search ordering. But... we renamed it pro search and now Low Reorder doesn't find it because the name changed.

This renames it back- but now I'm wondering if this is the right fix.  Do we need to keep old name and new name- for like all the hooks? Kind of surprised this is the first time I ran into this.


robinsowell:
Discussed this with the team, and we're leaning toward including both a low and a pro version of all hooks, perhaps with pro having priority and going first.

PR #4815, opened 2025-05-04
Resolved #4773 where deprecation error could be shown on some MSM sites where Structure is installed

### LINKED ISSUE #4773: Error when saving an entry when Structure isn't set up for this site.
We have a system with four MSM sites. Some use Structure, some don't. On saving an entry where Structure isn't set up I get the following error:

<pre><code>
Deprecated

base64_decode(): Passing null to parameter #1 ($string) of type string is deprecated

ee/ExpressionEngine/Addons/structure/sql.structure.php, line 1437

Severity: E_DEPRECATED
</code></pre>

The entry still saves but the success screen isn't shown, instead the error blocks the output.

**Environment Details:**
 - Version: 7.5.7
 - PHP Version 8.3.19
 - MySQL Version 5.7.42
 - OS: Linux
 - Web Server: Apache

**Possible Solution**
Replace line 1437 of sql.structure.php with the following:

<pre><code>
if(is_string($pages_array))
{
		$all_pages = unserialize(base64_decode($pages_array['site_pages']));
}
else
{
	$all_pages = array();
}
</code></pre>

There could be a better solution, but this seems to work for me.

**Additional context**
Our site has been updated from Version 2 via intermediate versions, so it's possible that this is the source of the issue.

intoeetive:
Resolved #4773 where deprecation error could be shown on some MSM sites where Structure is installed

While I wasn't able to replicate the original issue, this should fix it

### DIFF
--- a/system/ee/ExpressionEngine/Addons/structure/sql.structure.php
+++ b/system/ee/ExpressionEngine/Addons/structure/sql.structure.php
@@ -1434,7 +1434,11 @@ class Sql_structure
         if ($cache_bust === true) {
             $sql = "SELECT site_pages FROM exp_sites WHERE site_id = $this->site_id";
             $pages_array = ee()->sql_helper->row($sql);
-            $all_pages = unserialize(base64_decode($pages_array['site_pages']));
+            if (empty($pages_array) || empty($pages_array['site_pages'])) {
+                $all_pages = array();
+            } else {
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
While I could not replicate either- and I did try- I tested the fix and no ill effects.

PR #4816, opened 2025-05-04
Resolved #4623 where autosaves for new entry could be overwritten by another member

### LINKED ISSUE #4623: Autosaves for new entries overwritten by other members
EE 7.5.3

It seems that autosaves for new, unsaved entries are shared among members, leading to overwrites. When someone works on a new entry, their autosave  overwrite the autosave of another member working on a different new entry. 

And yes, I still think that autosaved drafts for new entries should remain private to the individual who created them until the entry is explicitly saved or published. Autosaved drafts often contain incomplete thoughts, preliminary ideas, personal notes that should remain private, or sensitive information that the creator is not yet ready to share.

It's hard to believe that someone is actually using this to pick up unfinished work of others (see this issue: https://github.com/ExpressionEngine/ExpressionEngine/issues/3101)

intoeetive:
Resolved #4623 where autosaves for new entry could be overwritten by another member

Note: OP suggests that autosaves should not be author-specific at all, however I think some people might be using existing behaviour to pass unpublished work between editors, so we should not change that. This PR is only adding author check for new entries

### DIFF
--- a/system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
+++ b/system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
@@ -1615,6 +1615,23 @@ class ChannelEntry extends ContentModel
 
         return false;
     }
+
+    public function getAutosaves()
+    {
+        if ($this->isNew()) {
+            return ee('Model')->get('ChannelEntryAutosave')
+                ->filter('original_entry_id', 0)
+                ->filter('site_id', $this->site_id)
+                ->filter('channel_id', $this->channel_id)
+                ->filterGroup()
+                    ->filter('author_id', $this->author_id)
+                    ->orFilter('author_id', ee()->session->userdata('member_id'))
+                ->endFilterGroup()
+                ->all();
+        }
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Model/Channel/ChannelEntry.php
```diff
@@ -1615,6 +1615,23 @@ public function livePreviewAllowed()
 
         return false;
     }
+
+    public function getAutosaves()
```
TomJaeger, line 1619: @intoeetive I'm assuming this is a new helper method?

intoeetive: This is one of EE's "magic methods" - when you call `get<RelationshipName>` and the method is not defined, it will return you directly `<RelationshipName>`. This method is already being called in a couple of places, but wasn't defined, so returned all autosaved. I'm here defining the methods, making the results more filtered

bryannielsen (approved):

PR #4819, opened 2025-05-05
Resolved #4399 where Structure URI for cloned entry was built differently from cloned URL Title

### LINKED ISSUE #4399: When cloning an entry using Structure, the url_title & the structure_url are given different values.

When you are using Structure & you clone an entry, the url_title & the structure_url are given different values.

In my case, I cloned the Entry titled 'Food & Dining". The url_title was set to "copy-food-dining" and the structure_url was set to "food-dining-1" 

This isn't a HUGE issue I guess since the structure_url is what is looked at for the page, however, if a client is using "live Preview" it will use the url_title and tosses a 404 error on the page, until you change the structure_url. 

**How To Reproduce**
1. select an entry
2. Clone it
3. review the the entry to see differing url_title and structure_url's

In my case, I changed the status to "draft" to make changes to the page and saved it, and then when I click the "Preview" button on the publish tab, I get a 404 error.

**Error Messages**
tosses a 404 error on the preview page

**Environment Details:**
 - Version: 7.4.11

**Possible Solution**
only solution I can find is to edit the structure_url to match the url_title

**Additional context**

intoeetive:
Resolved #4399 where Structure URI for cloned entry was built differently from cloned URL Title

Note: this is not really a bug as such, as we still have unique value and editor would still probably want to change it... But it's nice when both are built using same pattern


robinsowell (changes requested):
Ah - the uri for structure was ending up without the copy on it- so basically a duplicate (which I thought wouldn't validate, but did and that might be another issue).

Basically- if I dump $_POST['structure__uri'] = $values['uri'] = $value; and die right before it returns- I get:

"copy-/a-photograph-for-the-ages" 

because we're putting that leading slash on the value before we add 'copy' to it.  Structure doesn't allow slashes inside the uri so I guess the whole copy bit gets dropped on save and you end up with just the original 'a-photograph-for-the-ages'.

TomJaeger:
@robinsowell any chance you can test this again?

robinsowell:
Hrm.  So definitely solved the first issue from testing and everything seem ok except for 1 minor inconsistency which ... I don't think is a problem, but is an inconsistency.

I create an entry, structure uri is blog-child-dos-uri and it is set as the child of another page. I clone that entry, cloned entry has structure uri of blog-child-dos-uri-1 and it has the proper parent selected.

See how it's blog-child-dos-uri-1? If I do the same thing but it's NOT a child, original uri is blog-top-uno-uri and copy uri is copy-blog-top-uno-uri.

In other words- always the clone url_title and structure/page uri is created with copy-whatever.  In the case of a subpage, it's not doing that.  

Again- not sure whether this is a problem or not.

@TomJaeger your thoughts?

robinsowell (changes requested):
Not requesting change per se, but I'd like someone else to weigh in on inconsistency noted in my last comment.

TomJaeger:
I would say we need to change this so it matches the URL title regardless of where it is in the structure tree

intoeetive:
The fix for this is now in place

robinsowell (dismissed):
Fixed for me!

I did discover this bug while testing: https://github.com/ExpressionEngine/ExpressionEngine/issues/4959 BUT that happens without these changes as well, so different issue.

TomJaeger:
Added a code review request from either @matthewjohns0n and @bryannielsen Im wondering if we need to change some of the logic in the sql.structure file code updates.

TomJaeger (approved):

PR #4822, opened 2025-05-05
Added Structure link from entry edit page; #4821

intoeetive:
Please see https://github.com/ExpressionEngine/ExpressionEngine/discussions/4821 for the discussion.

@Yulyaswan we might want to improve the UX for the button/link here? 
image

### DIFF
--- a/cp-styles/app/styles/components/_buttons.scss
+++ b/cp-styles/app/styles/components/_buttons.scss
@@ -311,34 +311,20 @@ input[type="button"] {
 
 .structure-uri-link {
   display: inline-flex; 
-  margin-left: 5px;
+  margin-left: 10px;
 
-  a {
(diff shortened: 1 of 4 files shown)


Yulyaswan:
Screenshot 2025-05-14 at 10 37 46
I added styles

I think it would be better if instead of an icon inside the button, we would have `"View" `word
@intoeetive what do you think?

intoeetive:
@Yulyaswan It's coming from a component that we use in Entry Manager listing, so if we can stick to same look, it is much easier.

However I wonder if we can make it less prominent? Maybe more like Preview button colors? Or even the same grey ones as on original, just smaller (although I think those don't play well in dark mode - we need to have that in mind too)

Yulyaswan:
Screenshot 2025-05-15 at 10 52 24
@intoeetive

PR #4830, opened 2025-05-11
Added column in Entry Manager to trigger Live Preview

intoeetive:
Added column in Entry Manager to trigger Live Preview

This PR adds a button for each row in Entry Manager where Live Preview is set up or available
image
What then it does is opening the saved entry's preview (according to channel settings or Structure URL) in a full-page iframe

This gives editor to have a quick peak of what entry this is on the front-end without having to go through entry's content


system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
```diff
@@ -659,85 +659,6 @@ protected function entryCloningEnabled(ChannelEntry $entry)
 
         return false;
     }
-
-    protected function createLivePreviewModal(ChannelEntry $entry)
```
TomJaeger, line 663: Is this a breaking change?  Should we really leave this method here and call the new one thats moved to the service?

`$livePreviewReady = ee('LivePreview')->createLivePreviewModal($entry);`

PR #4836, opened 2025-05-13
Display preview of Grid data in collapsed row header; #3262

intoeetive:
Display preview of Grid data in collapsed row header; #3262

See the FR in #3262

The implementation is quite basic, we just display a piece of data from the Grid's first column

That does not get auto-updated, just comes from the saved data

@Yulyaswan we need to add some code so it only displays when the row is collapsed, and also just make the UI look right

### DIFF
--- a/cp-styles/app/styles/components/_grid-field.scss
+++ b/cp-styles/app/styles/components/_grid-field.scss
@@ -689,8 +689,24 @@
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Yulyaswan:
https://github.com/user-attachments/assets/a7e45169-b66a-43fb-83d3-87cfbf25f107

Added styles and script

PR #4840, opened 2025-05-14
Resolved #4832 where RTE image button opens blank on first load

### LINKED ISSUE #4832: EE 7.5.11 RTE image button opens blank on first load
Came in from a user but I confirm.

Added an RTE field set to full CKeditor.  It's at the bottom of my 'About' entry form.  I click to make a new entry in about, I click on the button to add an image/file:

Image

And it ... kind of opens like a modal exept the whole screen is just white with a faint opacity:

Image

IF I click on it, it goes away and I'm back to seeing the entry form.  If I now click on the add image button again- it loads properly.

Is weird.

Edit to add- it can be a little quirky to reproduce, so it didn't happen on another channel.  Also didn't happen on 7.5.10, but when I upgraded that one to 7.5.11 and cleared my caches it DID happen.  I think it's a legit bug.

Yulyaswan:
Resolved #4832 where RTE image button opens blank on first load

### DIFF
--- a/cp-styles/app/styles/components/_modals.scss
+++ b/cp-styles/app/styles/components/_modals.scss
@@ -173,6 +173,11 @@
   width: 98%;
   max-height: 90vh;
   overflow-y: auto;
+  position: fixed;
+  top: 40px;
+  left: 50%;
+  transform: translateX(-50%);
+  z-index: 9999;
 }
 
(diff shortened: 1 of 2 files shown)


intoeetive (approved):
Fixing the issue for me with regular File field

PR #4848, opened 2025-05-15
Resolved #4847 where the Entry Manager could become unusable depending on Column configuration

### LINKED ISSUE #4847: Entry Manager and Column Expectations

If a user sets their Column grouping within the Entry Manager to a grouping that does NOT include a Column that can be sorted/ordered, a fatal PHP error is thrown and the Entry Manager becomes unusable for said member. 

Image

Further, a PHP notice gets thrown if the above since it's assumed the `entry_id` key would exist. I does not. 

This happens across PHP and MySQL versions and has no bearing on output. 

**How To Reproduce**
1. Within a Channel Entry List
2. Deselect all checked columns except Categories
3. Refresh the page
4. page borks

**Error Messages**

<pre><code>
Call to a member function getEntryManagerColumnSortField() on null
Edit.php:167

</code></pre>

**Possible Solution**
Checking for the existing and instanceof said variable

**Additional context**

eric-lamb62:
Resolves #4847 Where the Entry Manager could become unusable depending on Column configuration

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/Edit.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/Edit.php
@@ -16,6 +16,7 @@ use ExpressionEngine\Model\Channel\ChannelEntry as ChannelEntry;
 use ExpressionEngine\Service\Validation\Result as ValidationResult;
 use Mexitek\PHPColors\Color;
 use ExpressionEngine\Library\CP\EntryManager;
+use ExpressionEngine\Library\CP\EntryManager\Columns\Column;
 
 /**
  * Publish/Edit Controller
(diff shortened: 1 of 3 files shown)


intoeetive (commented):
@mithra62 do we really need `instanceof` check here? It looks like `isset` is sufficient as we always get columns from `ColumnFactory`

eric-lamb62:
@intoeetive I wouldn't say "need", as the `$columns` array is defined in the method, and it's clearly created through a `EntryManager\ColumnFactory`. But as that conditional contains logic that does require the existence of said variable and it being an instance of the `Column`  abstract, ensuring both makes sense to my mind. 

That said, that conditional being so strict is 100% a stylistic choice more than an immediate concern. More a "provides additional reliability so why not?" choice on my part.

bryannielsen (approved):

PR #4853, opened 2025-05-19
Resolved #4385 where Pro Search did not index File fields data well

### LINKED ISSUE #4385: Pro search file searches with new file format
Currently, pro search indexes file data by simply using the content of the field's table- i.e., 

Screenshot 2024-07-08 at 1 38 59 PM

This isn't useful- we need the filename in there.  We could put more- like captions or some such.  But at least the filename.

I have a solution that works, but I'm not at all sure it's the correct solution, so making this an issue rather than a pr.

In ft.file.php if you add this method:

```
	// prep data for pro search- give it the filename
	function third_party_search_index($data) {

		//ee()->file_field->parse_field($field_data);
		
		if (is_string($data) && preg_match('/^{file\:(\d+)\:url}/', $data, $matches)) {
        // If the file field is in the "{file:XX:url}" format
        // Set upload directory ID and file name
        $file_id = $matches[1];
        $file = ee()->file_field->get_file($file_id);
		
		
		return (string) $file['file_name'];
		
    } elseif (is_string($data) && preg_match('/^{filedir_(\d+)}/', (string) $data, $matches)) {
        // If the file field is in the "{filedir_n}image.jpg" format
        $dir_id = $matches[1];
        $file_name = str_replace($matches[0], '', $data);
		return $file_name;

	}
		return '';

	}
```
Then when you index, you get the filename in the index.  It sanitizes it, stripping out dashes and periods, but... that may well be fine.  I'm just not 100% sure this is the best way to go.  But it does work.

intoeetive:
Resolved #4385 where Pro Search did not index File fields data well

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -1136,6 +1136,36 @@ JSC;
         $modifiers = ['resize', 'crop', 'rotate', 'webp', 'resize_crop', 'length', 'raw_content', 'attr_safe', 'limit', 'form_prep', 'rot13', 'encrypt', 'url_slug', 'censor', 'json', 'replace', 'url_encode', 'url_decode'];
         return $modifiers;
     }
+
+    /**
+     * Prepare data for Pro Search
+     *
+     * @param array $data Field value
+     * @return string File name and title
+     */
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
Oh nice- I like subfolders too.

PR #4854, opened 2025-05-20
Added option to control field’s required status through channel layout

Yulyaswan:
Added option to control field’s required status through channel layout


robinsowell (commented):
Wow, this one is a bear to test.  Lots of moving parts! But- very cool feature.

There is one issue- I think it probably needs to be addressed but I'm not sure.

I had an install with an existing layout.  When I drop the updated files in, NEW layouts have the 'required' checkbox.  But the existing layout does not.

Screenshot 2025-08-05 at 4 41 42 PM

Left is existing layout, right is for one after I dropped in the new files.

It might take a db update to address and I'm not 100% sure it's a bug.  BUT... worth discussion.

Yulyaswan:
@robinsowell I added code for what you mentioned.
But apart from visual testing, I need someone to review code, and maybe there is a way to improve it

robinsowell (dismissed):
I tried REALLY hard to break it and I could not.  It worked with an existing layout that had a field that was set to required.  Without changing the layout, it showed up as checked.  Without saving anything in the layout, it was still required.  If I edited the layout to remove it as required in the layout, that worked.

I do think a couple of times I could trigger it not saving as unchecked when it was the only one checked but I could not reliably replicate.  Like it happened twice in maybe 30 tries.  So I'm writing that off for now and approving.  Suspect it's nothing to do with the new feature in any case.

Yulyaswan:
last changes

Fix publish-layout required toggle regressions for custom fields

Normalize required-state handling so layout UI, saved JSON, and validation stay consistent.
Also remove unnecessary per-field layout queries and tighten required-toggle scope to
custom channel fields only.

- FieldDisplay: use get_bool_from_string() for isRequired()
- FieldDisplay: write required state as boolString values ('y'/'n')
- ChannelLayout: remove per-field channel/category lookup in transform() (N+1)
- ChannelLayout: required_condition fallback is true only for field_id_* fields
- ChannelLayout: parse required_condition and required with get_bool_from_string()
- ChannelLayout: add required_condition=true when synchronizing new custom fields
- Legacy Layout sync defaults: set required_condition=false for system-managed fields
- Layout JS: persist required checkbox state with $(this).is(':checked')
- Layout JS: treat checked custom-required fields as required for tab hide/move guards
- Cypress: add scenario verifying tab hide is blocked when a custom-required field is moved into that tab

PR #4855, opened 2025-05-20
Resolved #3556 where some date fields still could not accommodate dates past year 2038

### LINKED ISSUE #3556: Date field cannot accept older dates
If you put a date with year 1901 or older in a date field- including entry date- it won't validate and you can't save it.

Summarizing: You can't have dates prior to 1901 or after 2038. This is the size limitation of int data types that we use for date columns.

2 possible solutions:
- allow changing column type from int to bigint same as we do for RTE fields. This is quick and easy, because we have the code already
- start using datetime column instead of int for dates. This is IMO the most proper solution and this how it should have been done 15 years ago, but making the change will be more involved.

intoeetive:
Resolved #3556 where some date fields still could not accommodate dates past year 2038

Turns out the https://github.com/ExpressionEngine/ExpressionEngine/pull/3911 PR was incomplete, some of date fields were omitted. This PR is fixing it


TomJaeger (dismissed):
Please see question below.

.gitignore
```diff
@@ -4,6 +4,7 @@
 Thumbs.db
 env_local.rb
 *.swp
+.osp
```
TomJaeger: This needed @intoeetive ?

intoeetive: Not directly related to this PR, just a piece from my dev environment that needs to be ignored

TomJaeger:
Looks like date picker isn't showing as it should in PHP 8.4 cypress tests so it keeps failing.. we'll have to look into this.

TomJaeger:
So other PR's are failing for same reason.. I think we need to look into why its happening, but it's likely not part of this PR.. merging

TomJaeger (approved):

PR #4856, opened 2025-05-21
Enabled inline display of SVG files using `wrap="svg"` parameter on file field;  Enabled passthrough of HTML attributes for `wrap="image"` parameter on file field

intoeetive:
Enabled inline display of SVG files using `wrap="svg"` parameter on file field;

Enabled passthrough of HTML attributes for `wrap="image"` parameter on file field

User Guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/992

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -335,7 +335,7 @@ JSC;
             $full_path = ee()->file_field->parse_string($data['url']);
 
             if (isset($params['wrap'])) {
-                return $this->_wrap_it($data, $params['wrap'], $full_path);
+                return $this->_wrap_it($data, $params['wrap'], $full_path, $params);
             }
 
             return $full_path;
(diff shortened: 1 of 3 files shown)


Yulyaswan (changes requested):
A few moments here
1. `{news_image wrap="image" width="width" height="height"}`

This tag returns the image with its original dimensions (as stated in the documentation).
However, in practice, the width and height attributes remain empty, for example:
``
Screenshot 2025-10-14 at 09 27 27
Screenshot 2025-10-14 at 09 27 51

This is effectively the same as using {news_image wrap="image"} without specifying any dimensions.

A more correct behavior would be to have the server detect the actual image size and populate these attributes with real values, e.g.:
``

2. `{news_image wrap="svg" class="my-svg" id="my-svg" style="border: 1px solid #000;" title="My SVG" alt="My SVG" width="100" height="100"}`

When using the wrap="svg" option, all provided parameters are applied correctly except for width and height, which are ignored.
Screenshot 2025-10-14 at 09 29 39
Screenshot 2025-10-14 at 09 30 19

intoeetive:
@Yulyaswan the 2nd is fixed
The 1st we probably can't do anything, as we don't have width/height of SVG image server-size. Please test with a regular image, it should work

Yulyaswan:
tested. Now everything works!!!
thanks, @intoeetive

Yulyaswan (approved):

PR #4860, opened 2025-05-21
Resolved #4849 where some inline errors in Channel Form were duplicated

### LINKED ISSUE #4849: Channel form inline errors {field_errors} tag pair
The[ {field_errors} tag pair](https://docs.expressionengine.com/latest/channels/channel-form/overview.html#field_errors) just outputs the error - no label or any indication of what the error is in relationship to.  It also looks like it shows double the errors.  

So if you have 2 missing required fields you get:

This field is required.
This field is required.
This field is required.
This field is required.

I see where that happens- there's one error for field_id_x key and one for the actual label- in the channel_form_lib.php around 1602.

We need to either give them a {label} variable or include the label in the {error} like it's done in the message when not using inline errors.  And also not double it up.

intoeetive:
Resolved #4849 where some inline errors in Channel Form were duplicated

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -1130,8 +1130,17 @@ GRID_FALLBACK;
             $conditional_errors['field_errors'] = array();
 
             foreach ($this->field_errors as $field => $error) {
-                $conditional_errors['field_errors'][] = array('field' => $field, 'error' => $error);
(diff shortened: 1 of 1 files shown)


robinsowell (approved):
All good.
```            {field_errors}
            
            {field}: {error}<br>

            {/field_errors}
```
Old:
Screenshot 2025-06-13 at 1 04 22 PM

New:
Screenshot 2025-06-13 at 1 04 31 PM

robinsowell:
See docs- yay:  https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/993

PR #4864, opened 2025-05-28
Show warning banner when editing entry that has recent autosave by other user; #4593

intoeetive:
Show warning banner when editing entry that has recent autosave by other user; #4593

This alert is show when opening an entry if the latest autosave for it was don by another user within latest hour. This gives a chance that they are still editing it, so we're informing 

image

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
@@ -285,6 +285,7 @@ abstract class AbstractPublish extends CP_Controller
         }
 
         $currentAutosaveId = null;
+        $loopIndex = 0;
         foreach ($entry->getAutosaves()->filter('channel_id', $entry->channel_id)->sortBy('edit_date')->reverse() as $autosave) {
             if (! isset($authors[$autosave->author_id]) && $autosave->Author) {
                 $authors[$autosave->author_id] = $autosave->Author->getMemberName();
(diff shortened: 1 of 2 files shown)


robinsowell:
I heard back from the user who was needing this for some clients, and I'm not sure the approach is going to do the trick.  The wrote up a good overview of how they tested, so including it below:

We tried it two ways:

> Both people go into the entry. One person edits it, and waits for the auto-save, while the other person does nothing but observe. No banner pops up.
> 
> One person goes into the entry while the other is at the main entries list. The one in the entry edits it, and waits for the auto-save. The other person then opens the entry. No banner pops up.
> 
> I also tested a third way by creating a new user for myself, logging into the site via an Incognito window, and going into the entry that displays the banner on the list screen showing it had auto-saves (like my previous email). There was still no banner appears on the entry once inside.

I'm going to setup a test myself, but I think it may need tweaks.

They were using the Too Many Cooks add-on, and gave a run down of how that worked:

And for what it is worth, this is a nutshell of how Too Many Cooks works:

- User is editing an entry.
- A different person opens the same entry; they are immediately presented with a red (pink) alert banner that advises that someone else is in the process of editing the entry, and the Save button is disabled.
- The person editing the entry saves their work and exits (or not, the trigger is saving it).
- The waiting person sees a green banner telling them the other person is done editing, but to REFRESH the entry to get the new data, or what this person adds will overwrite the other’s work.

I do not know how it works if two people open an entry at nearly the same time or closely one after another, don’t start editing right away but then one person does. We haven’t tested it that extensively.

We don't need to match that workflow, but it was good info, so including it.

PR #4871, opened 2025-05-30
Only initialize text formatting plugins when necessary

intoeetive:

### DIFF
--- a/system/ee/legacy/libraries/Typography.php
+++ b/system/ee/legacy/libraries/Typography.php
@@ -59,6 +59,7 @@ class EE_Typography
     public $censored_replace = '';
     public $text_fmt_types = array('xhtml', 'markdown', 'br', 'none', 'lite');
     public $text_fmt_plugins = array();
+    public $formatting_plugins_initliazed = false;
     public $html_fmt_types = array('safe', 'all', 'none');
     public $yes_no_syntax = array('y', 'n');
     public $code_chunks = array();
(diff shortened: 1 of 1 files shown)


system/ee/legacy/libraries/Typography.php
```diff
@@ -59,6 +59,7 @@ class EE_Typography
     public $censored_replace = '';
     public $text_fmt_types = array('xhtml', 'markdown', 'br', 'none', 'lite');
     public $text_fmt_plugins = array();
+    public $formatting_plugins_initliazed = false;
```
matthewjohns0n: This has a typo in the name - should be `formatting_plugins_initialized`

system/ee/legacy/libraries/Typography.php
```diff
@@ -219,12 +219,6 @@ public function initialize($config = array())
         if (bool_config_item('enable_censoring')) {
             $this->word_censor = true;
         }
-
-        /** -------------------------------------
-        /**  Fetch plugins
-        /** -------------------------------------*/
-        ee()->load->model('addons_model');
-        $this->text_fmt_plugins = ee()->addons_model->get_plugin_formatting();
```
matthewjohns0n, line 227: Looks like we are moving this to the `_set_preferences()` function, called by `parse_type()`. 

If you do not call `parse_type()` or `_set_preferences()`, the public var `text_fmt_plugins` array will be uninitialized. 

Based on looking at the code, I am approving this PR and think it should not be a problem. I just wanted to raise the possibility that a third-party add-on could actually be using this (although unlikely). However, since it is so unlikely, I think it is safe enough to approve.

matthewjohns0n (changes requested):
This looks good other than the typo in `formatting_plugins_initliazed`

system/ee/legacy/libraries/Typography.php
matthewjohns0n: Just fixed the typo 👍🏼

matthewjohns0n (approved):

TomJaeger:
failing tests... going to push off to another release

TomJaeger:
This keeps failing tests... I think we need to look at it a bit more indepth.

PR #4874, opened 2025-06-01
Added setting to collapse saved rows when loading Grid and Fluid fields; #4873

intoeetive:
Added setting to collapse saved rows when loading Grid and Fluid fields; #4873

### DIFF
--- a/cp-styles/app/styles/components/_grid-field.scss
+++ b/cp-styles/app/styles/components/_grid-field.scss
@@ -378,7 +378,7 @@
 			tr.grid__item--collapsed {
 				padding-bottom: 0 !important;
 
-				// td:not(.grid-field__item-fieldset) { display: none !important;}
+				td:not(.grid-field__item-fieldset) { display: none !important;}
 
 				td.grid-field__item-fieldset {
 					border-bottom: none !important;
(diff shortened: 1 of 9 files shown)


Yulyaswan:
I think it needs a bit more work
1. when we have row numbers -> need more styles (this I can do)
Screenshot 2025-08-19 at 09 49 29
2. Without row numbers, it hides grid rows, but basically, you can't do anything from Entry. No option to open and edit row. Unless it was the idea? @robinsowell (here we need to change the design to add a 'collapse' button somewhere)
Screenshot 2025-08-19 at 09 50 48
3. With Fluid field it works well because we have already option to collapse rows. and because of this we can change data from entry
Screenshot 2025-08-19 at 09 54 46

intoeetive:
@Yulyaswan there has been an error in the logic here - the "collapse" setting is only intended to be used with Vertical view. I fixed that, however there are still few issues that maybe you could fix

1) the setting is shown initially, before you make any selection for layout type. I believe we already had fix for this in place? Not a big deal though

2) if you have the setting on, and also have Row Numbers on, the row numbers don't seem to show up

I also added a colspan fix for new row feature - it was breaking the columns now that you have row numbers. If this PR gets delayed maybe you'll want to include that separately

PR #4883, opened 2025-06-05
Resolved #4881 where "No access redirect" was not performed if restricted template is used as layout

### LINKED ISSUE #4881: Template access, redirect doesn't work
1. If i have a {layout='group/_layout1'} , 
that has only access for members, 
then the no-access redirect on the layout doesn't work 
(it just shows a blank page).

2. in general the no-access redirect is unclear, 
as it does not do a redirect but show the other content in place (see below)

---
### Clarification

```php
//  template group/_layout1 (Access: members only, no-access redirect: group/nope)
info: {layout:contents}
```
```php
//  template group/test (Access: all)
{layout='group/_layout1}
Hello world
```
```php
//  template group/nope (Access: all)
Nope
```
1. **Bug**
When logged in and visiting group/test you get, `info: Hello world`
When logged out (e.g. private window) you get a blank page
Expected output, `Nope` (due to the no-access redirect on layout)
(When setting no-access on group/test template, then the "nope" is shown, it does not work for layouts)

2. **in-discrepancy**
    In general, the template setting "No access redirect" has unclear wording.
    It *redirects* the **"template-engine"** to use and show another template in its place,
    instead of *redirecting* the **"user"** to another group/template as the wording suggests.
    It just serves the group/nope content on group/test url with a 200 header

    What is expected, depends on the programmers intention
        - 403 / 200 shown in place on current uri (e.g. content with link to login-page)
        - 302 redirect to another page (e.g. http redirect to login-page)
        - showing 404 content in place almost never seems adequate, because the resource does exist but has access restrictions

intoeetive:
Resolved #4881 where "No access redirect" was not performed if restricted template is used as layout


matthewjohns0n (dismissed):
This code change looks good to me.

matthewjohns0n:
@robinsowell I approved this code change, but think it would be best if you checked out the issue and weighed in on it.

robinsowell (dismissed):
Solid!

I agree that we should likely change the wording on the template form, but I also think that should be a seperate issue.

TomJaeger:
I'd like to have us get the new unit tests merged in, and run this against those tests.

TomJaeger (approved):

PR #4893, opened 2025-06-16
Version 7.5.13

TomJaeger:
Version 7.5.13

### DIFF
--- a/system/ee/legacy/libraries/Session.php
+++ b/system/ee/legacy/libraries/Session.php
@@ -395,14 +395,14 @@ class EE_Session
         $this->sdata['sess_start'] = $this->sdata['last_activity'];
         $this->sdata['fingerprint'] = $this->_create_fingerprint((string) $crypt_key);
         $this->sdata['can_debug'] = ($can_debug) ? 'y' : 'n';
-        $this->sdata['mfa_flag'] = ($this->member_model->enable_mfa === true) ? 'show' : 'skip';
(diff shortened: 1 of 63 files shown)


bryannielsen (approved):

matthewjohns0n (approved):

robinsowell (approved):

system/ee/legacy/libraries/Session.php
```diff
@@ -395,14 +395,14 @@ public function create_new_session($member_id, $admin_session = false, $can_debu
         $this->sdata['sess_start'] = $this->sdata['last_activity'];
         $this->sdata['fingerprint'] = $this->_create_fingerprint((string) $crypt_key);
         $this->sdata['can_debug'] = ($can_debug) ? 'y' : 'n';
-        $this->sdata['mfa_flag'] = ($this->member_model->enable_mfa === true) ? 'show' : 'skip';
+        $this->sdata['mfa_flag'] = ((ee()->config->item('enable_mfa') === false || ee()->config->item('enable_mfa') === 'y') && $this->member_model->enable_mfa === true) ? 'show' : 'skip';
```
bakin1999, line 398: is this correct? The logic of false or y?

I dont know exactly what we are trying to do here but my gut says maybe we wanted true or yes?

bakin1999 (changes requested):

system/ee/legacy/libraries/Session.php
matthewjohns0n: I think that is specifically to address this one case:

https://github.com/ExpressionEngine/ExpressionEngine/pull/4845

bakin1999 (approved):

PR #4909, opened 2025-07-16
Fixed 4891: Removed the “*” displayed on individual input boxes of a required Grid field.

### LINKED ISSUE #4891: Remove the "*" that's displayed on individual input boxes of a required Grid field
The Grid field sometimes has a star (*) on input boxes to indicate Required, and sometimes does not.

I think it's sufficient to have it in the Grid row header, and not on every individual row.

This is because that would match having next to the Field Header, not the field box, and also because the star makes the grid rows take up more space than would otherwise be necessary, and display unevenly, which totally bugs me.

**How To Reproduce**
Create a Grid field with some required columns. Assign it to an entry.  Look at that entry.

Note that the star appears only when you are editing a pre-existing Grid row, not for new rows that you add  -- so that's another reason to remove the star.

Image

**Environment Details:**
 - Version: 7.5.12

Yulyaswan:
Fixed #4891: Removed the “*” displayed on individual input boxes of a required Grid field.

https://github.com/user-attachments/assets/a0bf44c6-0f48-4a40-bd3f-dca44a5aab4d

https://github.com/user-attachments/assets/81c7f2fb-3e93-4d47-8c22-1dd7b828de55

### DIFF
--- a/cp-styles/app/styles/components/_grid-field.scss
+++ b/cp-styles/app/styles/components/_grid-field.scss
@@ -34,7 +34,7 @@
 				position: relative;
 				top: 0;
 				line-height: 0.7;
-				display: inline-block;
+				display: none;
 			}
 		}
 	}
--- a/themes/ee/cp/css/common.min.css
+++ b/themes/ee/cp/css/common.min.css
@@ -31977,7 +31977,7 @@ fieldset .field-control #eye {
   position: relative;
   top: 0;
   line-height: 0.7;
-  display: inline-block;
+  display: none;
 }
 
 .grid-field__table th {


robinsowell (changes requested):
This fixes the extra '*' but if you set the grid to show row numbers, it somewhat messes that up.

Without fix:
Screenshot 2025-08-18 at 3 38 54 PM

With fix:
Screenshot 2025-08-18 at 3 37 53 PM

If you set things to show vertical:
Without fix:

Screenshot 2025-08-18 at 3 40 11 PM

With Fix:

Screenshot 2025-08-18 at 3 41 06 PM

robinsowell (approved):
AHA- works fine if I pull down the whole branch.  This is good to go!

PR #4934, opened 2025-08-07
Resolved #4933 changed row getter to account for model object as well

### LINKED ISSUE #4933: PHP Error with Channel form selects
when using a select fieldtype  in channel form it throws an error due to an object being present when it expects an array in some circumstances.

array_key_exists(): Argument #2 ($array) must be of type array, ExpressionEngine\Addons\Pro\Model\Channel\ChannelEntry given
ee/legacy/fieldtypes/EE_Fieldtype.php:179

TomJaeger:
Resolved #4933 changed row getter to account for model object as well.. also flattened out the code a bit

### DIFF
--- a/system/ee/legacy/fieldtypes/EE_Fieldtype.php
+++ b/system/ee/legacy/fieldtypes/EE_Fieldtype.php
@@ -163,10 +163,9 @@ abstract class EE_Fieldtype
     }
 
     /**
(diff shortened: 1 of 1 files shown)


620studio:
I can confirm it works on EE 7.5.14

matthewjohns0n (approved):

PR #4941, opened 2025-08-17
Remove extra revision from Revisions table

swierczek:
This PR removes the the first "dummy" row in the revisions table and adds the "Current" tag to the top revision to indicate that it's the most recent content. This also updates some toolbar config options to allow adding an inline "Current" tag.

This is something that's always been confusing to me, because it adds an extra "revision" to the top, so the table view didn't match the database contents in the `entry_versioning` table.

### After creating a new entry

Previous functionality:
image

New functionality:
image

### After creating a new entry (not viewing a specific revision)

Previous functionality:
image

New functionality:
image

### Viewing a previous revision

Previous functionality:
image

New functionality:
image

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
+++ b/system/ee/ExpressionEngine/Controller/Publish/AbstractPublish.php
@@ -184,10 +184,10 @@ abstract class AbstractPublish extends CP_Controller
             if ($index === 0) {
                 $toolbar_items['current-revision'] = array(
                     'button' => false,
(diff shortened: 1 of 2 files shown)


intoeetive (commented):
This is right - "current" is not a revision. This is state of entry as it is currently. We keep it so that if you go to view a revision, you would have a way to return to currently saved state of entry without using the revision data.

swierczek:
Updated this to add a link back to the "current" tag to maintain the way back to the current entry when viewing a revision:

image

PR #4951, opened 2025-08-27
Add extension hooks for the Updater service

bryannielsen:
Add 3 new hooks - `updater_start`, `updater_error`, `updater_complete`

### DIFF
--- a/system/ee/ExpressionEngine/Cli/Commands/CommandUpdate.php
+++ b/system/ee/ExpressionEngine/Cli/Commands/CommandUpdate.php
@@ -87,6 +87,17 @@ class CommandUpdate extends Cli
 
         $this->runUpgrade();
         $this->postFlightCheck();
+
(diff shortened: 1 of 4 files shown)


intoeetive (commented):
@bryannielsen are we intentionally not adding `end_script` check for 2 other hooks?

Also, needs to ne documented

bryannielsen:
Great question @intoeetive!  Yes I left the `end_script` check off of `updater_complete` and `updater_error` because I think those should be available as notification only.  I don't think we want an Extension to be able to change the flow for how ExpressionEngine handles a finished update or a failed update.  However for `updater_start` it seemed appropriate that an Extension might want to control whether or not an update can be performed.

TomJaeger:
@bryannielsen do you have a docs PR for this as well?

TomJaeger (approved):

PR #4958, opened 2025-08-29
Add/pro search mod tests

TomJaeger:
This is based off the structure branch.. so that branch should be merged first.
Add/pro search mod tests


bryannielsen (commented):
Should these be passing?

TomJaeger:
I think they were... we'll take a peek.

TomJaeger:
not using this... using another similar one instead.

PR #4960, opened 2025-08-29
check for path before setting it in tests

TomJaeger:
check for path before setting it in tests


TomJaeger:
included in another PR

PR #4978, opened 2025-09-14
Resolved #4967 where it was possible to create Upload Directory in site base directory

### LINKED ISSUE #4967: EE Root Directory Creation Bug
ExpressionEngine currently allows any writable folder path to be used as a file upload destination, including the web root itself. It defaults to {base_path} when creating a new Upload Directory, and choosing ‘Local’ file adapter, the form defaults to {base_url} and {base_path}. When submitting the form, there is no validation to warn or prevent submitting with default values. This should be prohibited or at least a confirmation should be required to use these default values. It is not a good idea to use these defaults!

This absence of validation on this form becomes much more impactful when the site also uses Assets addon. When Assets is installed & enabled, and a file upload destination is created at the root level, critical errors/issues will likely result. When this occurs, the Assets addon will scan the entire site folder structure recursively, creating database entries for every single folder in the file system, leading to severe database bloating and performance degradation.

We had this happen when a client created a new upload destination using the web root (base_path}. It caused significant bloating of the Assets data and needed some painstaking and meticulous cleanup to rectify. We think that this should NOT be possible to add a file upload destination at the web root, or provide some serious alerts/confirmation if someone attempts to do it. 

Image

**Specific Issue:**
- Created a new EE file destination with pointing to site root  using default value of {base_path}
- Assets addon automatically scanned entire site folder structure recursively
- Created over 6k database rows in exp_assets_folders table
- Database table exp_assets_folders became severely bloated
- Files allowed to be uploaded to web root; this shouldn’t be allowed

Expected Behavior: 
EE should prevent configuration of upload destinations pointing to site root (base_path). Some validation should prevent this, or some confirmation step should be required when adding/updating a file upload destination. Documentation and/or instructions should indicate that the default values can’t be used without adding sub-folder to the path/url. 

Actual Behavior: 
EE allows the creation of upload destinations at the web root (base_path). This causes severe issues, including the Assets add-on silently scanning and indexing all site files and folders, leading to performance problems and system instability.

**Steps to reproduce:**
1. Navigate to Control Panel > Files > File Manager
2. Create a new Upload Directory
3. Set the server path to site root (e.g., /var/www/html)
4. In Assets, trigger ‘refresh’ on the newly created upload directory.
5. Assets will scan the entire site structure and create database entries for every file & folder in the web root.
6. Check database: SELECT COUNT(*) FROM exp_assets_folders WHERE filedir_id = 'YOUR_ID'

**Error Messages:**
No explicit error messages are generated, but the following issues manifest:
- Severe database performance degradation
- Assets plugin displays thousands of duplicate/inappropriate folder & file records
- Potential database timeouts and memory issues
- Unusable Assets file manager interface

**Database bloating evidence:**
**Query:** _SELECT COUNT(*) FROM exp_assets_folders WHERE filedir_id = 'YOUR_ID'_
**Result:** Your entire site folder structure will be indexed by assets  

Environment Details:
Version: EE 7.4.8
PHP Version: 8.2
MySQL Version: mariadb:10.6
OS: Linux
Web Server: nginx

**Possible Solution:**
Immediate Preventive Measures:
- Add validation, confirmation or error messaging in EE file uploads create/edit (cp/files/uploads/create)  to prevent root level paths being added
- Update documentation to be explicit about why you shouldn’t use {base_path} as upload destination, advise on a best practices pattern like creating an /assets folder with all upload directories under that (as example)

intoeetive:
Resolved #4967 where it was possible to create Upload Directory in site base directory

### DIFF
--- a/system/ee/ExpressionEngine/Library/Filesystem/Adapter/Local.php
+++ b/system/ee/ExpressionEngine/Library/Filesystem/Adapter/Local.php
@@ -64,7 +64,7 @@ class Local extends Flysystem\Adapter\Local implements AdapterInterface, Validat
                 'fields' => [
                     'url' => [
                         'type' => 'text',
-                        'value' => $settings['url'] ?? (strrpos(ee()->config->item('base_url'), '/', strlen(ee()->config->item('base_url')) - 1) === false) ? '{base_url}/uploads' : '{base_url}uploads',
+                        'value' => $settings['url'] ?? (strrpos(ee()->config->item('base_url'), '/', strlen(ee()->config->item('base_url')) - 1 === false) ? '{base_url}/uploads' : '{base_url}uploads'),
                         'required' => true
                     ]
                 ]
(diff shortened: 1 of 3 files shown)


bryannielsen (approved):
There is one failing test from this PR's actions but all tests pass for me locally

PR #4985, opened 2025-09-17
temporarily disables query caching IF there any filters for use durin…

### LINKED ISSUE #4984: Pro Search / Custom Filters / Query Cache
I had an issue where custom Pro Search filters, within my custom Add-on, were executing but the results weren't being applied to the output. For example, I would exclude a specific `entry_id` yet said entry would be within the search results. 

After digging into it, stepping through EE, I discovered that if `enable_sql_caching` was set to `y` then filter data isn't applied to the results. Disabling `enable_sql_caching` confirmed this to be the case. 

**How To Reproduce**
1. Create custom Add-on with a filter object
2. Set `enable_sql_caching` to `y`
3. Within your filter, manually add an `entry_id` to the `exclude` method.  
4. Perform a search using Pro Search
5. Note in wonder at your unexpected but totally expected result with your removed entry

**Error Messages**
None

None

**Environment Details:**
All 

**Possible Solution**
I discussed this offline and have a solution in mind where we can avoid touching the `Channel` object. Basically, dynamically setting and reversing `enable_sql_caching` when/if filters are applied to any results. Easy, clean, avoids updating tests.

eric-lamb62:
…g the entry compilation of pro search results

Fixes #4984 

just a simple enable / disable flow if channel query caching is enabled AND any filters are to be processed.

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
+++ b/system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
@@ -753,6 +753,12 @@ class Pro_search
         ee()->load->library('Pro_search_fields');
         ee()->load->library('Pro_search_filters');
 
+        //disable entry sql caching if using filters
+        $enable_sql_caching = ee()->config->item('enable_sql_caching');
+        if(count(ee()->pro_search_filters->names()) >= 1 && ee()->config->item('enable_sql_caching') == 'y') {
+            ee()->config->config['enable_sql_caching'] = 'n';
+        }
+
         ee()->pro_search_filters->filter();
 
         // --------------------------------------
@@ -894,6 +900,11 @@ class Pro_search
 
         if (ee()->extensions->active_hook('pro_search_channel_entries') === true) {
             $tagdata = ee()->extensions->call('pro_search_channel_entries');
+        }
+
+        //enable anew if needed
+        if($enable_sql_caching != ee()->config->item('enable_sql_caching')) {
+            ee()->config->config['enable_sql_caching'] =  ee()->config->item('changed_enable_sql_caching');
         }
 
         // --------------------------------------
--- a/system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
+++ b/system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
@@ -905,7 +905,7 @@ class Pro_search
         //enable anew if needed
         if($enable_sql_caching != ee()->config->item('enable_sql_caching')) {
             ee()->config->config['enable_sql_caching'] = $enable_sql_caching;
-;        }
+        }
 
         // --------------------------------------
         // If not set by the extension, use native channel:entries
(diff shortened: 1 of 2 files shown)


system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
```diff
@@ -753,6 +753,12 @@ public function results()
         ee()->load->library('Pro_search_fields');
         ee()->load->library('Pro_search_filters');
 
+        //disable entry sql caching if using filters
+        $enable_sql_caching = ee()->config->item('enable_sql_caching');
```
TomJaeger, line 757: wondering if this should be like $original_sql_caching

system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
```diff
@@ -896,6 +902,11 @@ public function results()
             $tagdata = ee()->extensions->call('pro_search_channel_entries');
         }
 
+        //enable anew if needed
+        if($enable_sql_caching != ee()->config->item('enable_sql_caching')) {
+            ee()->config->config['enable_sql_caching'] =  ee()->config->item('changed_enable_sql_caching');
```
TomJaeger: Is the logic right here?  I'm not seeing it set the changed_enable_sql_caching?

Should this actually be set to the $original_sql_caching above.. see comment above

TomJaeger (commented):
Thanks for submitting this!  Makes sense for us to get this in there..

Please see comments above.. wondering if we need a tweak before it merges in?

system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
eric-lamb62: @TomJaeger that's what line 907 is doing. I just named it poorly and, kinda, against logic lol 

Will add some sanity to the variable key naming and update :)

eric-lamb62:
@TomJaeger Will do a sanity run through to make sure it reads logically.

system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
```diff
@@ -896,6 +902,11 @@ public function results()
             $tagdata = ee()->extensions->call('pro_search_channel_entries');
         }
 
+        //enable anew if needed
+        if($enable_sql_caching != ee()->config->item('enable_sql_caching')) {
+            ee()->config->config['enable_sql_caching'] = $enable_sql_caching;
+;        }
```
TomJaeger: @mithra62 looks like a stray extra ; here

TomJaeger (commented):
noticed a stray ;

system/ee/ExpressionEngine/Addons/pro_search/mod.pro_search.php
eric-lamb62: @TomJaeger no, that's legit, I think. The previous problem was erroneous setting value of `changed_enable_sql_caching` instead of the local variable `$enable_sql_caching`. Should be good but please lemme know if I'm missing the point.

TomJaeger: I think the first character of line 908 is a semicolon ?

eric-lamb62: oh geeze, yeah, completely missed that and misread the intent. Thanks for following up. Pushed.

TomJaeger (dismissed):

TomJaeger:
@codex please review

TomJaeger (approved):

PR #4988, opened 2025-09-20
Set the `:resize_crop` modifier behave the same as "Crop" pre-defined manipulation; #2924

intoeetive:
Set the :resize_crop modifier behave the same as "Crop" pre-defined manipulation; #2924

User Guide amendment: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/1071

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -463,6 +463,7 @@ JSC;
         }
 
         unset($params['wrap']);
+        $params['use_larger'] = true;
 
         $data['source_image'] = $resized = $this->process_image('resize', $data, $params, false, true);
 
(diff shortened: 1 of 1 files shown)


intoeetive:
Tested and confirmed by @pyrobob83

pyrobob83:
@intoeetive Just wanted to reiterate [this comment](https://github.com/ExpressionEngine/ExpressionEngine/discussions/2924#discussioncomment-14477620) as I discovered it after confirming the function worked:

After further testing, the efficiency the page doesn't seem right - as though the caching isn't working. With my old setup using {gallery:file:resize height='350' width='10000' quality='90'} (ie just resizing to 350px wide for all images), I get about a 7-8 second load time after clearing the _resize dir, then consistently around 0.8s for subsequent loads... but with the patch from https://github.com/ExpressionEngine/ExpressionEngine/pull/4988 and using {gallery:file:resize_crop width='350' height='350' position='center' crop:maintain_ratio='n' quality='80'}, I seem to get that 7-8s load time every time. The images don't appear to be regenerated on subsequent loads (based on the modified date not changing), so I'm not sure what this is... any ideas?

TomJaeger (approved):

TomJaeger (changes requested):
please see comment above.

intoeetive:
The only reason I can think about is us calling `get_image_properties` because everything else is just math... But I need someone else to have a look. @matthewjohns0n maybe?

PR #4999, opened 2025-09-29
Refactor drag-and-drop view for "Select Entries" (Pro Variable) to use Relationship field

Yulyaswan:
Request:
Screenshot 2025-09-29 at 11 55 01

Replaced the custom drag-and-drop view implementation for the "Select Entries" (Pro) variable  
Now it uses the built-in Relationship field instead

### DIFF
--- a/system/ee/ExpressionEngine/Addons/pro_variables/css/pro_variables.css
+++ b/system/ee/ExpressionEngine/Addons/pro_variables/css/pro_variables.css
@@ -247,12 +247,14 @@
     cursor:grab;
     -webkit-border-radius:3px;
     -moz-border-radius:3px;
-    border-radius:3px
+    border-radius:3px;
+    color: var(--ee-text-primary);
+    font-size: 14px;
 }
 
 .pro-drag-lists ul li:hover {
-    background:#eee;
-    color:#000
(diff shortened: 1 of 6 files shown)


robinsowell:
So I don't know if I missed something setting it up or not, but when on the Relationship version, it's not displaying more than one select after saving.

Before upgrading, I had 2 options selected.  After upgrading, only 1 was showing.  But the data were correct in the field.  If I save with only 1 showing, then it updates to only that 1 saved.  Which does make sense.

I go do it and select multiples- I save- it properly updates to a pipe delimited list in the database- but still only 1 shows.

If I have it in multi-select mode, it's working as expected.

Display and db data below:

Screenshot 2025-10-08 at 12 59 10 PM

It's not impossible that I didn't replace a file or something- but can you double check that multiples are displaying properly?

robinsowell (changes requested):
Not sure this is working- can you double check the display of multiples? More detail in comments.

robinsowell (approved):
This got it working for me.  I did find 1 tiny thing:

Screenshot 2025-10-23 at 5 21 22 PM

I'm not sure why that one has the unparsed markup in the message- I checked on relationship in a regular entry and it doesn't.

PR #5002, opened 2025-10-01
Improve entry delete confirmation when deleting a member #4998

### LINKED ISSUE #4998: Improve entry delete confirmation when deleting a member
More a FR but given the data loss if someone accidentally deletes entries when deleting a member...

Improve the member delete confirmation modal so that they have to basically double opt in to deleting rather than re-assigning entries.  Something similar to how file deletion works would be an improvement- toggle you're confirmation that you want the entries deleted. Or something.

Yulyaswan:
Resolved #4998 
Added confirmation toggle to the delete modal window

Also from slack - https://app.slack.com/client/T04CUNNQH/C04CUNNR9

### DIFF
--- a/cp-styles/app/styles/layout/_member-manager.scss
+++ b/cp-styles/app/styles/layout/_member-manager.scss
@@ -57,3 +57,5 @@
         }
     }
 }
+
+.member-delete-confirm { margin-bottom: 0;}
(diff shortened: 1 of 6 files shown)


robinsowell (changes requested):
Yea- I think this was a really nice addition.  Only 1 issue- you can also delete a user from their profile page- aka url /cp/members/profile/settings&id=4

Screenshot 2025-10-08 at 11 46 15 AM

I don't think it's being applied there.

Yulyaswan:
thanks @robinsowell 
These changes should fix the problem

robinsowell (approved):
Looks good!

PR #5020, opened 2025-10-20
Add profiler output on entry save

swierczek:
This PR adds some advanced functionality to allow developers to view the profiler on entry save. This was massively useful to us to help troubleshoot slow save requests with thousands of redundant queries.

It's a little bit tricky do want or need to do this outside of a development environment, because the save request _normally_ redirects back to the entry page, but this intercepts that and just outputs the raw profiler data to the page and dies. So I'm open to suggestions of alternative conditional logic to output this (i.e. a toggle in the CMS, but I thought that was more persistent than necessary when you probably just want to profile a single save, not _all saves until you toggle it off_).

For quick and easy testing, edit any entry and run:

```$('.form-standard form').prepend('<input type="hidden" name="profiler" value="1" />');```

in the browser console to add the hidden input field, and then save the entry.

I'd be happy to add documentation to the User Guide to demonstrate how to use this if the PR is accepted.

Profiler output:
Shows the usual user/server data:
image

And duplicate queries:
image

And a list of all queries:
image

### DIFF
--- a/system/ee/ExpressionEngine/Service/Model/Model.php
+++ b/system/ee/ExpressionEngine/Service/Model/Model.php
@@ -382,6 +382,29 @@ class Model extends SerializableEntity implements Subscriber, ValidationAware
 
         $this->emit('afterAssociationsSave');
 
+        // Output the profiler output on channel entry save.
+        // In order to trigger this, add a hidden input field named 'profiler' with a truthy value to the edit entry form.
+        if (ee()->input->post('profiler') && $this->getMetaData('hook_id') === 'channel_entry') {
+            $performance = array(
+                'database' => number_format(ee('Database')->currentExecutionTime(), 4),
+                'benchmarks' => ee()->benchmark->getBenchmarkTimings()
+            );
+
+            $profiler = ee('Profiler')
+                ->addSection('performance', $performance)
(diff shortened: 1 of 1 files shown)


litzinger:
Have you thought about moving this to a custom extension that can be enabled or disabled at any time, thus you won't need to add a profiler hidden field or variable somewhere, and instead of printing to the page, print to a log file so the request can complete as normal. Could timestamp the log files too so you have more insight to the save requests over a period of time.

The `die` statement feels abnormal for EE requests and might be confusing.

swierczek:
@litzinger it's definitely advanced functionality and would likely only be used/useful in rare circumstances. There's a lot of data output from this (10+ MB), so I don't think logs are a useful route in terms of longevity or ongoing logging of this full data. Though that _would_ let the request complete, which would be nice. But again, advanced functionality.

We added this to a custom extension hook initially (`after_channel_entry_save` to be executed last), but realized it might be useful for others to have it as part of the core, so here we are.

TomJaeger:
Just a quick update on this one.. We love it, and where it's heading... we're thinking we might make it a large integration into the core.. ideally where ever there is a redirect, to add a new tab / dropdown to bring this to light.

PR #5030, opened 2025-10-27
Resolved #5011 where file size was not updated when file is replaced

### LINKED ISSUE #5011: Replacing a file in the File Manager doesn't update file size
Uploading a file to the File Manager as a replacement for an existing file doesn't cause the file size to be updated.

Example case: if I upload a very large image, and then later upload a smaller version of the image to replace this — same upload directory, same filename — having cropped/resized, the file size information in the File Manager remains that of the original file. This is misleading and incorrect.

Note that the intention was that this would have been resolved with issue #4697, along with any other meta-data. The PR for this issue (#4743) was included in 7.5.11, and the issue was closed, but it looks like only the file *dimensions* were fixed. The file size issue remains.

[EE 7.5.16]

intoeetive:
Resolved #5011 where file size was not updated when file is replaced

### DIFF
--- a/system/ee/ExpressionEngine/Service/File/Upload.php
+++ b/system/ee/ExpressionEngine/Service/File/Upload.php
@@ -532,6 +532,7 @@ class Upload
                 }
 
                 $original->file_hw_original = $file->file_hw_original;
+                $original->file_size = $file->file_size;
 
                 $file->delete();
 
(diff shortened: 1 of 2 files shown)


TomJaeger:
tests keep failing on this one

bryannielsen (approved):

robinsowell (approved):
Fixed it for me!

PR #5041, opened 2025-10-31
Add database cache driver to caching library

bryannielsen:


system/ee/legacy/libraries/Cache/drivers/Cache_database.php
```diff
@@ -0,0 +1,322 @@
...
+
+    /**
+     * Check if the cache table exists
+     *
+     * @return	bool	TRUE if table exists, FALSE otherwise
+     */
+    protected function _table_exists()
+    {
+        $tables = ee()->db->list_tables();
+        return in_array(ee()->db->dbprefix . $this->_cache_table, $tables);
```
TomJaeger: can we use the AR table_exists?

ee()->db->table_exists("table_name")

system/ee/legacy/libraries/Cache/drivers/Cache_database.php
```diff
@@ -0,0 +1,341 @@
...
+
+        return $this->_create_cache_table();
+    }
+
+    /**
+     * Create the cache table
+     *
+     * @return	bool	TRUE on success, FALSE on failure
+     */
+    protected function _create_cache_table()
```
TomJaeger, line 253: @bryannielsen Think we should move this to the version update file? and install file?

TomJaeger (approved):

PR #5045, opened 2025-11-04
new php cli script for bumping ee version

TomJaeger:
new php cli script for bumping ee version

PR #5049, opened 2025-11-04
Resolved #5044 adding keys to the file_name and title

### LINKED ISSUE #5044: add file_name index to exp_files table
On a site with very many files (many tens of thousands) noticed that a model call 

```
$file = ee('Model')->get('File')->filter('file_name', $file_basename)->first();
```

was running unexpectedly slowly.

Upon investigation it turns out exp_files is only indexed on the `upload_location_id` and `site_id`

Have added some work-around code to compensate, but seems likely that many people will want to search for files by file_name ... so maybe would be helpful if this column was also indexed.

TomJaeger:
Resolved #5044 adding keys to the file_name and title

### DIFF
--- a/system/ee/installer/schema/mysql_schema.php
+++ b/system/ee/installer/schema/mysql_schema.php
@@ -1477,7 +1477,9 @@ class EE_Schema
 			KEY `upload_location_id` (`upload_location_id`),
 			KEY `directory_id` (`directory_id`),
 			KEY `file_type` (`file_type`),
-			KEY `site_id` (`site_id`)
+			KEY `site_id` (`site_id`),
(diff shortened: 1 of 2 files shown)


build-tools/README.md
bryannielsen, line 1: This looks like it came over from #5045, should we remove from this PR or make sure they go in the same release?

TomJaeger: yeah, Idea is that we merge the other... since this used that new bump version script.. then it's just the version bump, with the needed index's added.... need the version bump for the index's being added as well, in the upd file

bryannielsen (dismissed):

robinsowell (approved):

PR #5062, opened 2025-11-13
Update the react select component to use a virtualized list of options

bryannielsen:
This will improve render performance dramatically with a large list of options.  Along the same lines we are deferring the formatting of these options until they are visible to save render time.


Yulyaswan:
Script works for me well.
Thanks Bryan

TomJaeger (approved):

PR #5063, opened 2025-11-13
Resolved a performance issue in the CP Relationship field with many categories

bryannielsen:
Switching to active record to avoid the memory and execution time overhead which adds up when dealing with thousands of categories

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/libraries/Relationships_ft_cp.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/libraries/Relationships_ft_cp.php
@@ -88,47 +88,78 @@ class Relationships_ft_cp
 
         $from_all_sites = (ee()->config->item('multiple_sites_enabled') == 'y');
 
-        $categories = ee('Model')->get('Category as C0')
-            ->with(array('Children as C1' => array('Children as C2' => 'Children as C3')))
-            ->fields('C0.cat_id', 'C0.cat_name')
-            ->fields('C1.cat_id', 'C1.cat_name')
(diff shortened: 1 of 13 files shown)


system/ee/ExpressionEngine/Addons/relationship/libraries/Relationships_ft_cp.php
```diff
@@ -88,47 +88,78 @@ public function all_categories()
...
-            ->with(array('Children as C1' => array('Children as C2' => 'Children as C3')))
-            ->fields('C0.cat_id', 'C0.cat_name')
-            ->fields('C1.cat_id', 'C1.cat_name')
-            ->fields('C2.cat_id', 'C2.cat_name')
-            ->fields('C3.cat_id', 'C3.cat_name')
-            ->filter('parent_id', 0)
-            ->order('group_id', 'asc')
-            ->order('parent_id', 'asc')
-            ->order('cat_name', 'asc');
+        $query = ee()->db->select('parent_id, cat_id, group_id, site_id, cat_name, cat_order')
```
TomJaeger, line 91: Should we change this to * ? @bryannielsen

bryannielsen: I don't think we need to, this is only for use in the CP and these are the only fields we need.  However we almost have all the columns in use anyways so if you would feel better about using `*` we could

TomJaeger: I'm good with that.

TomJaeger (approved):

PR #5069, opened 2025-11-19
Fix resizing of webp files with transparency when using GD lib

robinsowell:
It was uploading them fine, but if resizing, the transparency was replaced by a solid color.

Note- it's not clear this always happened.  But at least some cases where running gd with webp support, they were losing the transparency.

Both myself and end user confirmed the fix worked, but it could definitely use a dev eyeball.

### DIFF
--- a/system/ee/legacy/libraries/Image_lib.php
+++ b/system/ee/legacy/libraries/Image_lib.php
@@ -803,8 +803,8 @@ class EE_Image_lib
      */
     public function image_preserve_alpha($new_img, $src_img)
     {
-        // Preserve transparancies for GIFs and PNGs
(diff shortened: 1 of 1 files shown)


bryannielsen (commented):
Does it work correctly if you use the webp constant `IMAGETYPE_WEBP` instead of `18`?  I think that would be more readable https://www.php.net/manual/en/image.constants.php#constant.imagetype-webp

intoeetive:
@bryannielsen the reason we used number is because the constant is not defined in older PHP versions

TomJaeger:
Closing this in favor of WEBP and AVIF support in PR#5243

PR #5091, opened 2025-12-07
Added `image_manipulation_quality` config override; #4992

intoeetive:
Added `image_manipulation_quality` config override; #4992

User Guide: https://github.com/ExpressionEngine/ExpressionEngine-User-Guide/pull/1101


robinsowell (approved):
Very cool- I tried hard to break- did a sync, cropped, rotated made a watermakr for an image manip.  I can't swear what quality was used, but it certainly didn't break.

PR #5099, opened 2025-12-21
Service and CLI command to pack and unpack add-on

intoeetive:
Service and CLI command to pack and unpack add-on

To unpack, the zip file for the add-on needs to be placed in `system/user/cache/addons` directory, the file name can be anything. The folders structure can be different, but that of course need to be tested. The themes folder need be placed in `themes/user`, or `themes/<addon-name>` within zip file, but that can also be nested few levels deep.

I am extending the already existing Unpacker class used for EE update. Once we go forward with on-line updates, we might as well add validation for downloaded file integrity. 

The installed add-on can be packaged, zip to be saved in same folder.

CLI commands are:
`php eecli.php addons:pack`
`php eecli.php addons:pack --addon cartthrob`

`php eecli.php addons:unpack`
`php eecli.php addons:unpack --addon cartthrob`
`php eecli.php addons:unpack --addon cartthrob --delete-zip`

### DIFF
--- a/system/ee/ExpressionEngine/Service/Updater/Downloader/Unpacker.php
+++ b/system/ee/ExpressionEngine/Service/Updater/Downloader/Unpacker.php
@@ -279,7 +279,8 @@ class Unpacker
             }
 
             $addon = ee('pro:Addon')->get($addonShortName);
-            $resultMessage = sprintf(lang('addons_unpack_complete'), $addon->getName());
+            $addonName = !is_null($addon) ? $addon->getName() : $addonShortName;
+            $resultMessage = sprintf(lang('addons_unpack_complete'), $addonName);
         } else {
             $resultMessage = sprintf(lang('addons_unpack_failed'), implode(', ', $failedToMove));
         }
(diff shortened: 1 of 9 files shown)


system/ee/ExpressionEngine/Service/Updater/Downloader/Unpacker.php
```diff
@@ -168,6 +169,215 @@ public function moveUpdater()
...
+                        $addonFolders[$addonShortName] = $item->getPathname();
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (empty($addonFolders)) {
+            $files = new FilesystemIterator($extracted, FilesystemIterator::UNIX_PATHS);
+            foreach ($files as $item) {
```
TomJaeger: @intoeetive Can we flatten this out some?  Maybe with early if checks, and continues or exists..  Or switching things to some methods?

intoeetive: I did code refactoring, it's now cleaner and should support all types of folder structure

PR #5114, opened 2026-01-16
Fixed #5106 where Pro Variables: Select Entries Relationship UI Issue (Search and Channel Filtering returning no entries)

### LINKED ISSUE #5106: Pro Variables: Select Entries Relationship UI Issue (Search and Channel Filtering returning no entries)
Select Entries variable type in Pro Variables returning `NO <B> Entries </B> FOUND` when utilizing the Relationship fieldtype UI search or channel filter.

**Screenshots**
Image<br>Attempting to use channel filtration.<br>
Image<br>Attempting to use the search.<br>

**Environment Details:**
 - Version: Latest
 - PHP Version 8.3
 - MySQL Version mariaDB 10.11
 - OS: Linux
 - Web Server: Apache/nginx

Yulyaswan:
Fixed #5106 where Pro Variables: Select Entries Relationship UI Issue (Search and Channel Filtering returning no entries)

### DIFF
--- a/cp-styles/app/styles/components/_dropdown.scss
+++ b/cp-styles/app/styles/components/_dropdown.scss
@@ -348,3 +348,7 @@
     }
   }
 }
+
+div[data-relationship-react] .dropdown {
+	overflow: visible;
+}
--- a/cp-styles/build/main.min.js
+++ b/cp-styles/build/main.min.js
@@ -156,7 +156,7 @@ require.register("cp-styles/app/ee-lang.js", function(exports, require, module) 
  * ExpressionEngine User Guide (https://github.com/ExpressionEngine/ExpressionEngine-User-Guide)
  *
  * @link      https://expressionengine.com/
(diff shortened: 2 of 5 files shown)


robinsowell (changes requested):
So I think I have two minor issues.

First, when I have 'no results' in the search, there's unparsed html.

Screenshot 2026-01-21 at 9 21 53 AM

And second, when I clicked to filter by blog, if I have a really long blog name, it gets a little squished.  I compared it to the relationship field for entries and that one displays a little different.

Screenshot 2026-01-21 at 9 21 32 AM

vs

Screenshot 2026-01-21 at 9 27 32 AM

I think it's because in the entry page, you've got the 'New Entry' button adding some padding.  This one is definitely NOT a bug, but worth looking at and making sure it's what we want.

robinsowell (approved):
Looks good from here.

PR #5136, opened 2026-02-12
Optimize/structure adv nav

robinsowell:
A lot of tweaks- mostly adding a field parameter so we don't have to pull in a ton of excess data to parse through.

Also- a potential bug in caching addressed.

Need a really good review and some testing.


robinsowell:
Adding some template debug to show homepage load without vs with this change:

Screenshot 2026-02-23 at 3 15 56 PM

Note- this really scales up with the more fields you have.  Testing on my default install, the difference wasn't really noticable.  But if you have a lot of custom fields and you really just need to show one or two in the advanced nav? This can have a big impact.

PR #5163, opened 2026-02-22
Resolved #5146 where custom fields could have not been used for sorting Entries in Relationship field

### LINKED ISSUE #5146: EE7 Relationship field — custom ordering in Order By
Thank you for confirming that this is not currently possible in ExpressionEngine 7.

We would like to formally submit a feature request regarding ordering options in the Relationship field.

At present, the “Order By” setting is limited to Title and Entry Date. For many real-world use cases, this does not provide sufficient flexibility. We would like to request consideration for the following enhancements:

Support for additional “Order By” options within Relationship fields, including the ability to sort by custom fields (such as a numeric priority field or custom date field).

The ability to sort by core fields such as url_title.

Alternatively, the introduction of a documented hook or extension point that allows developers to modify the ORDER BY clause of the relationship query.

Our use case specifically requires more granular control over how related entries are ordered, particularly for editorial workflows where ordering is driven by priority values or non-standard dates rather than Title or Entry Date.

We understand there may be architectural considerations, especially when custom fields span multiple channels. However, even a developer-level override would provide meaningful flexibility without requiring UI-level complexity.

We appreciate your consideration of this enhancement and would be happy to provide additional context if helpful.

intoeetive:
Resolved #5146 where custom fields could have not been used for sorting Entries in Relationship field

### DIFF
--- a/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
+++ b/system/ee/ExpressionEngine/Addons/relationship/ft.relationship.php
@@ -719,11 +719,8 @@ class Relationship_ft extends EE_Fieldtype implements ColumnInterface
                 'desc' => sprintf(lang('rel_ft_order_desc'), strtolower(lang('entries'))),
                 'fields' => array(
                     'relationship_order_field' => array(
-                        'type' => 'radio',
-                        'choices' => array(
-                            'title' => lang('rel_ft_order_title'),
-                            'entry_date' => lang('rel_ft_order_date')
-                        ),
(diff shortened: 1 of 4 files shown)


intoeetive:
trying to figure out why PHPUnit gives this error

Error: Call to a member function filter() on null
   │
   │ /home/runner/work/ExpressionEngine/ExpressionEngine/system/ee/ExpressionEngine/Addons/relationship/libraries/Relationships_ft_cp.php:334

The code seems to be valid and works otherwise

PR #5175, opened 2026-02-26
Fix a bug with MSM CP switch

robinsowell:
So MSM site, users with multiple roles.  The dropdown shows all sites that at least one of their roles allowed CP access to, but the 'switch' failed because cp bootstrap was only looking at primary role as a first condition.

I realize this fix looks weird and like it's not checking for permissions.  However, it DOES check all role permissions later on.  However, it was bailing out early because it wasn't finding the role in primary.

### DIFF
--- a/system/ee/legacy/libraries/Session.php
+++ b/system/ee/legacy/libraries/Session.php
@@ -1204,6 +1204,14 @@ class EE_Session
             ee()->db->where('member_id', (int) $member_id);
 
             $data = ee()->db->get();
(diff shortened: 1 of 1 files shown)


robinsowell:
Note- I want some more testing before we pull this.  Still poking options.

robinsowell:
Scratch this- did a new expanded pr: https://github.com/ExpressionEngine/ExpressionEngine/pull/5177

PR #5183, opened 2026-03-03
Allow limiting usage of fields inside Fluid and making those required; #2954

intoeetive:
Allow limiting usage of fields inside Fluid and making those required; #2954


intoeetive:
@Yulyaswan I need your help here. I have updated the UI for Fluid field options; the toggles show up fine, but the respective hidden fields are not being created _when the field is initially loaded_. (They are created fine when I change the selection)

I had this problem with the "main" checkbox as well (used to select the field) - I have resolved this by adding a chunk of code to select_list file - but I also don't know if that's the optimal solution

I'd like it to work just the same way as Category group selection in Channel settings - but apparently it does not

intoeetive:
This is still work in progress and I'm honestly not sure if I'll get time to finish it soon, so maybe better someone else jumps in. The settings part is ready, but the "required" check only works when you actually add a field and "allow multiple" is not implemented at all 

@TomJaeger

PR #5208, opened 2026-03-28
Added configuration for default CP Theme; #5207

### LINKED ISSUE #5207: Ability to set default CP theme
Right now you have to use the Jump menu or member account toggle to change your CP theme. That's great but if you clear your browser cache or CP cookies expire it reverts to the light theme, a pain if you always use one of the other themes.

Suggestion is to add the ability to set the default theme in Settings. You could also allow overrides per role, or per member so it can be tailored at any level.

Not urgent of course but a nice to have!

intoeetive:
Added configuration for default CP Theme; closes #5207

### DIFF
--- a/cp-styles/pages/_header.php
+++ b/cp-styles/pages/_header.php
@@ -22,7 +22,7 @@
 </head>
 <body data-theme="light">
 <script>
-	var currentTheme = localStorage.getItem('theme');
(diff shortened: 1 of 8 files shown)


Yulyaswan (changes requested):
https://github.com/user-attachments/assets/e0cc8f58-9df6-4718-93e8-f8a48f26beea

I expected the theme to change after selecting it in General Settings and saving the page. However, it doesn't seem to be working—I've demonstrated this in the attached video.

PR #5215, opened 2026-04-02
bugfix/post-max-size-fail-http-response

robinsowell:
This was driving me to distraction.  Was testing uploading a file that was larger than max post size- and I get the max post size error.  But I discovered ANY fail seems to get it.  Like- if I die with 'what up' in the file manager constructor?  Max post size error.  

But- ignoring THAT- in the response header I'm getting a csrf error.
Screenshot 2026-04-02 at 11 00 35 AM

This did not give me joy.  So basically- this fixes that.  But definitely needs eyeballs since I was very focused on fixing the specific thing that was making my life harder.

But as far as the fix:
fix(file-upload): surface post_max_size errors instead of CSRF expiry in file picker uploads
- add filepicker ajaxUpload() guard for oversized requests where PHP drops POST/FILES and upload_location_id is missing
- return upload_file_exceeds_limit JSON error instead of falling through to show_404()
- send CSRF token in upload XHR headers (X-CSRF-TOKEN and X-EEXID) so token validation is resilient when POST body is
- update drag/drop uploader fallback to prefer server error payloads in default response handling

### DIFF
--- a/system/ee/ExpressionEngine/Addons/filepicker/mcp.filepicker.php
+++ b/system/ee/ExpressionEngine/Addons/filepicker/mcp.filepicker.php
@@ -317,6 +317,24 @@ class Filepicker_mcp
         $subfolder_id = (int) ee('Request')->post('directory_id');
 
         if (empty($dir_id)) {
+            ee()->load->helper('number_helper');
+            ee()->lang->loadfile('upload');
+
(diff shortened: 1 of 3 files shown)


TomJaeger (approved):

TomJaeger:
@codex please review

PR #5218, opened 2026-04-10
Fix a bug when syncing international characters in files

robinsowell:
Had an issue syncing with a file with ö in the name.  Sync would appear to never finish, though the file did end up in the upload directory.  Same file could be uploaded fine.

Looks like there are a few differences in unicode character handling- namely getDirectoryMap() → Flysystem Local adapter + getWithMetadata() + basename() on macOS (darwin) can return decomposed Unicode (NFD) while PHP/DB/Flysystem normalization elsewhere uses composed (NFC). Comparisons like $fileInfo['path'] != $clean_filename or getFileByPath() DB lookups can behave unexpectedly.

This fix does a few things:

* Stray d in sync JS throws ReferenceError on the “undefined errors” branch and can break the sync flow
* doSyncFiles() returning false in a few early cases rather than json
* flattenDirectoryMap() assigning recursive return (null) - bigger change, but I think it's solid.  Wasn't critical to the issue at hand, but now can recurse without overwrite; keep parent paths for non‑empty dirs so subfolders still sync in a sane order

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Files/Uploads.php
+++ b/system/ee/ExpressionEngine/Controller/Files/Uploads.php
@@ -928,7 +928,10 @@ class Uploads extends AbstractFilesController
             ($current_files = ee()->input->post('files')) === false
             && $db_sync != 'y'
         ) {
-            return false;
+            return ee()->output->send_ajax_response([
+                'message_type' => 'failure',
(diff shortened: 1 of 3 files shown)


TomJaeger:
@codex please review

PR #5259, opened 2026-05-06
added ability include or exclude entry ids in search form

TomJaeger:
added ability include or exclude entry ids in search form

Docs PR#1142


robinsowell (approved):

jacksonmg:
@TomJaeger this is exactly what I was looking for from the Slack thread. I tested this out on my local instance and it works as expected!

PR #5273, opened 2026-05-11
Resolved #4705 where Control Panel log Search field can't be cleared

### LINKED ISSUE #4705: Control Panel log Search field can't be cleared
When viewing the Control Panel logs at `/admin.php?/cp/logs/cp`, there is a keyword search input field. If you type something in there and hit enter to submit the form, it returns results correctly. But if you clear out that field and submit the form again, the keyword search autofills again on page load.

Initial page loads `/admin.php?/cp/logs/cp`:

Image

Entering a search term loads `/admin.php?/cp/logs/cp&perpage=25`:

Image

Deleting the search term and submitting the form loads `/admin.php?/cp/logs/cp&filter_by_keyword=Created&perpage=25` - strangely, now the URL contains `filter_by_keyword`.

Image

So it actually looks like the search results are correct, but the keyword search input field doesn't get cleared when you submit. So if you clear the field and submit the form, and then change some other setting (e.g. Show 25 -> 50), then the search term is used again.

**How To Reproduce**
1. View the Control Panel logs at `/admin.php?/cp/logs/cp`
2. Type something into the Search input field and hit enter
3. Delete the text from the Search input field and hit enter
4. Now, the search results seem to have filtered correctly, but the search term you deleted is now back.

**Environment Details:**
 - Version: 7.5.7

**Possible Solution**
I suspect the cause of the issue is https://github.com/ExpressionEngine/ExpressionEngine/blob/7.dev/system/ee/ExpressionEngine/Controller/Logs/Cp.php#L89 because the URL contains the $_GET parameters already.

Yulyaswan:
Resolved #4705 where Control Panel log Search field can't be cleared

### DIFF
--- a/system/ee/ExpressionEngine/Controller/Logs/Cp.php
+++ b/system/ee/ExpressionEngine/Controller/Logs/Cp.php
@@ -40,7 +40,17 @@ class Cp extends Logs
         $sites = ee('Model')->get('Site');
         $logs = ee('Model')->get('CpLog');
 
-        if ($search = ee()->input->get_post('filter_by_keyword')) {
+        $search = null;
+        $keyword_post = ee()->input->post('filter_by_keyword');
+        $is_clearing_keyword = ($keyword_post !== false && trim((string) $keyword_post) === '');
+
+        // CP logs only: when keyword is intentionally cleared via POST,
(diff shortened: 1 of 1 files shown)


TomJaeger:
@codex please review

bryannielsen (approved):

Issue #4901, opened 2025-06-26
Cached dynamic image not cleared when image changed


robinsowell:
To replicate:

```        {exp:channel:entries channel="slides" entry_id="19"}
        <h1>{title}</h1>


            <div>

{my_file:webp height='200' master_dim='height' maintain_ratio='y' quality='82'}
{url}<br>
<img src="{url}" width="{width}" height="{height}" alt="{title}">
{/my_file:webp} 

<hr>

{my_file}
url: {url}<br>
<img src="{url}">

{/my_file}
                      

        {/exp:channel:entries}
```

Upload a file to a file field in a channel entry- look at a template with the above.  Edit the entry, replace the file with a different picture that has the same name- choose to replace the old file when you get a conflict warning.

Reload the page.  The standard url and any thumbs show the new image.  The webp generated on the fly is still the old image.  I suspect we need to clear those cached images when it's altered.

bryannielsen:
I'm pretty sure this is the same issue and solution as we saw here - https://github.com/ExpressionEngine/ExpressionEngine/issues/4217#issuecomment-2075507063

creativearc-paul:
This issue still occurs in EE 7.5.21, May 2026.

PR #5277, opened 2026-05-12
Resolved 4901 cached webp files not updated

robinsowell:
https://github.com/ExpressionEngine/ExpressionEngine/issues/4901

If you're using a file manipulation to create a webp file and then you go change the file to a new one, it does clear out those cached modified files, BUT it was missing webp because the format of the modified file names differs from what it was expecting.  So the webp file just stayed a manip of the old, now replaced file.


bryannielsen:
This fix looks good but I think we need to rebase it on top of #5342 because that includes a refactor of process_image that should be taken into account here.  I can put that on my plate

PR #5322, opened 2026-06-16
Resolved #5320 where Channel Form {field_id} was not always parsed

### LINKED ISSUE #5320: {field_id} not parsed inside {custom_fields} loop in Channel Form on PHP 8.1+

Inside an `{exp:channel:form}` tag, the documented `{field_id}` variable within the `{custom_fields}` loop is **not parsed** on PHP 8.1+. It renders literally as `{field_id}` instead of being replaced with the numeric field ID, while sibling variables in the same loop (e.g. `{field_name}`) resolve correctly.

This breaks any template that builds a dynamic variable/label name from the field ID, e.g. a bilingual label key like `{{field_name}_{field_id}_{segment:lang}}` (intended to resolve to `{title_ar_9_en}`) never forms because `{field_id}` stays literal.

The same template works on one server and fails on another with the **identical EE version**, which is what makes this confusing. The differentiator is the DB layer's numeric return type:

- On PHP ≤ 8.0 (or driver/builds that stringify numerics), `field_id` returns as a **string** → it is substituted correctly.
- On PHP 8.1+ with PDO_MySQL (which returns native int/float by default), `field_id` returns as an **integer** → it is skipped and left literal.

Root cause is in the `{custom_fields}` substitution loop in `Channel_form_lib.php`:

```php
foreach ($custom_field_variables_row as $key => $value) {
    if (is_array($value)) {
        $temp = $this->swap_var_pair($key, $value, $temp);
    } elseif (! is_int($value)) {
        // don't use our conditionals as vars
        $temp = ee()->TMPL->swap_var_single($key, $value, $temp);
    }
}
```

The `! is_int($value)` guard (meant to keep integer-valued conditionals from being swapped as variables) also skips `field_id` when the DB layer returns it as a native integer.

**How To Reproduce**

1. Run EE on PHP 8.1+ with the PDO MySQL driver (default native int return types).
2. Create a front-end template containing a Channel Form with the custom fields loop:
   ```
   {exp:channel:form channel="your_channel"}
     {custom_fields}
       [{field_id} : {field_name}]
     {/custom_fields}
   {/exp:channel:form}
   ```
3. Load the page.
4. `{field_name}` resolves but `{field_id}` is left literal, e.g. `[{field_id} : title]`.

(On a server where PDO returns numerics as strings, the same template renders `[9 : title]` correctly.)

**Error Messages**
<pre><code>
No PHP error or exception is thrown. The variable is silently left unparsed in the rendered output.
</code></pre>

Template:
```
{exp:channel:form channel="companies"}
  {custom_fields}
    [{field_id} : {field_name}]
  {/custom_fields}
{/exp:channel:form}
```

Output on PHP 8.1+ (PDO native int) — `field_id` literal:
```
[{field_id} : cr_number][{field_id} : title_ar][{field_id} : company_code] ...
```

Output on PHP ≤ 8.0 (PDO string) — correct:
```
[1 : cr_number][9 : title_ar][2 : company_code] ...
```

Verified empirically (same DB, same query `SELECT field_id FROM exp_channel_fields LIMIT 1`):
```
PHP 7.4.33  PDO default  field_id => string  '1'
PHP 8.3.30  PDO default  field_id => integer  1
```

**Environment Details:**
 - Version: 7.5.24
 - PHP Version: 8.3.30 (reproduces on 8.1+); works on 7.4.33
 - MySQL Version: MySQL 8.0
 - OS: macOS (local) / Linux (production)
 - Web Server: Apache (MAMP local), Apache (cPanel/A2 Hosting production)

**Possible Solution**

In the `{custom_fields}` substitution loop in `Channel_form_lib.php`, cast documented numeric loop variables to string before substitution so they resolve regardless of the DB layer's numeric return type. For example, handle `field_id` explicitly:

```php
} elseif (! is_int($value)) {
    $temp = ee()->TMPL->swap_var_single($key, $value, $temp);
} elseif ($key === 'field_id') {
    $temp = ee()->TMPL->swap_var_single($key, (string) $value, $temp);
}
```

A more general fix would be to whitelist the documented loop variables (`field_id`, etc.) and string-cast them, rather than relying on `is_int()` to distinguish "conditional" integers from substitutable ones — since that distinction is now sensitive to the PDO driver's native-type behavior on PHP 8.1+.

**Additional context**

The official Channel Form documentation lists `{field_id}` as an available single variable inside the `{custom_fields}` loop, so the expectation is that it resolves on all supported PHP versions. Because EE 7.5 officially supports PHP 8.1+, the `is_int()` guard combined with PDO's native-int return on those versions makes a documented variable silently fail depending on the environment's DB layer, with no error to indicate why.

bryannielsen:
Resolved #5320 where Channel Form {field_id} was not always parsed

### DIFF
--- a/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
+++ b/system/ee/ExpressionEngine/Addons/channel/libraries/channel_form/Channel_form_lib.php
@@ -305,14 +305,7 @@ class Channel_form_lib
                     );
                 }
 
-                foreach ($custom_field_variables_row as $key => $value) {
-                    if (is_array($value)) {
-                        $temp = $this->swap_var_pair($key, $value, $temp);
-                    } elseif (! is_int($value)) {
-                        // don't use our conditionals as vars
-                        $temp = ee()->TMPL->swap_var_single($key, $value, $temp);
-                    }
-                }
+                $temp = $this->_swap_custom_field_variables($custom_field_variables_row, $temp);
 
                 if ($custom_field_variables_row['field_type'] === 'catchall') {
                     $temp = $this->replace_tag($field_name, $this->entry($field_name), array(), $temp);
@@ -1109,6 +1102,29 @@ GRID_FALLBACK;
         }
 
         return $custom_field_variables;
+    }
+
+    /**
+     * Swap variables inside a single {custom_fields} row.
+     *
+     * @param   array   $custom_field_variables_row
+     * @param   string  $tagdata
+     * @return  string
+     */
+    private function _swap_custom_field_variables($custom_field_variables_row, $tagdata)
+    {
+        foreach ($custom_field_variables_row as $key => $value) {
+            if (is_array($value)) {
+                $tagdata = $this->swap_var_pair($key, $value, $tagdata);
+            } elseif ($key === 'field_id') {
+                $tagdata = ee()->TMPL->swap_var_single($key, (string) $value, $tagdata);
+            } elseif (! is_int($value)) {
+                // don't use our conditionals as vars
+                $tagdata = ee()->TMPL->swap_var_single($key, $value, $tagdata);
+            }
+        }
+
+        return $tagdata;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TomJaeger:
@codex please review

TomJaeger (approved):

PR #5324, opened 2026-06-17
Resolved #963 where random string generation was not strong

### LINKED ISSUE #963: Weak RNG used to generate tokens
I was trying to figure out how `authcode` is generated for new accounts, and it seems to use the insecure mt_rand() function to generate secure tokens.

From system/ee/ExpressionEngine/Addons/member/mod.member_register.php:

```php
                // We generate an authorization code if the member needs to self-activate
                if (ee()->config->item('req_mbr_activation') == 'email')
                {
                        $data['authcode'] = ee()->functions->random('alnum', 10);
                }
```

I'm pretty sure that comes from system/ee/legacy/libraries/Functions.php:

```php
        public function random($type = 'encrypt', $len = 8)
        {
                return random_string($type, $len);
        }
```

Which probably comes from system/ee/legacy/helpers/string_helper.php:

```php
        function random_string($type = 'alnum', $len = 8, $antipool = '')
        {
                switch($type)
                {
[...]
                        case 'alnum'    :
[...]
                                                case 'alnum'    :       $pool = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
                                                        break;
[...]
                                                $str .= substr($pool, mt_rand(0, strlen($pool) -1), 1);
```

And mt_rand() isn't designed for secure strings.. from PHP's site (https://www.php.net/manual/en/function.mt-rand.php):

> This function does not generate cryptographically secure values, and should not be used for cryptographic purposes. If you need a cryptographically secure value, consider using random_int(), random_bytes(), or openssl_random_pseudo_bytes() instead.

Cheers!

bryannielsen:
Resolved #963 where random string generation was not strong

### DIFF
--- a/system/ee/ExpressionEngine/Service/Encrypt/Encrypt.php
+++ b/system/ee/ExpressionEngine/Service/Encrypt/Encrypt.php
@@ -265,11 +265,11 @@ class Encrypt
     /**
      * Generates a random key to be used for anything, probably encryption
      *
-     * @return string 32-character key
+     * @return string 40-character key
      */
     public function generateKey()
     {
-        return sha1(uniqid(random_int(-PHP_INT_MAX, PHP_INT_MAX), true));
+        return bin2hex(random_bytes(20));
     }
 }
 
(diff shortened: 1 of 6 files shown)


TomJaeger:
@codex please review

TomJaeger (approved):

PR #5326, opened 2026-06-22
Resolved #5325 where File fieldtype in Grid triggered validation for unchanged values

### LINKED ISSUE #5325: Non-admin validation on grid > file fields missing col_id when checking for changes

Non-super-admin CP users cannot save channel entries that contain a **Grid** field with a **file** column when existing file values point at upload destinations the editor’s role is not assigned to — **even when the file column is unchanged**.

We hit this after migrating standalone file fields into a standard Grid field (`field_type: grid`) with `col_type: file` columns. SAEL-role editors need to update dropdown/text fields (e.g. Renewal Status) on entries that already have certificate PDFs in grid rows. Those PDFs live in upload destinations restricted to the Marketing role. The editors are not trying to upload or change files.

On save (including save with no edits), validation fails on every populated grid `:file` cell with:

`You do not have access to the directory specified for this field`

and the entry banner:

`There was a problem with one or more Grid fields`

Standalone (non-grid) file fields on the same channel do **not** exhibit this when the file value is unchanged — only grid file columns.

**CP URL pattern:** `/admin.php?/cp/publish/edit/entry/{entry_id}`

**Expected:** Unchanged grid file values skip upload-directory permission checks on edit (same as standalone file fields).

**Actual:** Permission check runs on every save; save fails if the member lacks access to the file’s upload destination.

---

**How To Reproduce**

1. Create (or use) two member roles: **Role A** (can edit entries) and **Role B** (can upload to a specific upload destination).
2. Create an upload destination assignable only to **Role B** (not Role A).
3. Upload a test PDF to that destination as a Role B / super-admin user.
4. Create a **Grid** channel field with a `file` column (`allowed_directories: all` is fine).
5. Create/edit an entry as super-admin; add a grid row with the PDF in the file column; save.
6. Log in as a **Role A** member who can edit that entry but is **not** assigned to the upload destination.
7. Open the entry in CP and click **Save** without changing anything (or change an unrelated field, e.g. a text/select field).
8. Save fails with `directory_no_access` on the grid file column(s).

Minimal setup: one grid field, one file column, one row with an existing file in a restricted upload destination, editor role without that destination assigned.

---

**Error Messages**

<pre><code>
You do not have access to the directory specified for this field

There was a problem with one or more Grid fields
</code></pre>

No PHP exception/stack trace — validation error returned from `File_ft::validate()`.

---

Image

Client provided screenshot of issue as it shows untouched file fields failing validation.

Relevant core code (stock EE, before any local patch):

`contentzsystem/ee/ExpressionEngine/Addons/file/ft.file.php` — `validate()` method, grid branch (~lines 80–96):

```php
if (isset($this->settings['grid_row_id']) || isset($this->settings['grid_row_name'])) {
    // ...
    $rows = ee()->grid_model->get_entry_rows(/* ... */);

    // If this filed was we need to check permissions.
    if (! isset($this->settings['grid_row_id']) || $rows[$this->content_id][$this->settings['grid_row_id']] != $data) {
        $check_permissions = true;
    }
}
```

`$rows[$entry_id][$row_id]` from `grid_model->get_entry_rows()` is the **full grid row array** (`row_id`, `entry_id`, `col_id_*`, …). `$data` is only the **file column string** (e.g. `{filedir_39}certificate.pdf` or `{file:123:url}`). Comparing array ≠ string is always true, so `$check_permissions` is always set on grid edits and `memberHasAccess()` always runs.

The non-grid branch correctly compares `$entry->$field_name != $data`.

---

**Environment Details:**

- Version: **7.5.22**
- PHP Version: **8.3**
- MySQL Version: **8.x**
- OS: **macOS** (local); **Linux** (production)
- Web Server: **nginx** (Docker `webdevops/php-nginx`)

---

**Possible Solution**

In `File_ft::validate()`, grid branch, compare the **specific file column cell** instead of the whole row:

**File:** `system/ee/ExpressionEngine/Addons/file/ft.file.php`  
**Function:** `validate()`

```php
$col_key = 'col_id_' . $this->settings['col_id'];
$existing = $rows[$this->content_id][$this->settings['grid_row_id']][$col_key] ?? null;

if (! isset($this->settings['grid_row_id']) || $existing != $data) {
    $check_permissions = true;
}
```

`col_id` is already set on grid-instantiated file fieldtypes via `Grid_parser::instantiate_fieldtype()`.

This mirrors the non-grid logic: skip upload-directory permission checks when the file value is unchanged on edit; still check when the file is new or modified.

---

**Additional context**

- Regression surfaced when moving legacy standalone file fields into grid file columns; standalone fields worked for the same users/files before.
- Not a grid field configuration issue — `allowed_directories: all` on the column is correct.
- Workaround: assign every editor role to every upload destination referenced by existing grid files (undesirable for roles that should not upload).
- We have a local patch + change log documenting the vendor modification until an upstream fix ships.

bryannielsen:
Resolved #5325 where File fieldtype in Grid triggered validation for unchanged values

### DIFF
--- a/system/ee/ExpressionEngine/Addons/file/ft.file.php
+++ b/system/ee/ExpressionEngine/Addons/file/ft.file.php
@@ -90,9 +90,21 @@ class File_ft extends EE_Fieldtype implements ColumnInterface
                         $fluid_field_data_id
                     );
 
-                    // If this filed was we need to check permissions.
-                    if (! isset($this->settings['grid_row_id']) || $rows[$this->content_id][$this->settings['grid_row_id']] != $data) {
+                    // If this field changed we need to check permissions.
+                    if (! isset($this->settings['grid_row_id'])) {
                         $check_permissions = true;
+                    } else {
+                        $row = isset($rows[$this->content_id][$this->settings['grid_row_id']])
+                            ? $rows[$this->content_id][$this->settings['grid_row_id']]
+                            : array();
+                        $col_id = isset($this->settings['col_id']) ? 'col_id_' . $this->settings['col_id'] : null;
+                        $existing = (is_array($row) && $col_id && array_key_exists($col_id, $row))
+                            ? $row[$col_id]
+                            : null;
+
+                        if ($existing != $data) {
+                            $check_permissions = true;
+                        }
                     }
                 } else {
                     $entry = ee('Model')->get('ChannelEntry', $this->content_id)->first();
--- a/system/ee/ExpressionEngine/Tests/ExpressionEngine/Addons/file/FT/FileFtValidateTest.php
+++ b/system/ee/ExpressionEngine/Tests/ExpressionEngine/Addons/file/FT/FileFtValidateTest.php
@@ -205,11 +205,17 @@ class FileFtValidateTest extends FileFtTestBase
             'grid_row_id' => 7,
             'grid_field_id' => 33,
             'grid_content_type' => 'channel',
+            'col_id' => 5,
         ], 19);
         $file = new FileFtFileModelStub(false);
         $gridModel = $this->setGridRows([
             19 => [
-                7 => '{filedir_1}banner.png',
+                7 => [
+                    'row_id' => 7,
+                    'entry_id' => 19,
+                    'col_id_5' => '{filedir_1}banner.png',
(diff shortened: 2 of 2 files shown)


TomJaeger:
@codex please review

TomJaeger (approved):

PR #5341, opened 2026-06-30
Fixed #5340 react select list issue

### LINKED ISSUE #5340: virtualized React Select (select_list) — channel field list rows vanish on scroll
Weird one- came in via email and confirmed.

Have a lot of custom fields- more than 100, which is where virtualization kicks in.  Go edit a channel, go to the field select, start scrolling down.  And eventually, they just start to disappear and you don't see them.

Image

Our very clever user also provides a fix, but I'll get someone with better frontend chops to implement!

Rows scroll out and disappear; past ~halfway the whole list vanishes.
Console: TypeError: Cannot read properties of undefined (reading 'instructions') at getVirtualItemHeight (VirtualizedItemList.getOffsetTop → render).
Root cause — themes/ee/asset/javascript/src/components/select_list.es6, VirtualizedItemList:
The component uses two different height models. getVisibleRange() computes startIndex/visibleCount from a fixed 40px per row (Math.floor(scrollTop / 40)), but getOffsetTop()/getTotalHeight() position and size the list using the variable getVirtualItemHeight() (40/60/70/35 depending on instructions/toggles/isSection). Since real rows are taller than 40, offsetTop grows faster than scrollTop, so the absolutely-positioned window drifts below the viewport as you scroll → rows vanish. The same over-estimated indices push items[i] out of range, handing undefined to getVirtualItemHeight, which reads item.instructions → the TypeError.

Fix: compute the visible range by walking the same variable heights, use the configured virtualizationHeight, and guard against an undefined item.

 function getVirtualItemHeight(item) {
+  if (!item) return 40
   let height = 40
   if (item.instructions) { height += 20 }
   if (item.toggles && Object.keys(item.toggles).length > 0) { height += 10 }
   if (item.isSection) { height = 35 }
   return height
 }

 getVisibleRange() {
-  const scrollTop = this.state.scrollTop
-  const containerHeight = 400          // Default container height
-  const itemHeight = 40                // Base item height
-  const overscan = 10
-  const startIndex = Math.max(0, Math.floor(scrollTop / itemHeight) - overscan)
-  const visibleCount = Math.ceil(containerHeight / itemHeight) + overscan * 2
-  const endIndex = Math.min(this.props.items.length - 1, startIndex + visibleCount)
-  return { startIndex, endIndex }
+  const scrollTop = this.state.scrollTop
+  const containerHeight = this.props.virtualizationHeight || 400
+  const overscan = 10
+  const items = this.props.items
+
+  // Visible range must use the SAME variable heights as getOffsetTop()/getTotalHeight().
+  // The old fixed-40px math drifted out of sync with the variable-height offsets, so the
+  // rendered window slid below the viewport and rows vanished on scroll.
+  let startIndex = 0, cumulative = 0
+  while (startIndex < items.length) {
+    const h = getVirtualItemHeight(items[startIndex])
+    if (cumulative + h > scrollTop) break
+    cumulative += h
+    startIndex++
+  }
+
+  let endIndex = startIndex, visibleHeight = 0
+  while (endIndex < items.length && visibleHeight < containerHeight) {
+    visibleHeight += getVirtualItemHeight(items[endIndex])
+    endIndex++
+  }
+
+  startIndex = Math.max(0, startIndex - overscan)
+  endIndex = Math.min(items.length - 1, endIndex + overscan)
+  return { startIndex, endIndex }
 }

Note: patch the source select_list.es6 (compiles to compressed/components/select_list.js). getVisibleRange is now O(n) per scroll — same complexity class as the existing getOffsetTop/getTotalHeight; a memoized prefix-sum of heights would reduce all three if you want it cheaper for very large lists. Verified on a production 7.5.21 install: list scrolls correctly, scrollbar length correct, no vanishing, console error gone.

robinsowell:
https://github.com/ExpressionEngine/ExpressionEngine/issues/5340

Where large scrolls could eventually 'lose' the options.

### DIFF
--- a/system/ee/ExpressionEngine/View/_shared/form/fields/select.php
+++ b/system/ee/ExpressionEngine/View/_shared/form/fields/select.php
@@ -150,9 +150,12 @@ else:
         'noResults' => isset($no_results['text']) ? lang($no_results['text']) : null,
         'splitForTwo' => isset($split_for_two) ? $split_for_two : null,
         'toggles' => isset($toggles) ? $toggles : [],
-        'virtualizationThreshold' => isset($virtualization_threshold) ? $virtualization_threshold : 100,
-        'virtualizationHeight' => isset($virtualization_height) ? $virtualization_height : 400
+        'virtualizationThreshold' => isset($virtualization_threshold) ? $virtualization_threshold : 100
     ];
+
+    if (isset($virtualization_height)) {
+        $component['virtualizationHeight'] = $virtualization_height;
+    }
     ?>
 	<div data-select-react="<?=base64_encode(json_encode($component))?>" data-input-value="<?=$field_name?>" class="<?=$class?>">
 		<div class="fields-select">
--- a/tests/cypress/cypress/elements/pages/channel/Channel.js
+++ b/tests/cypress/cypress/elements/pages/channel/Channel.js
@@ -38,6 +38,7 @@ class Channel extends ControlPanel {
 
             // Settings Tab
             "channel_description": 'textarea[name=channel_description]',
+            "channel_lang_filter": 'div[data-input-value="channel_lang"] .search-input__input',
             "channel_lang": 'div[data-input-value="channel_lang"] input[type="radio"]',
 
             "channel_url": 'input[type!=hidden][name=channel_url]',
--- a/tests/cypress/cypress/integration/channel/create.ee6.js
+++ b/tests/cypress/cypress/integration/channel/create.ee6.js
@@ -203,7 +203,8 @@ context('Channel Create/Edit', () => {
         page.load_edit_for_channel(2) // 2nd row, not channel id 2
         page.get('settings_tab').click()
         page.get('channel_description').clear().type('Some description')
-        page.get('channel_lang').check('en', { force: true })
+        page.get('channel_lang_filter').clear().type('English')
+        page.get('channel_lang').filter('[value="en"]').check({ force: true })
         page.get('channel_url').clear().type('http://someurl/channel')
         page.get('comment_url').clear().type('http://someurl/channel/comment')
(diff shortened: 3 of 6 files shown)


robinsowell:
reviewed: Summary: Fixes the virtualized select list "losing" options during large scrolls (#5340). The root cause was that VirtualizedItemList.getVisibleRange() computed the visible window using a fixed 40px row height, while the actual DOM positioning (getTotalHeight() / getOffsetTop()) used variable heights via getVirtualItemHeight(). On lists with taller rows (items with instructions), the two drifted apart and the rendered slice no longer matched where items were placed — producing blank gaps / vanishing options on long scrolls.

Verdict: Approve. The fix is correct and well-scoped.

What's good
getVisibleRange() now walks actual item heights, making it consistent with getTotalHeight()/getOffsetTop(). This is the right fix for the reported bug.
virtualizationHeight is now properly forwarded into <VirtualizedItemList>, so containerHeight honors the configured height instead of being hard-locked at 400.
Sensible defensive additions: the if (!item) return 40 guard in getVirtualItemHeight() and the empty-list early return in getVisibleRange().
The .es6 source and the hand-maintained .js artifact were updated consistently.

Minor (non-blocking)
In the end-of-range loop, visibleHeight accumulates from the top of startIndex rather than from scrollTop. Since the first item's top usually sits slightly above the viewport, the bottom can be under-counted by up to one row. The overscan of 10 comfortably absorbs this in practice, so no change required — just noting it.
select_list.js appears to be hand-maintained alongside select_list.es6 (no gulp step compiles it). Worth confirming that's the intended workflow so the two don't diverge in future edits.

Yulyaswan:
checked
from my side it's ready to approve 

@robinsowell if it works for you as well left comment and I'll approve it
Thanks

Yulyaswan (approved):

PR #5389, opened 2026-09-07
Fix layout loop indexes after Multi Select rendering

### LINKED ISSUE #5388: Unused {index} in a Multi Select tag pair empties {index} in a later {layout:...} array loop
## Description
*(Disclaimer: written with support from Claude)*

When a Multi Select fieldtype tag pair is rendered **without using its `{index}` variable**, the `{index}` variable inside a `{layout:...}` array loop in a parent layout template resolves to an empty string. Every `{layout:some-array index='{index}'}` lookup in that loop then returns an empty string.

Both variables are documented:

- Multi Select tag pairs expose `{index}` as a loop variable — [Multi Select docs](https://docs.expressionengine.com/latest/fieldtypes/multiselect.html#variable-pair)
- Layout arrays require `{index}` to address parallel lists — [Template Layouts docs](https://docs.expressionengine.com/latest/templates/layouts.html#using-layout-lists-together)

**Expected:** the layout loop's own `{index}` yields 0, 1, 2 … and the `index=` lookups return the matching values.

**Actual:** `{index}` is empty inside the layout loop, so all `index=` lookups return empty strings. The loop iterates the correct number of times, and `{value}` still resolves correctly.

**Key finding:** simply *outputting* `{index}` anywhere inside the Multi Select tag pair makes the problem disappear entirely. This suggests the Multi Select pair registers `{index}` and, when the variable is never consumed within the pair, the registration leaks into the later layout-array parsing.

## Minimal reproduction

Three templates (here in a group named `_debug`). The Multi Select field here is `my-genre`; substitute any Multi Select field with at least one selected option.

### `_debug/test`

~~~
{layout='_debug/_bc'}

{layout:set:append name='crumb-label'    value='Directory'}
{layout:set:append name='crumb-position' value='2'}
{layout:set:append name='crumb-path'     value='directory'}

{exp:channel:entries channel='CHANNEL' limit='1' entry_id='ENTRY_ID'}

  {layout:set:append name='crumb-label'    value='{title}'}
  {layout:set:append name='crumb-position' value='3'}
  {layout:set:append name='crumb-path'     value=''}

  {layout:set name='body'}
    <h1>{title}</h1>
    <ul>{my-genre}<li>{item:label}</li>{/my-genre}</ul>
  {/layout:set}

{/exp:channel:entries}
~~~

### `_debug/_bc`

~~~
{layout='_debug/_gl'}

{layout:set name='crumb-html'}
<ol>
{if layout:crumb-label}{layout:crumb-label}
  <li>value=[{value}] path=[{layout:crumb-path index='{index}'}] position=[{layout:crumb-position index='{index}'}]</li>
{/layout:crumb-label}{/if}
</ol>
{/layout:set}
~~~

### `_debug/_gl`

~~~
<h2>Breadcrumb</h2>
{layout:crumb-html}

<h2>Body</h2>
{layout:body}
~~~

### Result

~~~
<li>value=[Directory] path=[] position=[]</li>
<li>value=[ENTRY TITLE] path=[] position=[]</li>
~~~

Expected:

~~~
<li>value=[Directory] path=[directory] position=[2]</li>
<li>value=[ENTRY TITLE] path=[] position=[3]</li>
~~~

### The workaround until bugfix:

Output `{index}` inside the Multi Select pair — the value itself is irrelevant:

~~~
<ul>{my-genre}<li>{index}{item:label}</li>{/my-genre}</ul>
~~~

The layout loop then resolves correctly. Using it in a data-attribute or something truly hidden works as a production workaround.

## Scope

Tested by swapping only the contents of `{layout:set name='body'}`:

| Body contents | Layout loop |
| --- | --- |
| `<h1>{title}</h1>` only | works |
| Multi Select **tag pair**, `{index}` unused | **broken** |
| Multi Select tag pair, `{index}` output somewhere inside | works |
| Multi Select as a **single tag** (`{my-genre}`) | works |
| Relationship field **tag pair** | works |

So this is specific to the Multi Select tag pair, and specifically to the case where its `{index}` is never used.

Additional notes from the larger template where this first appeared:

- The **Layout Variables** dump in the Template Debugger is byte-identical in the working and broken cases — the layout arrays are complete and correct.
- **Undefined Variables** does not list the `index=` lookups, so they do execute; they simply resolve against an empty index.
- Not size related: several hundert KB of additional markup, `{parents}` loops and third-party module tags do not trigger it.
- Removing `{layout:body}` from the outermost layout entirely does not help — the pair alone is enough, even when its output is never placed on the page.
- Replacing `{item:label}` with static text inside the pair does not help.

## Environment

- ExpressionEngine: 7.5.26
- PHP: 8.2

TomJaeger:
## Summary

- clear the cached missing-variable state when layout parsing starts
- preserve layout loop indexes after a Multi Select pair is rendered without `{index}`
- add regression coverage for layout counters and indexed lookups across multiple rows

## Root cause

Rendering a Multi Select pair without `{index}` leaves a cached "variable not found" entry. A later layout loop then skips its own index, causing indexed lookups into parallel layout arrays to return empty strings.

Fixes #5388

## Validation

- regression confirmed failing before the fix and passing afterward
- layout tests: 17 tests, 25 assertions on PHP 7.4, 8.2, and 8.5
- adjacent parser tests: 58 tests, 100 assertions on PHP 8.2
- six isolated diagnostics using the real Multi Select and template parsing methods on PHP 8.2
- independent review found no actionable issues
- `git diff --check` passed

Full browser rendering and the full test suite were not run.

### DIFF
--- a/system/ee/ExpressionEngine/Tests/ExpressionEngine/legacy/EE_Template/EE_TemplateParseLayoutVariablesTest.php
+++ b/system/ee/ExpressionEngine/Tests/ExpressionEngine/legacy/EE_Template/EE_TemplateParseLayoutVariablesTest.php
@@ -71,6 +71,32 @@ class EE_TemplateParseLayoutVariablesTest extends EE_TemplateTestBase
     }
 
     /**
+     * Parse layout loop variables even when a previous loop did not use them.
+     *
+     * @return void
+     */
+    public function testParseLayoutVariablesIgnoresPreviousLoopMissingVariables()
+    {
+        ee()->setMock('Variables/Parser', new \ExpressionEngine\Service\Template\Variables\LegacyParser());
+
+        $this->template->parse_variables_row('{item}', [
+            'item' => 'Rock',
+            'index' => 0,
+            'count' => 1,
+        ]);
+
+        $result = $this->template->parseLayoutVariables(
+            "{layout:labels}{index}:{count}:{value}:{layout:paths index='{index}'};{/layout:labels}",
+            [
+                'labels' => ['Directory', 'Entry'],
+                'paths' => ['directory', 'entry'],
+            ]
+        );
+
+        $this->assertSame('0:1:Directory:directory;1:2:Entry:entry;', $result);
+    }
+
+    /**
      * Test parseLayoutVariables handles undefined variables
      */
     public function testParseLayoutVariablesHandlesUndefinedVariables()
--- a/system/ee/legacy/libraries/Template.php
+++ b/system/ee/legacy/libraries/Template.php
@@ -698,6 +698,7 @@ class EE_Template
     {
         $this->log_item("Layout Variables:", $layout_vars);
         $this->layout_conditionals = [];
+        $this->unfound_vars = [];
         $layout_conditionals = [];
 
         // get all the declared layout variables (excluding layout:contents)


albuvee:
LGTM! Can confirm that this resolves my original problem ( #5388 ) – and it also runs smoothly in all my debug tests. 🚀

